PR #1, opened 2016-05-27
Cleaned up formatting, Enhanced SPA functionality. Enabled redux-dev-tools

gu3st:
Cleaned up some weird formatting. Added redux-dev-tools. Used Link component where appropriate to enable SPA navigation.

### DIFF
--- a/components/NavBar.js
+++ b/components/NavBar.js
@@ -1,53 +1,58 @@
 import React from 'react';
-import
-{
-   connect
-}
-from 'react-redux';
-const NavBar = (props) => (<div style={{marginBottom:"0px"}} className="navbar">
-   <div className="navbar-header">
-      <a href="/" className="navbar-brand">
-         <strong className="theme-blue">YASP</strong>
+import { connect } from 'react-redux';
+
+import { Link } from 'react-router';
+
+const NavBar = (props) => {
+   var navLeft, navRight
+   if(props.isFetching){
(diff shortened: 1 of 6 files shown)


components/NavBar.js
```diff
@@ -1,53 +1,58 @@
...
-   <div className="navbar-header">
-      <a href="/" className="navbar-brand">
-         <strong className="theme-blue">YASP</strong>
+import { connect } from 'react-redux';
+
+import { Link } from 'react-router';
+
+const NavBar = (props) => {
+   var navLeft, navRight
+   if(props.isFetching){
```
howardchung: own line

components/NavBar.js
```diff
@@ -1,53 +1,58 @@
...
+      navLeft = (<li><i className="fa fa-spinner fa-spin"></i></li>);
+      navRight = (<li><i className="fa fa-spinner fa-spin"></i></li>);
+   }
+   else{
```
howardchung: own line

mike-robertson:
I added all these changes (except a few style changes which wouldn't make sense to spend time on with the file shift). I think if we can get to one codebase quickly, it will be better so we aren't going 2 separate directions with design. I can probably refactor the rest of the codebase to be more component-ized (see: https://facebook.github.io/react/docs/thinking-in-react.html) relatively quickly though.

PR #5, opened 2016-05-29
Mike | Added the table component

mike-robertson:
To test, go to /players/:realPlayer…Id in url. There needs to be some work to move around some of the logic. I'm not sure where the transformations of the volvo data should live, right now they are in the yasp.config.js module. Also added in some basic css-modules and post css. See Table directory for examples as well as palette.css under components dir


components/AccountWidget/AccountWidget.jsx
```diff
@@ -12,7 +12,7 @@ const AccountWidget = ({ loading, error, player }) => (
     {error && <div >Error</div>}
     {!error && !loading && player ? (
       <div>
-        <li><Link to={`/players/${player.account_id}`}>Profile</Link></li>
+        <li><Link to={`/players/${player.profile.account_id}`}>Profile</Link></li>
```
howardchung: Assuming this player is generated from metadata property `user` (which it should be!) it should not be wrapped in profile.

If you are using the state holding the current player page then this will link to the profile of whatever player you have loaded ATM.

components/AccountWidget/AccountWidget.jsx
```diff
@@ -23,7 +23,7 @@ const AccountWidget = ({ loading, error, player }) => (
 
 class AccountWidgetWrapper extends React.Component {
   componentDidMount() {
-    this.props.getPlayer(this.props.playerId);
+    this.props.getPlayer(15446376);
```
howardchung: hardcoded?

components/Table/Table.jsx
```diff
@@ -1,6 +1,57 @@
...
+          ))}
+        </TableBody>
+      </MaterialTable>
+    </MuiThemeProvider>
+  );
+
+  return (
+    <div>
+      {loading && <Spinner />}
+      {!loading && error && <div>Whoops, something went wrong</div>}
```
howardchung: Standardize any error messages.

In fact, any string should be extracted out into a strings file for future internationalization.

reducers/gotPlayerMatches.js
```diff
@@ -0,0 +1,30 @@
+import { playerMatchesActions } from '../actions';
```
howardchung: Is it possible to combine all the gotPlayer* reducers into a single one that takes an additional param `tab` and extracts the necessary properties based on that?

yasp.config.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+
+// there is no possible way this is the right place for this logic...
+// TODO - put this shit in the right place
+const matchColumns = [{
```
howardchung: In whatever is rendering the component MatchTable?

yasp.config.js
```diff
@@ -0,0 +1,99 @@
...
+}];
+
+// there is no possible way this is the right place for this logic...
+// TODO - put this shit in the right place
+const playerMatchTransformSwitch = (match, field) => {
+  switch (field) {
+    case 'hero_id':
+      return 'https://yasp.co/apps/dota2/images/heroes/pudge_full.png';
+    case 'radiant_win':
+      return match.player_slot < 64 && match[field] || match.player_slot >= 64 && !match[field] ? 'W' : 'L';
```
howardchung: all strings need to be extracted out into a file

yasp.config.js
```diff
@@ -0,0 +1,99 @@
...
+    case 'game_mode':
+      return match[field] === 22 ? 'Ranked All Pick' : 'All Pick';
```
howardchung: see `json` directory in yasp core for mappings.  Should we have the server return them, or bundle them into the webpack build?

yasp.config.js
```diff
@@ -0,0 +1,99 @@
...
+    case 'start_time': {
+      const today = new Date();
+      const matchDate = new Date(match[field] * 1000);
+      const leftoverDays = Math.floor((today - matchDate) / 1000 / 60 / 60 / 24);
+      switch (leftoverDays) {
```
howardchung: Moment.js handles this for us automatically.  See our existing templates for how it works.

components/AccountWidget/AccountWidget.jsx
mike-robertson: testing, must have forgotten to remove it.

components/Table/Table.jsx
mike-robertson: I was thinking that we could create notifications type component one of which would be error messages.

howardchung: For now, just show a string that's defined in a string file.  Fancy errors are definitely outside MVP

reducers/gotPlayerMatches.js
mike-robertson: yea they can be separate reducers which are combined under gotPlayer so we don't have some mammoth reducer file. So like gotPlayer.matches or gotPlayer.otherStuff

components/Table/Table.jsx
mike-robertson: That's easier for me

yasp.config.js
```diff
@@ -0,0 +1,99 @@
...
+        case 0:
+          return 'Today';
+        case 1:
+          return 'Yesterday';
+        default:
+          return `${leftoverDays} days ago`;
+      }
+    }
+    case 'duration':
+      return `${Math.floor(match[field] / 60)}:${match[field] % 60}`;
```
howardchung: same here, moment will handle the date formatting

actions/constantsActions.js
```diff
@@ -0,0 +1,43 @@
...
+  payload,
+});
+const getConstantsError = (payload) => ({
+  type: ERROR,
+  payload,
+});
+
+export const getConstants = (host = HOST_URL) => (dispatch) => {
+  getConstantsRequest();
+  return fetch(`${host}${url}`, { credentials: 'include' })
```
howardchung: technically this isn't needed except for metadata and would save a tiny bit of traffic

actions/playerMatchesActions.js
```diff
@@ -0,0 +1,81 @@
+import fetch from 'isomorphic-fetch';
+import { HOST_URL } from './';
+
+const url = (playerId, numMatches) => `/api/players/${playerId}/matches?limit=${numMatches}`;
```
howardchung: For future consideration--the filters are also all implemented as query params.  Start thinking about how we can work that in.

actions/playerMatchesActions.js
```diff
@@ -0,0 +1,81 @@
...
+  payload,
+});
+
+// TODO - memoize this with selectors
+const playerMatchTransformSwitch = (match, field, heroes) => {
+  switch (field) {
+    case 'hero_id':
+      return `${HOST_URL}${heroes[match[field]].img}`;
+    case 'radiant_win':
+      return match.player_slot < 64 && match[field] || match.player_slot >= 64 && !match[field] ? 'W' : 'L';
```
howardchung: isRadiant is going to be a function used a lot, move it to some utility function file.  Also 128 is more correct

PR #6, opened 2016-05-29
basic styles and tag cleaning

Marvizm:


components/AccountWidget/AccountWidget.jsx
```diff
@@ -7,18 +7,18 @@ import { Link } from 'react-router';
 
 // Maybe we can factor out this ternary into a function?
 const AccountWidget = ({ loading, error, player }) => (
-  <div>
+  <menu>
```
howardchung: is this an actual HTML tag?

index.html
```diff
@@ -24,7 +24,9 @@
     <meta name="msapplication-TileColor" content="#da532c">
     <meta name="msapplication-TileImage" content="/public/images/mstile-144x144.png">
     <meta name="msapplication-config" content="/public/images/browserconfig.xml">
-    <meta name="theme-color" content="#ffffff"> </head>
+    <meta name="theme-color" content="#ffffff"> 
+    <link rel="stylesheet" href="yasp.css">
```
howardchung: should be bundled via webpack

howardchung:
@mike-robertson should review and ensure it merges cleanly with his change

components/AccountWidget/AccountWidget.jsx
gu3st: Yeah. It's part of HTML5.  There's a bunch of them. section, article, header, footer (and more).  They're all just block-level elements (divs) but hold semantic value for search engines (not that the SPA will work with search engines, unless we attempt to implement server-side rendering at a later date).

howardchung:
Rebase against master if you want to merge this

PR #8, opened 2016-05-30
Mike | Fixed issue where table loading player data before constants w…

mike-robertson:
…as loaded would cause an error. Use createConstantsWrapper container factory which will add a layer between the dependant component and the constants service. Just export your final component by calling the factory with the componet you actually want to render. WHile constants is loading, it will display a loading bar (or error if constants fails to load). All props given to the constants wrapper will be passed down to your component


components/Cheese/CheeseButton.jsx
```diff
@@ -6,16 +6,14 @@ import RaisedButton from 'material-ui/RaisedButton';
 
 export default () => (
   <Link to="/carry">
+
+    <FontIcon className={styles.cheese} />
     <RaisedButton
+      label="Help us out"
```
howardchung: use the constant string?

components/Table/PlayerMatchesTable.container.jsx
```diff
@@ -20,14 +20,17 @@ const mapDispatchToProps = (dispatch) => ({
...
 
-class TableWrapper extends React.Component {
+class RequestLayer extends React.Component {
   componentDidMount() {
     this.props.getPlayerMatches(this.props.playerId, this.props.numMatches);
   }
 
   render() {
     return <Table {...this.props} columns={playerMatchesColumns} />;
+    // return <div>HERE I AM</div>
```
howardchung: debug code?

mike-robertson: whoops

PR #9, opened 2016-05-30
implement match table

howardchung:
Reviewers:
@gu3st 
@mike-robertson


components/Table/MatchTable.container.jsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+const mapDispatchToProps = (dispatch) => ({
+  sort: (column) => dispatch(getMatch(column)),
+  getMatch: (matchId) => dispatch(getMatch(matchId)),
+});
+
+class RequestLayer extends React.Component {
+  componentDidMount() {
+    this.props.getMatch(this.props.matchId);
```
howardchung: request should be in Match?

components/Match/Match.jsx
```diff
@@ -1,10 +1,59 @@
...
+  getMatch: (matchId) => dispatch(getMatch(matchId)),
+});
+
+class RequestLayer extends React.Component {
+  componentDidMount() {
+    this.props.getMatch(this.props.routeParams.match_id);
+  }
+
+  render() {
+    return (<div>
```
mike-robertson: Can we put the div on a separate line like so:

```js
return (
  <div>
    <Component />
  </div>
);
```

components/Match/Match.jsx
```diff
@@ -1,10 +1,59 @@
...
+      {<Table {...this.props} data={this.props.data && this.props.data.players ? this.props.data.players : this.props.data} columns={columns} />}
+    </div>);
+  }
+}
```
mike-robertson: Can we put the div on a separate line like so:

```js
return (
  <div>
    <Component />
  </div>
);
```

Also, instead of doing `data={this.props.data ... this.props.data ... this.props.data}`, we can destructure the data object so you don't need to write it so verbosely (making the line longer and harder to read mostly, also a pain to type).

```js
  const { data } = this.props;
  return <div><Table {...this.props} data={data && data.players ? ...
```

actions/matchActions.js
```diff
@@ -25,15 +25,10 @@ const getMatchError = (payload) => ({
...
-  .catch(error => error);
-};
-
-export const getMatch = (matchId, host = HOST_URL) => (dispatch) => {
+export const getMatch = (matchId) => (dispatch) => {
   dispatch(getMatchRequest());
-  matchApiCall(matchId, host)
+  return fetch(`${HOST_URL}${url}${matchId}`)
+    .then(response => response.json())
     .then(json => dispatch(getMatchOk(json)))
```
mike-robertson: Any reason for moving api call into action? Right now we are doing it in a lot of places, but in teh future it may be easier if we factor that code out into a separate modules for api calls so the action creators and thunks can be cleaner.

components/Table/index.js
```diff
@@ -2,4 +2,4 @@ import Table from './Table';
 import PlayerMatchesTable from './PlayerMatchesTable.container';
 
 export default Table;
-export { PlayerMatchesTable };
+export { PlayerMatchesTable, Table };
```
mike-robertson: I don't know if we need to explicitly export table like that as it is the default export. I don't think it will harm anything though, it might be nice even to have both options.

actions/matchActions.js
howardchung: matching player action.  No other reason

components/Match/Match.jsx
howardchung: fixed

PR #12, opened 2016-06-01
Mike | table will now sort if you pass in a sorting function. Follow …

mike-robertson:
…the examples in teh playerMatchTable.container file as well as the playerMatchesColumns.js file. Using a memoized selector to sort (sortPlayerMatches.js in the selectors folder). Also changed the reducer (/gotPlayer/matches.js) to have some sorting state which is used by the selector to sort the table


howardchung:
looks okay to me.  ready to merge?  anyone else want to review?

mike-robertson:
I'm going to refactoring the table component to be more componentized as right now it's sort of a monolith. Once I push that and it passes review we can merge

components/Table/playerMatchesColumns.js
```diff
@@ -1,49 +1,60 @@
...
+  if (a[field] > b[field]) return 1;
+  return 0;
+});
+
+const useOriginalValueSort = (array, field) => {
+  const origField = field.slice(0, field.lastIndexOf('Display'));
+  return defaultSort(array, origField);
+};
+
 export default [{
```
howardchung: I think we should pass field as the name of the field (no "Display"), and pass the display function as one of the props (dispFn)

components/Table/columnDefinitions/utility.js
```diff
@@ -0,0 +1,10 @@
+export const defaultSort = (array, field) => array.sort((a, b) => {
+  if (a[field] < b[field]) return -1;
+  if (a[field] > b[field]) return 1;
+  return 0;
+});
+
+export const useOriginalValueSort = (array, field) => {
+  const origField = field.slice(0, field.lastIndexOf('Display'));
```
howardchung: I will probably clean this up (I don't like using "Display" as a suffix)

PR #15, opened 2016-06-03
Mike | refactor the columns/data to not use that string parsing metho…

mike-robertson:
…d. INstead, the match[field] = { display: displayValue, value: apiValue }; Then in the table we reference match[field].display, and in the sort we do the same thing, but now we pass in a parameter called 'property' which has a default value of 'display', and the orig value sort passes in 'value' and the actual sort reads it as match[field][property].


howardchung:
See my changes.  I went for a different approach that will conflict with this one.

mike-robertson:
I think this method is better as it allows easier and more flexible sorting and it feels cleaner since transform code is only called by the selector.

mike-robertson:
I don't mind doing the refactor work on your branch if that's an issue though

howardchung:
In my branch I import constants synchronously (from static JSON) so in the column defs I can add a new function called dispFn that uses the constants to render a display string.

mike-robertson:
I disagree, even if we are statically loading constants we should be doing this transform logic as a selector.

mike-robertson:
That's just the standard pattern for changing/filtering data as it exists in the store

howardchung:
The difference is that we don't mutate the store.  We just define a
function used to generate what gets displayed

mike-robertson:
We aren't mutating the store with this implementation. We are creating a copy of the store which has a different structure for the values/display. I don't think it's possible to sort on this other method. We can only sort on the original values which we don't want except for certain cases.

howardchung:
Maybe store the sort value in your data structure as well.  Basically, I
would like all the string mapping to be close together, instead of
scattered.

mike-robertson:
We can move the transformation logic module into the selectors module since it should (in theory) only be needed by selectors. Then all the logic is encapsulated within the selectors

PR #16, opened 2016-06-04
Mike | Made changes to the table so it works responsively. Made chang…

mike-robertson:
…es to package.json to lint all folders since we have added more. Need to move to a src dir for our code so we don't have to keep adding to that script


components/Table/Table.jsx
```diff
@@ -18,32 +15,34 @@ const Table = ({ data, columns, loading, error, sortState, sortField, sortClick
...
+            sortClick={sortClick}
             totalWidth={totalWidth}
-            index={index}
           />
-        ))}
-      </MaterialTableBody>
-    </MaterialTable>
+        </MaterialTableHeader>
+        <MaterialTableBody displayRowCheckbox={false} selectable={false}>
+          {data.map((match, index) => (
```
howardchung: rename to row

package.json
```diff
@@ -9,7 +9,7 @@
     "start": "webpack-dev-server -d",
     "ncu": "npm-check-updates -u",
     "build": "npm run lint && npm run webpack",
-    "lint": "eslint ./components ./actions ./reducers ./yaspv2.js ./yasp.config.js --ext .js,.jsx"
+    "lint": "eslint ./actions ./components ./reducers ./selectors ./transformations ./yaspv2.js ./yasp.config.js --ext .js,.jsx"
```
howardchung: can't we just lint everything except build and node_modules?

PR #21, opened 2016-06-05
Mike | Added player header info. Still a work in progress css wise an…

mike-robertson:
…d tooltips


.eslintrc.json
```diff
@@ -4,7 +4,7 @@
   },
   "extends": "airbnb",
   "rules": {
-    "react/prop-types": 1,
+    "react/prop-types": 0,
```
howardchung: too many warnings?

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 .idea
 npm-debug.log
-build
\ No newline at end of file
+build
```
howardchung: remove the built file

components/Player/Player.jsx
```diff
@@ -1,10 +1,58 @@
...
+      this.props.getPlayer(nextProps.playerId);
+    }
+  }
+
+  render() {
+    return <Player {...this.props} />;
+  }
+}
+
+export default withRouter(connect(mapStateToProps, mapDispatchToProps)(RequestLayer));
```
howardchung: I'm not sure you need this withRouter.  See what I did in the Match component, you can read the Router state directly from the component.

components/Player/PlayerName.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
howardchung: Note: We should try to reuse this component for rankings, peers, etc.

components/Table/PlayerMatchesTable.container.jsx
```diff
@@ -29,7 +29,7 @@ class RequestLayer extends React.Component {
     this.props.getPlayerMatches(this.props.playerId, this.props.numMatches);
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillUpdate(nextProps) {
```
howardchung: What is the difference?

reducers/gotPlayer/player.js
```diff
@@ -28,3 +31,19 @@ export default (state = initialState, action) => {
       return state;
   }
 };
+
+export const player = {
```
howardchung: Why is this necessary?  Can't you just point the component to the path in the state?

.eslintrc.json
mike-robertson: lol yep

components/Table/PlayerMatchesTable.container.jsx
mike-robertson: I think this is a preferred lifecycle method since I don't think it's called when the component is first loaded

reducers/gotPlayer/player.js
mike-robertson: Yes but I'd like to have better encapsulation of state so if we update the reducers we don't have to do any refactoring. Just trying this pattern out for now to see how it feels, cause it doesn't feel good that the containers need to know the structure of the store

components/Player/Player.jsx
mike-robertson: withRouter just passes any router props down to it, it's just a different way of doing it. Not sure which is better but want to try to avoid passing props down a lot of components if we can avoid it (just learned about this method)

reducers/gotPlayer/player.js
howardchung: I don't like how we basically have to redeclare all of the props here.

If we add a prop in the API, it's a lot of code to add just to get it to show up in the UI.

With Jade we could start editing the template immediately and just reference the new prop.

mike-robertson: if you change a prop you have to change it in every template as well

howardchung: In this case we have to add it here and add/change the component.  That's two places

PR #23, opened 2016-06-06
Bug fixes, minor style changes. PlayerMatches table now pulls data from the player api calls.

mike-robertson:


components/Player/PlayerHeader.css
```diff
@@ -24,10 +24,6 @@
   composes: image;
```
howardchung: I think you need to `rm -rf` your build directory to avoid adding the build files.

transformations/transformations.js
```diff
@@ -2,7 +2,7 @@ import { HOST_URL } from '../yasp.config';
 import moment from 'moment';
 
 // TODO move to utility module
-const isRadiant = (p) => p.player_slot < 128;
+const isRadiant = (playerSlot) => playerSlot < 128;
```
howardchung: In order to allow passing in player_match objects directly this function intentionally takes an object with `{player_slot: 128}`

mike-robertson: The way I changed this before I was passing the specific value, which meant there was an error with it. This fixes the error.

components/Player/PlayerHeader.css
mike-robertson: I'm so confused why this thing keeps adding that. I'll just do the rmrf before each push now I guess

transformations/transformations.js
howardchung: for consistency can you wrap the value in an object?

PR #24, opened 2016-06-06
[WIP] Feature search page

bguzryanto:
WIP, don't merge yet.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{js,jsx}]
+charset = utf-8
+indent_style = space
+indent_size = 2
--- a/components/Search/Search.jsx
+++ b/components/Search/Search.jsx
@@ -1,0 +1,38 @@
+import React, { Component } from 'react';
+
+import { connect } from 'react-redux';
+import { REDUCER_KEY } from '../../reducers';
+import { getSearchResult } from '../../actions';
+import createConstantsWrapper from '../Constants';
+
+import SearchForm from './SearchForm';
+import SearchResult from './SearchResult';
+
+const mapStateToProps = (state) => {
+  const { error, loading, searchResults } = state[REDUCER_KEY].gotSearch;
+
+  return {
+    loading,
+    error,
+    data: searchResults,
+  }
+};
+
+const mapDispatchToProps = (dispatch) => ({
+    dispatchSearch: (query) => dispatch(getSearchResult(query)),
+});
+class Search extends Component {
+
+    render() {
+      const { dispatchSearch, data } = this.props
+
+        return (<div>
+            <SearchForm 
+              hintText="Search by name or SteamID" 
+              onSubmit={(value) => dispatchSearch(value)}/>
+            <SearchResult players={data} />
+        </div>)
+    }
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(Search)
--- a/components/Search/SearchForm.js
+++ b/components/Search/SearchForm.js
@@ -1,0 +1,50 @@
+import React, { Component } from 'react';
+
+import TextField from 'material-ui/TextField';
+import RaisedButton from 'material-ui/RaisedButton';
+import ActionSearch from 'material-ui/svg-icons/action/search';
+
+class SearchForm extends Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { hintText, onSubmit, query } = this.props;
+
+    const onEnter = (e) => {
+      
+      if (e.keyCode !== 13)
+        return;
+
+      onSubmit(this.textField.input.value)
+    }
+
+    return (
+      <div>
+        <TextField 
+          name="search" 
+          ref={(ref) => this.textField = ref} 
+          onKeyDown={onEnter} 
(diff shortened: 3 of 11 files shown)


.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
```
howardchung: should this be committed?

components/Search/Search.jsx
```diff
@@ -0,0 +1,38 @@
...
+    dispatchSearch: (query) => dispatch(getSearchResult(query)),
+});
+class Search extends Component {
+
+    render() {
+      const { dispatchSearch, data } = this.props
+
+        return (<div>
+            <SearchForm 
+              hintText="Search by name or SteamID"
```
howardchung: single quotes for strings.  Also should refactor this out to string

components/Search/SearchForm.js
```diff
@@ -0,0 +1,50 @@
...
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const { hintText, onSubmit, query } = this.props;
+
+    const onEnter = (e) => {
+      
+      if (e.keyCode !== 13)
```
howardchung: brackets, also @mike-robertson is this the recommended way to do search forms in react?

gu3st: I think using onEnter to submit, rather than wrapping it in a form element and using onSubmit is a bit strange. 

Only time I think to use something like onEnter is if you're using textareas as text inputs, as enters are a carriage return

bguzryanto: @howardchung currently I do some refactoring to make this pure func. agree with @gu3st suggestion to wrap it on `form` element.

.editorconfig
bguzryanto: yea, I used to use editorconfig for my project, this will setting my text editor to set 2 spaces and newline.

components/Search/Search.jsx
bguzryanto: it pass linter. I think it should use double quotes unless it is variable

bguzryanto:
Some idea to improve search page

- Add information how many search results
- Add information if the result is zero
- Add loading

bguzryanto:
need review @howardchung @gu3st @mike-robertson

components/Search/Search.jsx
```diff
@@ -0,0 +1,36 @@
...
+import SearchResult from './SearchResult';
+
+const Search = ({ dispatchSearch, data, loading, done }) => (
+  <div>
+    <SearchForm
+      hintText="Search by name or SteamID" // TODO replace hardcoded with tooltips.json from yasp core
+      disabled={loading}
+      onSubmit={(value) => dispatchSearch(value)}
+    />
+    {done === true ? <SearchResult players={data} /> : ''}
```
mike-robertson: You don't need done === true, just done as js will coerce to a truthy value

components/Search/SearchForm.jsx
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from 'react';
+
+import TextField from 'material-ui/TextField';
```
mike-robertson: Can we make this a stateless component? Since you're not using any life cycle methods it should be fine. You will need to slightly modify the refs code for that to work, if it's too much of a pain it's ok not to.

components/Search/SearchForm.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+SearchForm.propTypes = {
+  onSubmit: React.PropTypes.func,
+  hintText: React.PropTypes.string,
+};
+SearchForm.defaultProps = {
+  onSubmit: () => {},
+  hintText: '',
+};
+
```
mike-robertson: Do we need default values? I assume the Function will throw an error if it's undefined on submit but undefined string should be OK.

components/Search/Search.jsx
mike-robertson: Is search always players result? If not, we should rename to something ambiguous instead of players.

mike-robertson:
Reviewed, most of the stuff looks good for me, mostly minor comments. Not even sure if they need changes, just style things mostly. I would prefer to use stateless components where we can since at some point a new react version will optimize rendering for stateless components.

components/Search/SearchForm.jsx
bguzryanto: I was so lazy to make this stateless component, because I need to get rid of refs yesterday, let me try.

components/Search/SearchForm.jsx
bguzryanto: yea I think we do not need that. I will omit propTypes too

components/Search/Search.jsx
bguzryanto: yes, search result always return players see https://github.com/yasp-dota/yasp/blob/master/routes/api.js#L163

bguzryanto:
@mike-robertson ! changed to stateless component for `SearchForm.jsx` and omit ===

mike-robertson:
I think it's good to merge once you add in master and get rid of merge conflicts, unless Howard has some other stuff he'd like to see changed.

bguzryanto:
updated to solve merge conflicts.

components/Search/Search.jsx
mike-robertson: text in html is supposed to use double quotes if you're curious

PR #26, opened 2016-06-08
Mike | match table is working now

mike-robertson:


components/Table/MatchTable.container.jsx
```diff
@@ -0,0 +1,31 @@
...
+    sortState,
+    sortField,
+    // important to set the columns here since we don't have wrapper anymore
+    columns: overviewColumns,
+  };
+};
+
+const mapDispatchToProps = (dispatch) => ({
+  sortClick: (field, sortState, sortFn) => dispatch(setMatchSort(field, sortState, sortFn)),
+  getMatch: (playerId, numMatches, host) => dispatch(getMatch(playerId, numMatches, host)),
```
howardchung: Why are we getting match here?  Doesn't that mean one API request per table?

selectors/transformMatch.js
```diff
@@ -0,0 +1,22 @@
+import { createSelector } from 'reselect';
+import { REDUCER_KEY } from '../reducers';
+import transform from '../transformations';
+
+const getMatch = (state) => state[REDUCER_KEY].gotMatch.match;
+const getConstants = (state) => state[REDUCER_KEY].gotConstants;
```
howardchung: are we going to static constants?

transformations/transformations.js
```diff
@@ -26,14 +26,25 @@ function formatSeconds(input) {
...
+const transformation = {
   hero_id: ({ field, constants }) => `${HOST_URL}${constants.heroes[field] ? constants.heroes[field].img : field}`,
   radiant_win: ({ field, match }) => (isRadiant({ playerSlot: match.player_slot }) === field ? 'W' : 'L'),
   game_mode: ({ field, constants }) => (constants.game_mode[field] ? constants.game_mode[field].name : field),
   start_time: ({ field }) => moment.unix(field).fromNow(),
   duration: ({ field }) => formatSeconds(field),
 };
 
+const transformMatchItem = ({ field, constants }) => {
+  if (field === 0) {
```
howardchung: what is this handling?

components/Table/MatchTable.container.jsx
mike-robertson: No, we are just pulling the state from the reducer. It won't make an api call

transformations/transformations.js
mike-robertson: This is a function we pass into the item_n transformations. So the items that are displayed in match table are item_0, item_1, ... , item_5 (for the 6 slots). That item number is the number in the item table in constants. However, if there is no item, the value is 0. So I am giving the field.display a false value and returning so I don't try to look up that item in the constants and also don't display that item's picture.

howardchung: maybe add a comment explaining that

selectors/transformMatch.js
howardchung: It's not merged, it's part of my ab_upgrades PR.

components/Table/MatchTable.container.jsx
howardchung: OK, sounds good

transformations/transformations.js
mike-robertson: Good idea, I'll do that and merge after

Issue #30, opened 2016-06-09
drawer should dismiss when page is clicked


howardchung:
Currently, the user has to click the triple-bar icon in order to close the drawer.

Issue #29, opened 2016-06-09
remove alphabetization in drawer items


howardchung:
Allow us to explicitly control order to prioritize important entries.

PR #32, opened 2016-06-09
Mike

mike-robertson:
There are some changes regarding routing, they shouldn't break anything that is already happening (just some middleware we will use to fire route changes programmatically through actions)


howardchung:
Tag the PR with the issues you are fixing.

mike-robertson:
resolve #29 resolve #30

PR #34, opened 2016-06-20
Mike | added tabbar component. Takes a prop called tabs which is an a…

mike-robertson:
…rray of objects with 'label' and 'route' values. The 'label' is what is displayed on the tab. the 'route' is where the react-router routes to when the tab is clicked. Player.jsx shows how to display the 'subroutes' inside the component


components/Player/Player.jsx
```diff
@@ -2,41 +2,60 @@ import React from 'react';
...
 const mapDispatchToProps = (dispatch) => ({
   getPlayer: (playerId) => dispatch(getPlayer(playerId)),
+  getPlayerMatches: (playerId, numMatches) => dispatch(getPlayerMatches(playerId, numMatches)),
 });
 
 class RequestLayer extends React.Component {
   componentDidMount() {
     this.props.getPlayer(this.props.playerId);
+    if (this.props.info === 'matches') {
+      this.props.getPlayerMatches(this.props.playerId, 50);
```
howardchung: why do we need 50 matches?

mike-robertson: we don't, I was just playing around with the values.

actions/playerActions.js
```diff
@@ -29,15 +30,18 @@ const getPlayerError = (payload) => ({
 export const getPlayer = (accountId, host = HOST_URL) => (dispatch) => {
   dispatch(getPlayerRequest());
   dispatch(getPlayerMatchesRequest());
+  dispatch(getPlayerHeroesRequest());
```
howardchung: is this making another request?  The /players endpoint is supposed to return both matches and heroes.

mike-robertson: no it doesn't actually make a request, it just tells that piece of state that it's in a loading state (because player request is happening here).

components/Match/Match.jsx
```diff
@@ -1,13 +1,25 @@
...
+  players,
+  state => transformAbilityUpgrades(state),
+  setMatchSort
+);
 
 const Match = () => (
   <div>
     <MatchTable columns={overviewColumns} />
-    <MatchTable columns={abUpgradeColumns} />
+    <AbilityUpgradesTable columns={abUpgradeColumns} />
```
howardchung: I would prefer not to have to declare a new name for each of the 15+ tables we're going to have

components/Player/PlayerHeader.jsx
```diff
@@ -3,13 +3,29 @@ import PlayerName from './PlayerName';
...
 import styles from './PlayerHeader.css';
+import TabBar from '../TabBar';
 
-export default () => (
-  <div className={styles.container}>
-    <PlayerName />
-    <div className={styles.playerInfo}>
-      <PlayerRecord />
-      <PlayerMMR />
+// TODO - once I have internet, maybe we already have this info from constats
```
howardchung: Yes, this should be built from json/player_pages.json

components/Table/columnDefinitions/playerHeroesColumns.jsx
```diff
@@ -0,0 +1,43 @@
...
+  sortFn: defaultSort,
+}, {
+  displayName: 'Win %',
+  field: 'win',
+  width: 2,
+  displayFn: ({ field, row }) => (
+    <div className={styles.percentContainer}>
+      <span className={styles.textContainer}>{getPercentWin(field.display, row.games.display).toFixed(1)}</span>
+      <span>
+        <svg viewBox="0 0 32 32" className={styles.pieChart}>
```
howardchung: you may want to refactor this into its own component.  We will likely use it a lot since there are lots of tables with win % columns.

mike-robertson: Yea I was thinking the same thing, I'll do that at some point tomorrow

PR #39, opened 2016-06-25
Header changes and player reducer change

mike-robertson:


howardchung:
looks ok

mike-robertson:
Not normalizing the reducers yet as api is in flux and want these tab changes live. May need to revert the profile picture instead of profile word to go to logged in player profile.

PR #41, opened 2016-06-29
wip building map

howardchung:
early WIP.

@mike-robertson is it possible to pass the match object to the BuildingMap component so it can be a "dumb" component and render itself based on passed props?


mike-robertson:
yes that should definitely be doable. One other thing that you could do (once I have completed reducer restructuring) is pass an id of the match to the component, and then it could figure out which match it should retrieve from the store based on that. Regardless, if you want to just start with the pure dumb component, it's a pretty good pattern. We don't have to put the 'container' logic inside the same file even, and in some ways it's better not to do that.

components/BuildingMap/BuildingMap.jsx
```diff
@@ -0,0 +1,207 @@
...
+    HOST_URL
+}
+from '../../yasp.config';
+import
+{
+    pad
+}
+from '../../util/utility';
+const buildingData = [
+    {
```
mike-robertson: can we move this buildingData to its own file as a module and import it so our component doesn't have to have that living in it directly?

components/BuildingMap/BuildingMap.jsx
```diff
@@ -0,0 +1,207 @@
...
+import
+{
+    HOST_URL
+}
+from '../../yasp.config';
+import
+{
+    pad
+}
+from '../../util/utility';
```
mike-robertson: please do imports like 
```js
import { HOST_URL } from '../../yasp.config';
// or
import {
  HOST_URL,
  otherThing,
  moreStuff,
} from '../myModule';
```

components/BuildingMap/BuildingMap.jsx
```diff
@@ -0,0 +1,207 @@
...
+    {
+        id: "brtd",
+        style: "position: absolute; top: 12%; left: 72%;"
+        },
+    {
+        id: "bmtd",
+        style: "position: absolute; top: 14%; left: 72%;"
+        },
+    {
+        id: "ar",
```
mike-robertson: Make this a stateless component, i.e., 
```js
export default ({ prop1, prop2, prop3 }) => {
  // logic here
  return (
    <div>Your Stuff Here</div>
  );
};
```

PR #42, opened 2016-06-29
Fix W/L not working with new API

Chris92de:
See commit descriptions. Works, but increases load time considerably, not sure what exactly the issue is.
npm run build fails because of some unused vars and some indents, no newline at the end of one file and spaces in a parenthesis where there shouldn't be. 

take a look if you want to merge it this way or not.


actions/playerHeroesActions.js
```diff
@@ -38,6 +38,6 @@ export const getPlayerHeroes = (playerId, numMatches, host = HOST_URL) => (dispa
   dispatch(getPlayerHeroesRequest());
   return fetch(`${host}${url(playerId, numMatches)}`, { credentials: 'include' })
     .then(response => response.json())
-    .then(json => dispatch(getPlayerHeroesOk(json.matches)))
+    .then(json => dispatch(getPlayerHeroesOk(json.slice(0, 10))))
     .catch(error => dispatch(getPlayerHeroesError(error)));
```
mike-robertson: We shouldn't be slicing this here, if they want more hero data, then we have to make a separate request to retrieve the data we just threw away here. It's better to just limit the number of heroes we return to that specific table. I'm not sure if that functionality is built in yet though. Just something to keep in mind

reducers/gotPlayer/player.js
```diff
@@ -39,8 +39,8 @@ export const player = {
   getProfile: state => player.getPlayer(state).profile,
   getPlayerName: state => player.getPlayer(state).profile.personaname,
   getLastLogin: state => player.getPlayer(state).profile.last_login,
-  getWins: state => player.getPlayer(state).win,
-  getLosses: state => player.getPlayer(state).lose,
+  getWins: state => state.yaspReducer.gotPlayer.winloss.win,
+  getLosses: state => state.yaspReducer.gotPlayer.winloss.lose,
   getMmrEstimate: state => player.getPlayer(state).mmr_estimate,
```
mike-robertson: Actually, let's just leave it like this for now. It's sort of weird to import wins/losses here, this should be responsible for all info associated player I think.

mike-robertson:
I just pushed all these changes into the new_api branch so we can kill this pr and have it all go through that pr

Chris92de:
Sounds good to me :)
Am 29.06.2016 20:09 schrieb "mike-robertson" <notifications@github.com>:

PR #43, opened 2016-06-30
Mike

mike-robertson:
Restructur the player reducer so that it is like a database. This means that we are now storing 2 things there:

allIds: An array of all the player account_id's

byId: A map of players indexed by their id

ex:
```js
allIds = [1234, 4324, 8583];
byId = {
  1234: {
    //normal player data in here
  },
  ...
}
```
This allows us to 1, store multiple players data 2, persist data so that we don't need a request every time we load that player page, and 3, make constant time accesses to the player since it is a map. This is the groundwork for what we will do to a lot of the data (matches, heroes, etc. etc.) so that we don't need to make lots of requests. If the user does want to refresh data, they can refresh the page or we can provide a refresh button.

### DIFF
--- a/reducers/gotPlayer/player.js
+++ b/reducers/gotPlayer/player.js
@@ -50,6 +50,9 @@ const byId = (state = {}, action) => {
 const allIds = (state = [], action) => {
   switch (action.type) {
     case playerActions.OK:
+      if (state.includes(action.id)) {
+        return state;
+      }
       return [...state, action.id];
     default:
       return state;
--- a/reducers/gotPlayer/player.js
+++ b/reducers/gotPlayer/player.js
@@ -1,13 +1,5 @@
 import { playerActions } from '../../actions';
-// const byId = (state = {}, action) => {
-//   switch (action.type) {
-//     case expression:
-//
-//       break;
-//     default:
-//
-//   }
-// };
+import { combineReducers } from 'redux';
 
 const initialState = {
   loading: true,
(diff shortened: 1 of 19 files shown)


reducers/gotPlayer/player.js
```diff
@@ -1,57 +1,96 @@
...
-  getLastLogin: state => getPlayer.getPlayer(state).profile.last_login,
-  getMmrEstimate: state => getPlayer.getPlayer(state).mmr_estimate,
-  getSoloMmrEstimate: state => getPlayer.getPlayer(state).solo_competitive_rank,
-  getCompetitiveRank: state => getPlayer.getPlayer(state).competitive_rank,
-  getPicture: state => getPlayer.getProfile(state).avatarmedium,
-  getSteamLink: state => getPlayer.getProfile(state).profileurl,
-  getCheese: state => getPlayer.getProfile(state).cheese,
+  getPlayerById: (state, id) => {
+    // this feels like a hack
+    if (!state.yaspReducer.gotPlayer.playerReducer.byId[id]) {
```
howardchung: explain why it's a hack.

Also, isn't it unsafe to access deeply nested properties like this?  If any of them are undefined, the application will crash.

mike-robertson: I don't think it necessarily is a hack, it just doesn't feel 100% right if that makes sense. 

Normally that would be true but we define these properties in the initial state so it would only ever be undefined if something set them to undefined (if that player doesn't exist, we get the initial States set in the reducer as their state, and all those properties are always defined).

Issue #50, opened 2016-07-03
Win/Loss in player matches tab is wrong


mike-robertson:
the win/loss column in the player matches table is wrong

PR #51, opened 2016-07-03
Mike | fix w/l bug, increase to 20 matches

mike-robertson:
resolved issue #50

### DIFF
--- a/components/Player/Player.jsx
+++ b/components/Player/Player.jsx
@@ -41,7 +41,7 @@ const getOverviewTab = playerId => (
       <div className={styles.heroesContainer}>
         <Text className={styles.tableHeading}>HERO STATS</Text>
         <Card className={styles.card}>
-          <PlayerHeroesTable columns={playerHeroesOverviewColumns} id={playerId} numRows={10} />
+          <PlayerHeroesTable columns={playerHeroesOverviewColumns} id={playerId} numRows={20} />
         </Card>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


components/Player/Player.jsx
```diff
@@ -88,15 +88,15 @@ const mapDispatchToProps = (dispatch) => ({
 class RequestLayer extends React.Component {
   componentDidMount() {
     this.props.getPlayer(this.props.playerId);
```
howardchung: at some points, should probably refactor this and the very similar code below to deduplicate.

util/utility.js
```diff
@@ -131,7 +131,7 @@ function buildMap() {
     */
 }
 
-const isRadiant = (p) => p.player_slot < 128;
+const isRadiant = (playerSlot) => playerSlot < 128;
```
howardchung: I mentioned before that the server side implementation of this function works on an object.  It's fine if you want to change it, but keep it mind it may confuse developers who work on both sides.

transformations/transformations.js
```diff
@@ -5,7 +5,10 @@ const constants = require('../constants');
 
 const transformation = {
   hero_id: ({ field }) => constants.heroes[field],
-  radiant_win: ({ field, match }) => (isRadiant({ playerSlot: match.player_slot }) === field ? 'W' : 'L'),
+  radiant_win: ({ field, match }) => {
+    const isRadiantResult = isRadiant(match.player_slot);
+    return ((isRadiantResult && field) || (!isRadiantResult && !field) ? 'W' : 'L');
```
howardchung: If !field, the match is "no stats recorded".  Up to you if you want to make a specific case for this.

util/utility.js
mike-robertson: yea, it wasn't working right like that and I think that in general functions should only pass around the data they care about. In this case, we literally just care about checking the player slot, so there is no reason to pass around the entire match object. I think we can just be aware that all the real data from the server is snake_case, maybe we can put something in the readme.

transformations/transformations.js
mike-robertson: thanks, I'll add that in

PR #52, opened 2016-07-03
Mike | add the peers tab

mike-robertson:


components/Cheese/Cheese.jsx
```diff
@@ -24,7 +24,7 @@ const Cheese = ({ donations, error, loading }) => {
         <div>Buy some cheese. Help pay for servers. Reaching the goal every month keeps us running.</div>
       </div>
       <div className={styles.section}>
-        <div className="meter_wrapper">
+        <div>
           <h3 style={{ fontWeight: 700, marginTop: 0 }}>Monthly Cheese Goal</h3>
           <div>
             {error && <Error />}
```
howardchung: why is components/Player/Pages/index.js added?

components/Player/Player.jsx
```diff
@@ -83,21 +106,25 @@ const mapDispatchToProps = (dispatch) => ({
...
+  props.getPlayerPeers(props.playerId);
+};
+
 class RequestLayer extends React.Component {
   componentDidMount() {
-    this.props.getPlayer(this.props.playerId);
-    this.props.getPlayerMatches(this.props.playerId, 20);
-    this.props.getPlayerHeroes(this.props.playerId);
-    this.props.getPlayerWinLoss(this.props.playerId);
+    getData(this.props);
```
howardchung: will this preload data/make requests for peers data if any tab is loaded?

We want to only load tabs if players click on them.

components/Table/columnDefinitions/utility.jsx
```diff
@@ -0,0 +1,35 @@
...
+    return 0;
+  });
+};
+
+export const winPercentTransform = numGamesField => (list, field, property) =>
+  (list[numGamesField][property] ? list[field][property] / list[numGamesField][property] : 0);
+
+export const getPercentWin = (wins, games) => (games ? Math.ceil(1000 * (wins / games)) / 10 : 0);
+
+export const PercentComponent = ({ wins, games }) => (
```
howardchung: component should go in its own directory in components dir?

reducers/gotPlayer/peers.js
```diff
@@ -0,0 +1,99 @@
...
+        sortField: action.sortField,
+        sortFn: action.sortFn,
+      };
+    default:
+      return state;
+  }
+};
+
+
+const byId = (state = {}, action) => {
```
howardchung: do we have to set this up for every endpoint?  It feels very verbose, can we refactor it so we only do it once?

components/Player/Player.jsx
mike-robertson: yea that's what the pages is for, I am going to put all the pages we don't want to preload into there as their own components. I'll push that change up today.

components/Table/columnDefinitions/utility.jsx
mike-robertson: yea probably

reducers/gotPlayer/peers.js
mike-robertson: I think that I have found a pattern for the reducers which should be the same for everything. So what we can do is factor out a lot of the shared logic and just have some sort of factory function. I'm pretty sure it's possible since changing it is a lot of copy pasting. We can probalbyd o the same for some of the actions and selectors as well, though it may not be worth it for some of the duplication (calling the factory will be just as much code as not calling it).

transformations/transformations.js
```diff
@@ -7,6 +7,9 @@ const transformation = {
   hero_id: ({ field }) => constants.heroes[field],
   radiant_win: ({ field, match }) => {
     const isRadiantResult = isRadiant(match.player_slot);
+    if (field === undefined) {
+      return 'N/A';
```
howardchung: Maybe "N" with gray text.

Then W with green, and L with red, like how we currently do it.

PR #65, opened 2016-07-07
Local navbar

howardchung:
Define navbar elements in SPA instead of reading from store/constants

### DIFF
--- a/components/NavBar/NavBar.jsx
+++ b/components/NavBar/NavBar.jsx
@@ -1,55 +1,50 @@
 import React from 'react';
-import { connect } from 'react-redux';
-import Spinner from '../Spinner';
-import Error from '../Error';
 import { Link } from 'react-router';
 import Divider from 'material-ui/Divider';
 import styles from './NavBar.css';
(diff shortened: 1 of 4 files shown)


components/NavBar/NavBar.jsx
```diff
@@ -8,28 +8,55 @@ import styles from './NavBar.css';
...
+};
 
-const NavBar = ({ loading, error, links = navLinks, toggleMenu }) => {
-  const getNavLinks = (navLinks) =>
-    navLinks.map((link, index) => (
-      <Link key={index} to={link.path} onTouchTap={toggleMenu}>
+export default function NavBar({ toggleMenu })
+{
+    const links = Object.keys(pages).map((link, index) => (
+      <Link key={index} to={pages[link].path} onTouchTap={toggleMenu}>
```
mike-robertson: Can we just make this an array and iterate over that? There's no reason we need to make the thing an object

mike-robertson: Also can we put this in its own module? It's another hard coded configuration object

PR #67, opened 2016-07-09
icons

### LINKED ISSUE #54: login status section should use consistent text
Use "profile" instead of image, since image is not available unless the player profile has been loaded.

### LINKED ISSUE #66: refactor to use bundled material icons
Currently the index.html references an external dependency:
```
    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
```

This should be bundled in with webpack so that the site functions if Google changes the route or stops hosting the icons.

### LINKED ISSUE #69: Package.json needs to have dependencies properly sorted
dependencies and devDependencies are 2 different things and they need to be separated. devDependencies are things that are required for dev/build process (webpack, webpack loaders etc.). dependencies are things which the application requires to run (react, react-dom, etc.).

### LINKED ISSUE #73: use hashed file name in index.html
We should put the hash in the generated bundle name and inject the name to prevent GitHub pages from caching the bundle when we update.

```
var Path = require("path");
var FileSystem = require("fs");

var webpackConfig = {
    ...
    plugins: [
        ...
        function() {
            this.plugin("done", function(statsData) {
                var stats = statsData.toJson();

                if (!stats.errors.length) {
                    var htmlFileName = "index.html";
                    var html = FileSystem.readFileSync(Path.join(__dirname, htmlFileName), "utf8");

                    var htmlOutput = html.replace(
                        /<script\s+src=(["'])(.+?)bundle\.js\1/i,
                        "<script src=$1$2" + stats.assetsByChunkName.main[0] + "$1");

                    FileSystem.writeFileSync(
                        Path.join(__dirname, "%your build path%", htmlFileName),
                        htmlOutput);
                }
            });
        }
    ]
};
```

howardchung:
fixes #54 
fixes #66 
fixes #73 
fixes #69 

- [x] swap to using material icons


mike-robertson:
don't use font awesome, just use the material ui font icons library

components/Header/Header.jsx
```diff
@@ -141,7 +141,7 @@ export default ({
   openMenu,
   params,
   location,
-}) => (<div>
```
mike-robertson: Can we use css files when we can (so there is one source of truth), and can we space it like html instead of like curly brackets?

howardchung:
I'm not sure the material-icons library has everything we need.

components/Player/PlayerMMR.jsx
```diff
@@ -5,15 +5,37 @@ import { connect } from 'react-redux';
...
+        </span>
+        <span className={"rating"}>
+          <span>
+            <abbr title={"MMR estimate based on available data from peer players. This is an estimate of the population mean MMR of the recent matches played by this user."}>
+              <i className={"fa fa-fw fa-question-circle"}></i>
+            </abbr>
+          </span>
+          <small>{mmrEstimate.estimate}</small>
+        </span>
       </div>
```
mike-robertson: Do icons like I have in the past (using the material-ui <FontIcon> element), and don't use font-awesome (we don't need it as we are already using material icons). We can just load material icons into the project.

mike-robertson: Also those class names won't work with css-modules, and in general it's a bad pattern to use globally named spaced names like that (that's the whole reason to use css-modules, to avoid namespace collisions and have modular css). Also, why are we even changing this? If we are trying to update the site front end, we don't have to model everything the way it was in the old site.

components/Player/PlayerPicture.jsx
```diff
@@ -6,33 +6,23 @@ import Avatar from 'material-ui/Avatar';
...
       </Link>
     );
   }
-  return (
-    <a href={steamLink} className={getPlayerStyle(registered, cheese)}>
-      <Avatar src={picture} />
-    </a>
-  );
+  return <div />;
 };
```
mike-robertson: why are you changing perfectly working code?

howardchung:
Potential replacement icons:

soloMMR: 'person'
partyMMR: 'group'
estimate: 'help'
record: 'poll'

components/Player/PlayerRecord.jsx
```diff
@@ -5,15 +5,22 @@ import { connect } from 'react-redux';
...
+        <span className={"text-primary"}>
+          <i className={"fa fa-trophy"}></i>
+          <small>
+            <span className={"text-success"}>{wins}</span>
+            <span> - </span>
+            <span className={"text-danger"}>{losses}</span>
+            <span>{` (${(wins/(wins+losses)*100).toFixed(2)}%)`}</span>
+          </small>
+        </span>
       </div>
```
mike-robertson: Why are we changing working code? Also see the same issues as above. There's no reason we have to copy the old site exactly; I thought part of this process was to revamp the site to be better UI/UX design.

Also, there is no reason to make any of these 'dumb' components. They already are dumb components. That's the whole point of using react-redux, we can design dumb components and use the connect()() function to create smart components.

index.html
```diff
@@ -1,36 +1,15 @@
...
-    <link href="https://fonts.googleapis.com/icon?family=Material+Icons"
-      rel="stylesheet">
-    <link rel="manifest" href="/public/images/manifest.json">
-    <link rel="shortcut icon" href="/public/images/favicon.ico">
-    <meta name="msapplication-TileColor" content="#da532c">
-    <meta name="msapplication-TileImage" content="/public/images/mstile-144x144.png">
-    <meta name="msapplication-config" content="/public/images/browserconfig.xml">
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.3/css/font-awesome.min.css" rel="stylesheet">
     <meta name="theme-color" content="#ffffff"> </head>
```
mike-robertson: there's no reason to use 2 font icon libraries, just use material icons.

components/Player/PlayerPicture.jsx
howardchung: Simplifying.  I think the square picture looks better, and we don't need the other cases.

components/Player/PlayerMMR.jsx
howardchung: Not everything will match, but I think the current player page header is effective.

I can swap to using material icons

components/Player/PlayerPicture.jsx
mike-robertson: Ok, I don't really care either way about the square vs circle, but I don think that circlular profile pictures is something a lot of modern apps use (gchat/hangouts in particular). I think hard edges looks strange, and slightly rounded edges doesn't really look great either. I don't really care either way though I guess.

components/Player/PlayerRecord.jsx
howardchung: I changed the win/loss computation to what we are using now (more readable).

Shouldn't we avoid using connect() whenever possible?  If we can pass everything the component needs via props, then we can reduce boilerplate.  In this case, the parent component (PlayerHeader) can take care of store + dispatches and then just pass props to the children.

components/Player/PlayerMMR.jsx
mike-robertson: Ok that seems fine, though I am not a huge fan of tooltips (and the estimate requires a tooltip since it isn't fully intuitive).

components/Player/PlayerRecord.jsx
mike-robertson: There are some cases we wouldn't want to use connect.

1. We already have the information pulled into the parent (for example the Table component is being connected, but the table row isn't being connected)
2. We are certain we won't use that component anywhere else (where 1. wouldn't apply there).
3. We want to do some sort of presentational component such as PiePercent.

However, we don't pull any of this info into PlayerHeader (it's just accepting a playerId as a prop and rendering these child components with that playerId passed down), so we would have to move that mapStateToProps to there anyways, and we would have to pull in all the information individually because passing down the entire player object will mess up when components should be re-rendered as well as tightly couple those components to the structure of the player object (they should be dumb and not care about the state structure). If we are moving all mapStateToProps to the PlayerHeader, we aren't reducing boilerplate (unless you mean the very minimal 3 or so lines of declaring const mapStateToProps and the connect wrapping), and we are tightly coupling everything together. 

One of the major benefits of react-redux is that we can avoid passing lots of props down to children (or through several layers of children) which is much more confusing (having to trace through several components where things came from).

components/Player/PlayerMMR.jsx
howardchung: I don't know if there's a way to get around tooltips, since there's quite a bit of text explaining how the estimate works.

Also, I think we'll need tooltips for the items/abilities anyway.

mike-robertson: Yea we should just go with the tooltips idea unless we come up with a better one (we can always refactor). Best to stick with the current solution since what I came up with isn't all that great.

components/Player/PlayerMMR.jsx
```diff
@@ -4,16 +4,44 @@ import Spinner from '../Spinner';
...
+          <span>
+            <abbr title={"Solo MMR"}>
+              <SocialPerson color={blue500} />
+            </abbr>
+          </span>
+          <small>{soloRank}</small>
+        </span>
+        <span>
+          <span>
+            <abbr title={"Party MMR"}>
```
mike-robertson: Feels like there are too many spans in a row in here. I haven't looked closely but you might not need this many tags nested.

components/Player/PlayerPicture.jsx
```diff
@@ -1,59 +0,0 @@
...
-    };
-  }
-
-  return {
-    picture: player.getPicture(state, ownProps.playerId),
-    steamLink: player.getSteamLink(state, ownProps.playerId),
-    cheese: player.getCheese(state, ownProps.playerId),
-  };
-};
-
```
mike-robertson: Make sure that this component isn't used elsewhere. I think it is being used by the player pages but I might be wrong.

components/Player/PlayerRecord.jsx
```diff
@@ -3,17 +3,22 @@ import Error from '../Error';
...
     return (
-      <div className={styles.mmr}>
-        <div>{`${wins} - ${losses} (${calcWinPercent(wins, losses)}%)`}</div>
+      <div>
+        <span>
+          <small>
+            <span>{wins}</span>
+            <span> - </span>
+            <span>{losses}</span>
+            <span>{` (${(wins / (wins + losses) * 100).toFixed(2)}%)`}</span>
```
mike-robertson: I don't think we need so many different span tags unless it's for line breaking reasons, but you can use css for that.

components/Player/PlayerPicture.jsx
howardchung: It built okay, so I think it's fine

components/Player/PlayerRecord.jsx
howardchung: The reason originally is to style the win/loss counts with different colors.

webpack.config.js
```diff
@@ -84,7 +88,25 @@ if (isProd) {
     }),
     new webpack.DefinePlugin({
       'process.env': { NODE_ENV: JSON.stringify(env) }
-    })
+    }),
```
mike-robertson: I prefer arrows since they autobind this context to the function, and it can avoid weird error cases. I don't care that much though, just keep this context in mind since javascript this is different than other things in regards to function context.

howardchung: do you know how to arrowify it?  I copied this from a github issue :)

PR #71, opened 2016-07-10
Mike | add a numRows arg to be passed to the playerHERoes selector (t…

mike-robertson:
…ransformation) so that we can only return a certain number of rows from the global state. This lets us sort on that subset of data instead of the entire data set (but only showing the top 20 or whatever rows)

### DIFF
--- a/components/Player/Pages/Overview.jsx
+++ b/components/Player/Pages/Overview.jsx
@@ -24,7 +24,7 @@ const PlayerMatchesTable = createTable(
 );
 const PlayerHeroesTable = createTable(
   playerHeroes.getPlayerHeroesById,
-  (state, sortState, playerId) => (sortState ? sortPlayerHeroes(playerId)(state) : transformPlayerHeroesById(playerId)(state)),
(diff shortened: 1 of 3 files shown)


components/Player/Pages/Overview.jsx
```diff
@@ -24,7 +24,7 @@ const PlayerMatchesTable = createTable(
 );
 const PlayerHeroesTable = createTable(
   playerHeroes.getPlayerHeroesById,
-  (state, sortState, playerId) => (sortState ? sortPlayerHeroes(playerId)(state) : transformPlayerHeroesById(playerId)(state)),
+  (state, sortState, playerId) => (sortState ? sortPlayerHeroes(playerId, 20)(state) : transformPlayerHeroesById(playerId, 20)(state)),
```
howardchung: shouldn't sorting occur on the full data, not the limited data?  (e.g., if we have 100 rows but only display 20, we should sort on the full data set and render the top 20 rows, not just sort on the displayed 20 rows).

mike-robertson: It doesn't seem very useful for tables with limited number of rows displayed. Thid I'd only giving us the ability for this kind of behavior anyways. Once we add pagination we don't need to use this, but until then it feels weird to do the full data set (for example on heroes) when you sort and it's rows that you normally can't get to.

PR #72, opened 2016-07-12
Mike | basic filtering on tables, still wip (only works on overview matches atm)

mike-robertson:


actions/playerHeroesActions.js
```diff
@@ -1,8 +1,10 @@
 import fetch from 'isomorphic-fetch';
 import { HOST_URL } from './';
 import { playerHeroes } from '../reducers';
+import { getUrl, addQueryString, defaultOptions, getModifiedOptions } from './utility';
 
-const url = (playerId, queryString) => `/api/players/${playerId}/heroes?${queryString}`;
+const excludedOptions = ['limit'];
```
howardchung: why is this excluded?

actions/playerHeroesActions.js
```diff
@@ -38,13 +40,15 @@ export const getPlayerHeroesError = (payload, id) => ({
...
+export const getPlayerHeroes = (playerId, options = defaultOptions, host = HOST_URL) => (dispatch, getState) => {
   if (playerHeroes.isLoaded(getState(), playerId)) {
     dispatch(getPlayerHeroesOk(playerHeroes.getHeroList(getState(), playerId), playerId));
   } else {
     dispatch(getPlayerHeroesRequest(playerId));
   }
-  return fetch(`${host}${url(playerId, numHeroes)}`, { credentials: 'include' })
+  const modifiedOptions = getModifiedOptions(options, excludedOptions);
+
+  return fetch(`${host}${getUrl(playerId, addQueryString(modifiedOptions), url)}`, { credentials: 'include' })
```
howardchung: where is addQueryString defined?

actions/utility.js
```diff
@@ -0,0 +1,28 @@
+import { camelToSnake } from '../utility';
+
+export const getUrl = (playerId, options, url) =>
+  `${url(playerId)}${options.reduce((previous, current) =>
+    current.values.reduce((total, value) => `${previous}${current.queryParam}=${value}&`, ''), '?')}`;
```
howardchung: won't there be a trailing &?

actions/utility.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const addQueryString = options => Object.keys(options).map(key => ({
```
howardchung: consider using the querystring module to handle the URL manipulation?

https://www.npmjs.com/package/querystring

components/Explorer/Explorer.jsx
```diff
@@ -5,12 +5,9 @@ require('brace/theme/monokai');
 import fetch from 'isomorphic-fetch';
 import { HOST_URL } from '../../yasp.config';
 import RaisedButton from 'material-ui/RaisedButton';
+import TextField from 'material-ui/TextField';
```
howardchung: it looks like you made your edits on an older version of this file.

components/Form/FormField.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import AutoComplete from 'material-ui/AutoComplete';
+import ChipList from './ChipList';
+import { connect } from 'react-redux';
+import { addChip, setFieldText } from '../../actions';
+import { form } from '../../reducers';
+import { getClosestMatch } from './utility';
+import styles from './FormField.css';
+
+const onNewRequest = (value, index, formName, name, dataSourceConfig, dataSource, strict, addChip, setFieldText, limit) => {
```
howardchung: when you have this many params isn't it better to destructure and pass an object?

components/Form/TableFilter.config.js
```diff
@@ -0,0 +1,66 @@
...
+export const heroConfig = {
+  text: 'localized_name',
+  value: 'id',
+};
+export const genericConfig = {
+  text: 'text',
+  value: 'id',
+};
+
+// This has to be a list in order to pass it in.
```
howardchung: Just do it here :)  It's a one time operation.

components/Form/TableFilterForm.jsx
```diff
@@ -0,0 +1,133 @@
...
+import ClearButton from './ClearButton';
+import styles from './TableFilterForm.css';
+import * as data from './TableFilter.config';
+import { connect } from 'react-redux';
+import { submitForm, clearForm } from '../../actions';
+
+const TableFilterForm = ({ submitForm, clearForm }) => (
+  <Form name="tableFilter" className={styles.form}>
+    <FormField
+      name="withHeroId"
```
howardchung: shouldn't the name be snake_case to match the API params?  I think you have a conversion function somewhere, but isn't it more obvious to make it match directly?

components/Player/Player.jsx
```diff
@@ -38,6 +38,7 @@ const Player = ({ playerId, info }) => {
     </div>
   );
 };
+// need to fix this
```
howardchung: mark it with //TODO

actions/playerHeroesActions.js
mike-robertson: whenever I pass back the limit query param to heroes, it limits the data on heroes. Since heroes is an aggregation I assumed that we might not want that behavior. I might not want all my matches back on overview, for example, but I want all the hero data (if that makes sense). I'm not sure what the usecase would be for limiting the hero data except if we were able to sort serverside the matches returned (top 100 games sorted by kills or something).

actions/utility.js
mike-robertson: Yea that's true. I can fix that, though it doesn't seem to cause any issues

components/Explorer/Explorer.jsx
mike-robertson: i just merged from master. are there updates on a different branch I should pull in?

components/Form/FormField.jsx
mike-robertson: yea this sickens me. There are several options I can see here.

1. Keep as is. Hard to read the method signature and easy to mess up.
2. Pass an object. Still hard to read but can't mess up signature. Incurs object creation overhead (minor)
3. Put the function into the stateless component and don't pass anything into the component. No arguments then (inheriting from function scope), but confusing to understand where all the variables come from.
4. Hopefully something magically better than these three that you or I think up.

I will convert this to the destructuring for now but, let's try to find a better way to do this.

components/Form/TableFilter.config.js
mike-robertson: yea it's not really that bad 😄

components/Player/Player.jsx
mike-robertson: I meant to remove this comment before pushing since it was just something I needed to resolve before pushing. Whoops

actions/playerHeroesActions.js
howardchung: It should probably be user's choice.  Maybe they want the heroes played in their last 100 games or something

components/Explorer/Explorer.jsx
howardchung: Not sure, but Spinner, Popover, etc. should be there and this diff shows them being removed.

actions/playerHeroesActions.js
mike-robertson: yea that makes sense, i'll add it back in. The limiting is just by last played right now, right (since there is no sorting on api)?

actions/utility.js
mike-robertson: I don't have a lot of experience with node so I can take a look into adding that

components/Explorer/Explorer.jsx
mike-robertson: if those were unused then they would have caused a linting error so I would have removed them. Or are they in use in master at the moment?

components/Explorer/Explorer.jsx
```diff
@@ -25,137 +22,126 @@ function jsonResponse(response) {
...
-          >
-            <Menu onChange={this.handleExampleChange}>
-              {Object.keys(queries).map(function example(k) {
-                const e = queries[k];
-                return <MenuItem value={k} primaryText={e.name} />;
-              })
-              }
-            </Menu>
-          </Popover>
+      <div style={{ display: 'flex' }}>
```
howardchung: Yes, the left side is in use.  See all the removals referencing Popover, etc.

actions/utility.js
howardchung: Since we are on the client side, you'd probably have to npm install it as a dependency.

mike-robertson: I think we can add this but it would be a refactoring effort since our data structure isn't shaped like this { prop: value, prop2: [array, of, values] }. I can keep this module in mind though and see if there is an opportunity to refactor that at some point

PR #79, opened 2016-07-18
howard

### LINKED ISSUE #57: "last played" text on player/heroes
Move column to leftmost after hero column.  Replace "last match" with "last played"

### LINKED ISSUE #77: add skill/parse state column to matches table
skill = skill prop.  Number->string mappings are in constants.js.
parse state = version prop.  If truthy (not null and >0), match is parsed.

### LINKED ISSUE #78: add hero names to matches/heroes tables

howardchung:
fixes #78 
fixes #77 
fixes #57 

- [x] Need to either make skill a default column or add it to projection
- [x] need parse state column


actions/utility.js
```diff
@@ -1,20 +1,12 @@
...
-export const addQueryString = options => Object.keys(options).map(key => ({
-  ...options[key],
-  queryParam: key,
-}));
+export const getUrl = (playerId, options, url) => `${url(playerId)}?${querystring.stringify(options)}`;
 
 export const defaultOptions = {
-  limit: {
-    values: [
+  limit: [
```
howardchung: @mike-robertson why is the array wrapped in an additional property "value"?

mike-robertson:
I'm already using query string module in my pr

howardchung:
ok, when you merge I'll rebase

components/Explorer/Explorer.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import React from 'react';
```
mike-robertson: careful about using eslint-disable. Try to only use it for things that are still being worked on, but even then I would advise against (except in some edge cases where one is doing a bad pattern and needs to do refactor work before it can pass linting).

actions/playerMatchesActions.js
```diff
@@ -47,6 +47,7 @@ export const getPlayerMatches = (playerId, options = {}, host = HOST_URL) => (di
   } else {
     dispatch(getPlayerMatchesRequest(playerId));
   }
+  modifiedOptions.project = ['skill'].concat(modifiedOptions.project || []);
   return fetch(`${host}${getUrl(playerId, modifiedOptions, url)}`, { credentials: 'include' })
```
mike-robertson: I don't understand what's happening here. Could you add some comments or explain?

components/Explorer/Explorer.jsx
howardchung: It's costing me a lot of time to have to fix the linting on this and I'm still iterating on it rapidly.

actions/playerMatchesActions.js
howardchung: Skill is not one of the default columns right now.  I'm just adding it to the projection parameter.

components/Explorer/Explorer.jsx
mike-robertson: that's fine. I think the point of linting though is that eventually you don't need to 'fix' it ever, it's just the style you code in. I think it might help if you had linting in your editor/ide (for example, atom can display linting errors in the file in realtime). But don't worry about it too much right now.

PR #88, opened 2016-07-26
howard

### LINKED ISSUE #28: search page should update url
@bguzryanto 

I think the search page should update the URL in the address bar when a search is performed.  Any query string produced by the user should be reflected in the address bar to allow URL sharing.

### LINKED ISSUE #44: /login and /logout need to point to API_HOST
For now, the redirect will take the user back to the API host (which is current site).

### LINKED ISSUE #58: rename HOST_URL to API_HOST
more descriptive name

### LINKED ISSUE #76: nav drawer empty on match pages
not quite sure why.

### LINKED ISSUE #84: show N/A if MMR data not available

howardchung:
Not done yet, still figuring out search.

fixes #84 
fixes #28 
fixes #58 
fixes #76 
fixes #44


components/Search/SearchForm.jsx
```diff
@@ -4,44 +4,52 @@ import TextField from 'material-ui/TextField';
...
 
-const SearchForm = ({ hintText, onSubmit, disabled }) => {
-  let queryValue = '';
-
+const SearchForm = ({ location, dispatchSearch, dispatchSetQuery, hintText, onSubmit, disabled, query }) => {
   const formSubmit = (e) => {
     e.preventDefault();
-    onSubmit(queryValue);
+    browserHistory.push(`/search?q=${query}`); 
+    dispatchSearch(query);
```
mike-robertson: You can use the react-router router object to do this as well, it may be the more accepted pattern. See TabBar.jsx for an example of using it (importing withRouter and then wrapping your component with it which injects the router as a prop into your component). If you look for Programmatic Navigation in this website (https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md), it goes over the change. I'm not sure if there is a benefit to using it, but I would assume there is a reason they prescribe that method.

components/TabBar/TabBar.css
```diff
@@ -11,6 +11,6 @@
 
 .tab {
   font-weight: 600 !important;
-  font-size: 11 !important;
+  font-size: 12px !important;
   color: var(--primaryTextColor) !important;
 }
```
mike-robertson: Remember we only need !important for when we are overriding element inline styles (such as those applied to things by material-ui). This may be one such case, I didn't check, but we should try to avoid when possible.

mike-robertson:
changes look good to me so far

PR #91, opened 2016-07-31
styling updates

howardchung:


components/Match/Match.jsx
```diff
@@ -44,9 +50,38 @@ class RequestLayer extends React.Component {
...
+          <MatchTable columns={overviewColumns} />
+        </Card>
+        <Card>
+          <AbilityUpgradesTable columns={abUpgradeColumns} />
+        </Card>
+        <Card>
+          <BuildingMap match={this.props.match} loading={this.props.loading} />
+        </Card>
+        <MatchTable columns={benchmarksColumns(this.props.match)} />
       </div>
```
mike-robertson: you might want to consider doing something like the player pages components in the future once the match tabs are increased, then you can put the header in this matches component and similar to the player header.

components/Table/Table.jsx
```diff
@@ -36,13 +37,17 @@ const getTable = (data, columns, sortState, sortField, sortClick, numRows) => {
...
+              style={{ borderBottomWidth: '0px' }}
+            >
+              {columns.map((column, colIndex) => (
+                <MaterialTableRowColumn key={colIndex} style={getWidthStyle(column.width, totalWidth)}>
+                  {row && column.displayFn && column.displayFn({ row, column, field: row[column.field] })}
+                  {row && row[column.field] && !column.displayFn && (row[column.field].display)}
+                </MaterialTableRowColumn>
+              ))}
+            </MaterialTableRow>
           ))}
```
mike-robertson: Is there a reason we are decomponentizing the table?

components/Table/TableRowColumn.jsx
```diff
@@ -1,10 +0,0 @@
-import React from 'react';
-import { TableRowColumn as MaterialTableRowColumn } from 'material-ui/Table';
-import { getWidthStyle } from './tableHelpers';
-
-export default ({ row, column, totalWidth }) => (
-  <MaterialTableRowColumn style={getWidthStyle(column.width, totalWidth)}>
-    {row && column.displayFn && column.displayFn({ row, column, field: row[column.field] })}
-    {row && row[column.field] && !column.displayFn && (row[column.field].display)}
-  </MaterialTableRowColumn>
-);
```
mike-robertson: any reason we are doing this? Less maintainable in my opinion to create a huge table component, and the table is concerned with layout. ONly the table row column needs to care about data.

components/Table/Table.jsx
howardchung: * It was breaking the ability to pass the material-ui props down (workaround is to add them into the wrapper component, but that seems unnecessarily verbose)
* It's easier to follow what's going on

Did we need to componentize for a specific reason?

components/Table/TableRowColumn.jsx
howardchung: See the other comment

PR #92, opened 2016-08-03
Heroes features which include ranking and benchmark

bguzryanto:


components/Heroes/Benchmark.jsx
```diff
@@ -0,0 +1,51 @@
...
+    );
+  }
+}
+
+const mapStateToProps = (state) => ({
+  heroes: state[REDUCER_KEY].heroes,
+  heroId: state[REDUCER_KEY].gotBenchmark.hero_id,
+  result: state[REDUCER_KEY].gotBenchmark.result,
+  isLoading: state[REDUCER_KEY].gotBenchmark.loading,
+  isDone: state[REDUCER_KEY].gotBenchmark.done,
```
howardchung: is this prop needed?

components/Heroes/BenchmarkBadge.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Grid, Row, Col } from 'react-flexbox-grid';
```
howardchung: @mike-robertson should we adopt this for the rest of the components as well?  Current table don't show up well on mobile.

components/Heroes/BenchmarkTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+  TableHeader,
+  TableHeaderColumn,
+  TableRow,
+  TableRowColumn,
+} from 'material-ui/Table';
+
+import style from './heroes.css';
+
+
+const metrics = [
```
howardchung: This should be based on the server response, not defined in the client.

components/Heroes/BenchmarkTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+        return (
+          <TableRowColumn key={i} style={alignCenter} className={style.BenchmarkColNo}>
+            {parseFloat(percentile) * 100}%
+          </TableRowColumn>
+        );
+      }
+
+      return (
+        <TableRowColumn key={i} style={alignCenter}>
+          {roundFn(parseFloat(data[metric][id].value))}
```
howardchung: Just use toFixed()

components/Heroes/BenchmarkTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+const roundFn = (number) => Math.round(number * 100) / 100;
+const alignCenter = { textAlign: 'center' };
+
+const BenchmarkRow = ({ data, id, percentile }) => (
+  <TableRow selectable={false} className={style[`BenchmarkRow-${id % 2}`]}>
+    {metrics.map((metric, i) => {
+      if (metric === 'percentile') {
+        return (
+          <TableRowColumn key={i} style={alignCenter} className={style.BenchmarkColNo}>
+            {parseFloat(percentile) * 100}%
```
howardchung: tofixed()

components/Heroes/BenchmarkTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+      );
+    })}
+  </TableRow>
+);
+
+export default ({ data }) => (
+  <Table wrapperStyle={{ overflow: 'visible' }} className={style.BenchmarkTable}>
+    <TableHeader displaySelectAll={false} adjustForCheckbox={false}>
+      <TableRow>
+        <TableHeaderColumn tooltip="Percentile" style={alignCenter} className={style.BenchmarkColNo}>%</TableHeaderColumn>
```
howardchung: tooltips should already be defined in dotaconstants.  Reuse them here

components/Heroes/heroes.css
```diff
@@ -0,0 +1,79 @@
+.Header {
```
howardchung: I think our standard practice is to capitalize the names of CSS files

package.json
```diff
@@ -17,15 +17,18 @@
   },
   "dependencies": {
     "brace": "^0.8.0",
+    "classnames": "^2.2.5",
```
howardchung: what is this used for?

yaspv2.js
```diff
@@ -63,6 +64,11 @@ render(
           </Route>
         </Route>
         <Route path="explorer" component={Explorer} />
+        <Route path="heroes" component={Heroes}>
+          <IndexRoute component={HeroesIndex} />
+          <Route path="ranking/:hero_id" component={Ranking} />
```
howardchung: rankings

yaspv2.js
```diff
@@ -63,6 +64,11 @@ render(
...
+          <Route path="benchmark/:hero_id" component={Benchmark} />
```
howardchung: benchmarks

package.json
bguzryanto: classnames used by react-flexbox-grid, I tried to omit this dep from packages.json but Travis-CI build result a failed build -> https://travis-ci.org/yasp-dota/ui/builds/149625615

howardchung: Apparently this is an open issue:
https://github.com/roylee0704/react-flexbox-grid/issues/49

components/Heroes/Benchmark.jsx
bguzryanto: yes, `isDone` needed by this component, I was tried to use `isLoading`, the problem with `isLoading` once you load the page, first initial state is `false`, however `result` initial state is an empty object, checking whether object is empty or not cost more than using `isDone` and I don't want to check whether an object is empty or not because in my opinion it will make code slightly unpleasant to read.

components/Heroes/BenchmarkTable.jsx
bguzryanto: aw, my mistake, never use toFixed before.

components/Heroes/BenchmarkTable.jsx
bguzryanto: current yasp's implementation use something like humanize/(startCase in lodash) function to transform object props to humanize string. in the other hand not all tooltips present in dotaconstants such as 'Hero damage per minute', 'Last hits per minute', do I have to make PR for this?

components/Heroes/BenchmarkTable.jsx
bguzryanto: ok, will change it 👍

components/Heroes/BenchmarkTable.jsx
howardchung: Up to you.  In my opinion, the prop name are self-explanatory.  However, if we want to localize, we'll need to externalize the strings anyway, so might as well do it now.

You may have to add missing strings to tooltips.json and PR it.

components/Heroes/heroes.css
bguzryanto: then we need to add this standard to contribution page and rename every non capitalize css files, because not all css files have capitalize name in this project

howardchung: @mike-robertson , what do you think?  Is it important to be consistent?

bguzryanto: I think it's important to make convention for naming css files and components, because for this particular problem current css files have different naming convention.

package.json
bguzryanto: I don't know really well how NPM js dependencies works, but I think this problem can be solved by adding classnames and flexboxgrid in 'dependencies' instead only in 'devDependencies'. right?

components/Heroes/BenchmarkTable.jsx
bguzryanto: regarding dotaconstans are needed as sharing data between frontend and backend. if in the future we want to localize our strings, what is the best way to put externalized strings? in the dotaconstants or string file in the ui repo?

package.json
howardchung: I believe the maintainer of that library has to either remove the dependency, or move it in their package.json

components/Heroes/BenchmarkTable.jsx
howardchung: I think it only needs to be in the UI repo.  I don't see a use case where the server would need display strings.  

The reason they are in dotaconstants now is that the Jade templates (on server side) require the strings as well as the UI.  However we should be able to retire those with the SPA and therefore move tooltips.json from dotaconstants to here as well.

You may as well start a new string file for any strings that are UI-only, and we'll merge them with the ones from tooltips.json when we don't need them on the server side anymore.

bguzryanto: ok I will make new strings file in 'utility/language/en.js', do you think this is the appropriate way? or do we need use i18n/L10n?

howardchung: I'd put it as a root level dir (language/en.js).  Utility should be for code

package.json
bguzryanto: after reading stackoverflow answer and npmjs blog, I think it is because the maintainer use peerDep, because peerDep doesn't install peered dep when we run npm install, peerDep expect project which use 'react-flexbox-grid' to use babel-polyfill, classnames, flexboxgrid and react too.

howardchung: you might want to leave a note on that repo to see if they can do that (or submit the PR yourself)

howardchung:
If you want to get a little more detailed, we could have the server return the counts per "bucket" in each stat, and produce a histogram similar to the one we have now for MMR distribution.

actions/benchmarkActions.js
```diff
@@ -0,0 +1,40 @@
+import fetch from 'isomorphic-fetch';
+import { API_HOST } from './../actions';
+
+const url = '/api/benchmarks';
+
+const START = 'yasp/benchmark/START';
+const DONE = 'yasp/benchmark/DONE';
+const ERROR = 'yasp/benchmark/ERROR';
+
```
mike-robertson: I'd rather use similar naming schemes as compared to the other async actions (I think it's REQUEST, OK, and ERROR), but I don't think it's a big deal if you don't feel like refactoring that.

components/Heroes/BenchmarkTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+        <TableHeaderColumn tooltip="Hero healing per minute" style={alignCenter}>HHM</TableHeaderColumn>
+        <TableHeaderColumn tooltip="Tower damage per minute" style={alignCenter}>TDM</TableHeaderColumn>
+      </TableRow>
+    </TableHeader>
+    <TableBody displayRowCheckbox={false}>
+      {data.gold_per_min.map((gpm, idx) =>
+        <BenchmarkRow key={idx} data={data} id={idx} percentile={gpm.percentile} />)}
+    </TableBody>
+  </Table>
+);
```
mike-robertson: I'd rather we use our existing table component which has a lot of functionality such as sorting built in. It's fairly simple to set up as well. If there's something that our table isn't doing that you need we can look into that, but I don't see a reason to manually create your table like this (minus any sorting functionality). Plus, we are going to add more features to the table such as pagination, so this would be left out of that as well.

mike-robertson: Look at existing examples such as the /components/Player/pages/Matches.jsx for an example of how to use the table factory. You will also need to define columns in the /components/Table/columnDefinitions, and you would need to create some selectors similar to in the /selectors/transformPlayerMatches.js and /selectors/sortPlayerMatches.js. I think there is a wiki page on how to create the table. The table is also expecting certain reducer structure as well (see /reducers/gotMatch.js for an example).

components/Heroes/HeroesIndex.jsx
```diff
@@ -0,0 +1,19 @@
...
+const HeroesIndex = ({ heroes }) => (
+  <div>
+    <h1 className={style.Header}>All Dota 2 Heroes</h1>
+    <HeroList heroes={heroes} />
+  </div>
+);
+
+const mapStateToProps = (state) => ({
+  heroes: state[REDUCER_KEY].heroes,
+});
```
mike-robertson: I'd rather use the 'selector' syntax (doesn't need to be a reselect selector) in future components. So for example, instead of coupling your components to the shape of the state, have an exported object of functions which retrieve specific parts of the state base and accept state as a parameter. Look at /reducers/gotPlayer/matches.js for some examples.

components/Heroes/RankingTable.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import {
+  Table,
+  TableBody,
+  TableHeader,
+  TableHeaderColumn,
+  TableRow,
+  TableRowColumn,
+} from 'material-ui/Table';
+import Avatar from 'material-ui/Avatar';
```
mike-robertson: Same comment about the table as above

components/Heroes/BenchmarkBadge.jsx
mike-robertson: I'm not sure if I'm a huge fan of it but we could definitely do it. I don't particularly like the grid system because not everything divides well into 12's but I don't see why we couldn't do it where that isn't an issue.

components/Heroes/heroes.css
mike-robertson: Components and anything related to components directly (such as css), we have tried to capitalize. Other things can be lower case. I don't know if we have done that in every case though.

package.json
mike-robertson: Also classnames is a good package in general so we can make use of it in other areas.

components/Heroes/BenchmarkBadge.jsx
howardchung: I think we should either use it everywhere or nowhere.  If we can achieve the same effect without adding more dependencies then we shouldn't use it

reducers/gotRanking.js
```diff
@@ -0,0 +1,41 @@
...
+    case rankingActions.ERROR:
+      return {
+        ...state,
+        done: false,
+        error: true,
+        loading: false,
+      };
+    default:
+      return state;
+  }
```
mike-robertson: Here is where I would put selectors specific to state shape (so knowledge about the state shape is encapsulated in the reducer)

reducers/heroes.js
```diff
@@ -0,0 +1,18 @@
...
+// lazy evaluation
+// add API_HOST to hero img path so components
+// don't need to know API_HOST
+const transformHeroImagePath = (heroes) => {
+  const heroList = {};
+  Object.keys(heroes).forEach((key) => {
+    heroList[key] = { ...heroes[key] };
+    heroList[key].img = API_HOST + heroList[key].img;
+  });
+
```
mike-robertson: If you use the selector pattern described above you won't need to do this anymore.

components/Heroes/BenchmarkBadge.jsx
mike-robertson: We definitely can achieve a similar effect by just writing media queries for things that need to be responsive. It just depends on whether you want a strict grid system or not (which makes it easier to do breakpoints). However, you don't always need a different width for each breakpoint.

actions/benchmarkActions.js
bguzryanto: ok, I used to use START, DONE, ERROR as action constant on my own react project.

components/Heroes/BenchmarkTable.jsx
bguzryanto: I thought Table components were intended for Match and Player components, since there are columnDefinition explicitly define columns. I think if Table component has intend to be general components, we need to define columnDefinition inside Player and Match components

components/Heroes/heroes.css
bguzryanto: So we need to write this on the wiki and rename our css files

components/Heroes/HeroesIndex.jsx
bguzryanto: I understood what '/reducers/gotPlayer/matches.js' does. But I don't understand what do you mean by "accept state as parameter". Can you give me some example?

reducers/heroes.js
bguzryanto: As my current understanding selector used as to compute derivative data from state,
the reducer created because HeroesIndex needs list of all heroes, Rankings and Benchmarks too. in previous implementation I put this code to `connect'ed component HeroesIndex direct from dotaconstants, but this code make components harder to read. Other consideration for sake of good encapsulation, I don't want components to know where yasp data came from, so I put all data to state, and let component being presentational component as minimum as possible without knowing the source of data.

components/Heroes/BenchmarkBadge.jsx
bguzryanto: writing media queries need lots more of time than using react-flexbox-grid. iirc, my idea was because someone on discord has problem accessing data from mobile, so I did something diffrent to minimalize this problems by using flexbox. Also, react-flexbox-grid isn't for table components.

package.json
bguzryanto: @howardchung I make a comment on their issue, hopefully it will fix soon.

components/Heroes/BenchmarkBadge.jsx
mike-robertson: yea, I mean that library is just using media queries under the hood. We use flexbox, but some displays of data just won't 'squash down' properly or overflow to the next line properly. I'm cool with using this, I just don't like adding lots of markup to the component vs. just adding media queries in css.

reducers/heroes.js
mike-robertson: So a selector is just something which retrieves something from your state. It can be a derived thing (like the getVisibleTodos in the redux docs), but it can also just be a piece of your state. So the pattern I am trying to establish here is that in mapStateToProps, we never have to do something like:
```js
const mapStateToProps = state => ({
  something: state.deep.in.your.tree.something,
  somethingElse: state.other.place.thats.deep.somethingElse,
});
```
This is coupling the connected component to the state of your shape. Maybe someday, you want to flatten your state, or you want to refactor your state to be normalized (more like the reducers in gotPlayer/match.js or any of those). So the pattern would be to encapsulate that state shape knowledge in selectors located in the reducers. For example:
```js
// reducer.js

const initialState = {
  reducer: {
    prop1: false,
    prop2: true,
    list: [],
  }
};

// ...your reducer code goes here...

export const reducer = {
  getReducer: state => state.yaspReducer.reducer,
  getProp1: state => getReducer.getReducer(state).prop1,
  getProp2: state => getReducer.getReducer(state).prop2,
  getList: state => getReducer.getReducer(state).list,
};

// connectedComponent.js
import { reducer } from '../../reducers';
import Component from './Component';

const mapStateToProps = state => ({
  prop1: reducer.getProp1(state),
  list: reducer.getList(state),
});

export default connect(mapStateToProps)(Component);
```

So in this form, your component will never know what the state shape is, it will just be using essentially getters to access the state (passing in the state as a param to the getter). So instead of keeping presentational components unaware of the state shape, but coupling your connected components to the state shape (so if you refactor you have to change all connected components as well or you could get mysterious and hard to diagnose errors).

components/Heroes/HeroesIndex.jsx
mike-robertson: I put an example on your last comment.

components/Heroes/BenchmarkTable.jsx
mike-robertson: I'm not 100% sure I understand what you mean here. Do you mean that we should move the /Table/columnDefinitions directory and contents to their respective components (like playerMatchesColumns goes into /Player directory)? I think that is a good idea, but my one concern would be confusing people as to where they can find examples of table column definitions (since they wouldn't all be located in the Table directory). 

But yes, the Table component is supposed to be a generalized table.

components/Heroes/heroes.css
mike-robertson: so we don't want to capitalize all css files, just css files that are directly related to a component (i.e., being imported by that component). This is because components are capitalized. This convention is due to react requiring capitalized component names, so we just capitalize all component related files to be consistent. I think this should already be the case in our files though.

bguzryanto:
@howardchung what are counts per "bucket" mean?

howardchung:
we would set a min/max for each stat, like 0-1000 for GPM, then count the number of performances in each bucket (ex. 0-100, 101-200, etc.)

bguzryanto:
I think those stats could be computed in the client side, right?
I tried to prototype the histogram, do you think histogram like this? https://tonicdev.com/bagus/57a82e4f211cfb1300c36caf

howardchung:
The current response doesn't contain the necessary data.  It gives the points where a given % of players fall below, but not how many players are in a given bin.  Histogram bins should be equally sized.

bguzryanto:
ah I see, so this is more advance than current benchmarks page. yes, it could be great if we put those data to benchmarks API. later, we can design the histogram :)

reducers/heroes.js
bguzryanto: thank you to pointing out the selector pattern example. give me some time to refactoring my code :)

bguzryanto: thank you to pointing out the selector pattern example. give me some time to refactoring my code :)

components/Heroes/BenchmarkTable.jsx
bguzryanto: yes I mean we should move /Table/ColumnDefinitions directory and contents to their respective components, and put some comments to document our table components for new developers convenient

components/Heroes/Benchmark.jsx
bguzryanto: I found smarter idea to make this more readable.

bguzryanto:
I updated PR, things that not updated yet : 

1.  Use Table component for Benchmarks and Rankings Table (will do this once Table component generalized). I think is optional
2. Implement histogram

howardchung:
is this good to merge?

yaspv2.js
howardchung: @bguzryanto can you add the plurals?

bguzryanto:
@howardchung hi, I am busy with my academic things lately, I see this branch has conflicts. when Histogram already implemented, we can merge this PR.

howardchung:
I don't think we'll do the histogram for MVP (might be added as a later feature).

bguzryanto:
@howardchung ok, I will resolve this conflicts then we can merge this

bguzryanto:
@howardchung already tested, this is good to go :)

index.js
```diff
@@ -65,6 +66,11 @@ render(
           </Route>
         </Route>
         <Route path="explorer" component={Explorer} />
+        <Route path="heroes" component={Heroes}>
+          <IndexRoute component={HeroesIndex} />
+          <Route path="ranking/:hero_id" component={Ranking} />
```
howardchung: for parity with existing site I'd like this to be rankings and benchmarks.

I'll merge this in anyway, but it should be fixed (either you can submit a PR or I can patch it myself)

language/en/abbreviation.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable camelcase*/
```
howardchung: these should go into a string file.

howardchung:
couple of comments.

bguzryanto:
Updated @howardchung

PR #95, opened 2016-08-05
Adding records page (basic data) in table

mike-robertson:
I don't know where some of the aggregate data is coming from for this in the live site (doesn't get returned by api as far as I can tell), so right now I am just returning some basic stuff that is in the api response.


howardchung:
Yes the API does not match the object sent to Jade template currently.

Requires a server side change to count and return the data if we want it here

mike-robertson:
Yea, I figured it was something along those lines. I can do that if you point me to the file I should be modifying, I still don't have a great understanding of the code structure on the server side. Otherwise, we could just live without it, but it would be nice to display some of that stuff. I am not sure though what the games column means (in existing site). I assume that some of these stats weren't available or something on games of a certain age?

howardchung:
Yes it is the count of games with that stat available.  It is generally only available after we have the parser extract it.  We cant backfill either since replays expire.

See routes/api.js

howardchung:
looks good.  I think we can merge first and edit later if we decide we want more data (like match counts per record)

mike-robertson:
That makes sense, I'll merge now.

PR #96, opened 2016-08-06
initial request implementation

howardchung:
Requests page
Distributions page


howardchung:
any comments @mike-robertson ?

components/Distributions/Distributions.jsx
```diff
@@ -0,0 +1,104 @@
...
+              }
+              return value;
+            },
+          },
+        },
+      };
+      c3.generate(options);
+    }
+  }
+  render() {
```
mike-robertson: You're not using redux?

mike-robertson:
Only comment is you're no longer using redux. I'd prefer we stick to that for state management so we can have a Flux pattern and be consistent.

components/Distributions/Distributions.jsx
howardchung: We could move it later.  For now, though, this doesn't rely on any global state, and it's less code, so quicker to prototype.

mike-robertson: The point of redux isn't to use it when it relies on global state. It's to force a single source of truth for all state which is immutable from the components' perspective. I'm against merging this for speed reasons, it's not like these are hard use cases for redux, all our stuff has been this exact scenario, so you should be able to base it off existing patterns in the code.

howardchung: OK I'll make actions and reducers for these

howardchung:
@mike-robertson can you re-review?

yasp.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
-  API_HOST: 'https://yasp.co',
-  // API_HOST: 'http://hchung.westus.cloudapp.azure.com:5000',
+  // API_HOST: 'https://yasp.co',
```
howardchung: undo before checkin

actions/distributionsActions.js
```diff
@@ -0,0 +1,40 @@
+import fetch from 'isomorphic-fetch';
+import { API_HOST } from '../actions';
+
+const url = '/api/distributions';
+
+const START = 'yasp/distributions/START';
+const DONE = 'yasp/distributions/DONE';
+const ERROR = 'yasp/distributions/ERROR';
+
```
mike-robertson: we should be consistent with these action names. Can you name them similar to other async requests?

actions/requestActions.js
```diff
@@ -0,0 +1,71 @@
...
+  const formData = new FormData();
+  formData.append('match_id', matchId);
+  dispatch(requestStart());
+  return fetch(`${API_HOST}${url}`, {
+    method: 'post',
+    body: formData,
+  })
+  .then(res => res.json())
+  .then((json) => {
+    function poll() {
```
mike-robertson: don't define this function inside of your `then()`. Just declare it above the method and then call it with dispatch as an argument. This way we aren't defining this function every time you dispatch this action, and it's also easier to read.

mike-robertson:
Only comments are one where to define that polling function and naming consistency with actions for requests. Other than that I think it's good to go.

actions/distributionsActions.js
howardchung: ok, also renamed the search ones.  requestRequest is a bit weird, but i guess consistency is more important.

PR #98, opened 2016-08-06
start item timing overlay

aeosynth:
not clear on whether to use inline styles or add a class to column.css

the css could be tweaked (match dotabuff exactly?)

index.html changed during `npm run build`

### DIFF
--- a/components/Table/columnDefinitions/matchColumns.jsx
+++ b/components/Table/columnDefinitions/matchColumns.jsx
@@ -109,7 +109,6 @@ const overviewColumns = [
       for (let i = 0; i < 6; i++) {
         const item = constants.items[constants.item_ids[row[`item_${i}`]]];
         if (item) {
-          const time = row.first_purchase_time[constants.item_ids[item.id]];
           itemArray.push(<span
             key={i}
             style={{ position: 'relative' }}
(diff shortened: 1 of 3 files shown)


components/Table/columnDefinitions/matchColumns.jsx
```diff
@@ -107,13 +107,22 @@ const overviewColumns = [
...
-          />);
+            style={{ position: 'relative' }}
+          >
+            <img
+              style={{ height: 25, margin: '0 3px' }}
+              role="presentation"
+              src={`${API_HOST}${item.img}`}
+            />
+            <span className={styles.timing}>
+              {time / 60 | 0}
```
howardchung: why | 0 ?

aeosynth: it's `Math.floor`

components/Table/columnDefinitions/matchColumns.jsx
```diff
@@ -107,13 +107,22 @@ const overviewColumns = [
...
+            </span>
```
mike-robertson: I think you mean to put `time / 60 || 0`

index.html
```diff
@@ -14,6 +14,6 @@
   </head>
   <body>
     <div id="react"></div>
-    <script src="/build/bundle.js"></script>
+    <script src="/build/313f39d7e16424d48a25.bundle.js"></script>
   </body>
```
mike-robertson: don't change this, it will fail to work when we deploy

index.html
```diff
@@ -14,6 +14,6 @@
   </head>
   <body>
     <div id="react"></div>
-    <script src="/build/bundle.js"></script>
+    <script src="/build/313f39d7e16424d48a25.bundle.js"></script>
```
howardchung: revert this, this should only be done by Travis

howardchung:
signoff

PR #101, opened 2016-08-08
Mike | pagination. Look into /Pagination dir in the Table component d…

mike-robertson:
…ir. It has a higher order function which wraps the table with pagination. That table has an id which is referenced in the reducer.


howardchung:
looks good.  I'm thinking we should make the Table component self-contained, though (not use global state).  It seems kind of strange to store this table specific state in the redux store.

mike-robertson:
It is self contained within the redux store. Part of the reason for using redux is that we can always know the data flow (action -> reducer updates state -> connected component updates). It's just easier to understand data flow (less bugs) , and it allows us to do more complex things if we want to without any refactoring.

mike-robertson:
But in general, I would say don't be squeamish about using the store for everything. Also this reducer will pave the way for the sorting to be localized by table instead of by dataset

PR #105, opened 2016-08-18
Small changes in footer layout

TheCaptainDate:
Hi!
I made some small changes in footer layout. It looks better now.
I've made screenshots:
[before](https://dl.dropboxusercontent.com/u/7899634/ShareX/2016/08/chrome_2016-08-19_01-07-05.png)
[after](https://dl.dropboxusercontent.com/u/7899634/ShareX/2016/08/chrome_2016-08-19_01-39-50.png)

I am pretty new in React and all this things with importing css instead of writing complex css looks strange for me, so hope anything is ok. Also, why you guys not using a scss preprocessor? It's pretty cool and can speed up developing and make it easy for future theming.

### DIFF
--- a/components/Cheese/Cheese.css
+++ b/components/Cheese/Cheese.css
@@ -10,13 +10,6 @@
   }
 }
 
-.percent {
-  position: relative;
-  bottom: -17px;
-  margin-top: -17px;
-  z-index: 1;
-}
-
 .section {
   /* this actually better than calc */
   width: 33.3333%;
--- a/components/Cheese/Cheese.css
+++ b/components/Cheese/Cheese.css
@@ -3,6 +3,7 @@
   flex-direction: row;
   justify-content: space-around;
   flex-wrap: wrap;
+  margin-top: 15px;
 
   @media only screen and (max-width: 650px) {
     flex-direction: column;
@@ -10,14 +11,44 @@
 }
 
 .percent {
-  margin-bottom: 10px;
+  position: relative;
+  bottom: -17px;
+  margin-top: -17px;
+  z-index: 1;
 }
 
 .section {
-  width: calc(100% / 3);
-  text-align: center;
+  /* this actually better than calc */
+  width: 33.3333%;
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  align-items: center;
+
   @media only screen and (max-width: 650px) {
     width: 100%;
     margin-bottom: 25px;
   }
+}
+
+/* Make button aligned to center of container due to huge padding at the bottom */
+.section:nth-child(3) {
+  justify-content: center;
+}
+
+.cheese_title {
+  font-weight: 700;
+  margin-top: 0;
+  margin-bottom: 10px;
+}
+
+/* this is actually pretty strange way, but it looks fine so.. */
+.cheese_promo:first-child{
+  margin: 0;
+}
+
+.content {
+  width: 100%;
+  height: 100%;
+  text-align: center;
 }
--- a/components/Cheese/Cheese.css
+++ b/components/Cheese/Cheese.css
@@ -36,14 +36,14 @@
   justify-content: center;
 }
 
-.cheese_title {
+.cheeseTitle {
   font-weight: 700;
   margin-top: 0;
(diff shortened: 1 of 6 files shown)


howardchung:
looks ok to me.  @mike-robertson any comments?

components/Cheese/Cheese.css
```diff
@@ -3,21 +3,52 @@
...
+  margin-top: -17px;
+  z-index: 1;
 }
 
 .section {
-  width: calc(100% / 3);
-  text-align: center;
+  /* this actually better than calc */
+  width: 33.3333%;
+  display: flex;
```
mike-robertson: Why is this better than Calc? We don't need to care about browsers that don't support Calc.

components/Cheese/Cheese.css
```diff
@@ -3,21 +3,52 @@
...
+.cheese_title {
+  font-weight: 700;
+  margin-top: 0;
+  margin-bottom: 10px;
+}
+
+/* this is actually pretty strange way, but it looks fine so.. */
+.cheese_promo:first-child{
+  margin: 0;
+}
```
mike-robertson: Use camelCase for consistency

components/Cheese/Cheese.css
howardchung: My guess is that it saves a bit of evaluation.  I'm fine with either, although I suppose it's more portable to be closer to pure CSS?

mike-robertson: But those decimals so uggo. Lol

mike-robertson:
I added comments. We use postcss instead of a preprocessor like sass. It's generally better but you can look up reasons for specifics. Anything you can do in sass is possible in some way with postcss but you can also do more (essentially). Definitely looks cleaner, not sure how I feel about the negative margin as it might blow up in mobile or other browsers or in general just look off by a pixel or 2 in some situations.

components/Cheese/Cheese.css
TheCaptainDate: Well calc produce same value (or almost same) as this constant, so there is no reason for calc if we can do is without, wasn't I right?

components/Cheese/Cheese.css
TheCaptainDate: fixed

TheCaptainDate:
Fixed camel case and give my thoughts about calc()

About negative margins - i use it to take back space taken by relative percents text. Since this text ```relavite -17 bottom``` and ```z-index 1``` he doesnt actually take any space (logically), but browser think he is.
The other way to achive same effect - make a new parent div  with ```position: relative``` for percents text, and make percents ```position: absolute; bottom: -17;```. Since absolute elements doesn't exist for browser layout - text won't take any space. But way with margin much more easier and clean.
Also, take a look: https://www.smashingmagazine.com/2009/07/the-definitive-guide-to-using-negative-margins/
Negative margin absolutely legit and valid by W3

mike-robertson:
I don't think the issue is that negative margin should never be allowed: it's just brittle. If we change anything about those widths then it's broken again. I'd much rather find a way to change the pseudo element `::after` `content` of the inner progress meter to be equal to the percent. However, I haven't found a good way of doing that yet, so let's just stick with the negative margin (unless you can figure that out).

TheCaptainDate:
@mike-robertson Well this can be solved at least by 2 ways 
1. We can define a variable to change both height and margin at same time (don't know how to make this in postcss, but think it's possible)
2. I really can play around before and after of progressbar but i am not shure how to edit his styles. There is some css encapsulation so seems like i can't just override styles. And also this is a big question how to set this percents values.

I actually think that much easier just fixt position if it will be broken afret some changes, because other ways too much complex for such small thing

mike-robertson:
I like approach 1. You could actually wrap the progress meter in its own component which takes the width as a prop and use that same value to do the negative margin on the value you are displaying (also a prop)

TheCaptainDate:
I am been away for some time. Now i am back.

 Made new Progress Bar component which receives height and percentage. So it just draws material progress bar and applies label to it.

mike-robertson:
Everything looks good to me. I'll merge it in now.

PR #112, opened 2016-08-29
Modified the Request a Parse Error Message

dhpaquette:
Hi there,
I made a small change to the functionality in Request.jsx

It looks something like this

- All I did was use the 'errorText' attribute built into 'TextField' from material-ui.
- I also cleared the entered value after submission because the text from the floating label would overlap with the number. 
- I know this was not listed as an issue, but it was quick and I think it's an improvement to the UI/UX , albeit very small.

***
I took interest in this project since I know about Yasp, and I think it'd be a solid way for me to learn/practice React. I had no idea you would be redesigning the UI. I'm still in the early stages with React, but I would like to continue contributing.

### DIFF
--- a/components/Request/Request.jsx
+++ b/components/Request/Request.jsx
@@ -8,6 +8,7 @@ import { REDUCER_KEY } from '../../reducers';
 
 const Request = ({ error, matchId, loading, progress, dispatchRequest, dispatchMatchId }) => {
   function submit() {
+    document.getElementById('match_id').value = '';
     dispatchRequest(matchId);
   }
   const progressIndicator = (progress ?
@@ -19,11 +20,11 @@ const Request = ({ error, matchId, loading, progress, dispatchRequest, dispatchM
       <TextField
         id="match_id"
         floatingLabelText="Match ID"
+        errorText={error ? 'Failed to get match data.' :false}
         value={matchId}
         onChange={(e) => dispatchMatchId(e.target.value)}
       />
       <div>Only works for public matches with replay available in client</div>
-      <div>{error ? 'Failed to get match data.' : ''}</div>
       {loading ? progressIndicator : <RaisedButton label="Submit" onClick={submit} />}
     </div>
   );
--- a/components/Request/Request.jsx
+++ b/components/Request/Request.jsx
@@ -8,8 +8,8 @@ import { REDUCER_KEY } from '../../reducers';
 
 const Request = ({ error, matchId, loading, progress, dispatchRequest, dispatchMatchId }) => {
   function submit() {
-    document.getElementById('match_id').value = '';
     dispatchRequest(matchId);
+    dispatchMatchId('');
   }
   const progressIndicator = (progress ?
     <CircularProgress value={progress} mode="determinate" /> :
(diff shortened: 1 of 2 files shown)


components/Request/Request.jsx
```diff
@@ -8,6 +8,7 @@ import { REDUCER_KEY } from '../../reducers';
 
 const Request = ({ error, matchId, loading, progress, dispatchRequest, dispatchMatchId }) => {
   function submit() {
+    document.getElementById('match_id').value = '';
```
howardchung: I think you should dispatchMatchId here.  React discourages direct mutation of DOM state.

components/Request/Request.jsx
```diff
@@ -19,11 +20,11 @@ const Request = ({ error, matchId, loading, progress, dispatchRequest, dispatchM
       <TextField
         id="match_id"
         floatingLabelText="Match ID"
+        errorText={error ? 'Failed to get match data.' :false}
         value={matchId}
```
howardchung: Space after colon

components/Request/Request.jsx
```diff
@@ -19,11 +20,11 @@ const Request = ({ error, matchId, loading, progress, dispatchRequest, dispatchM
       <TextField
         id="match_id"
         floatingLabelText="Match ID"
+        errorText={error ? 'Failed to get match data.' :false}
```
bguzryanto: nit space before 'false'

howardchung: I think eslint will fail if it's there, so it needs to be fixed

components/Request/Request.jsx
bguzryanto: instead of using document.getElementById, you can use `dispatchMatchId('')`

dhpaquette: Yeah, you're right. I was not sure about this part, but I wanted to clear the value because the overlap from the animation is jarring to look at. If I simply dispatchMatchId it won't re-render and change the 'value' attribute in the TextField though, as far as I'm aware. What do you have in mind?

components/Request/Request.jsx
dhpaquette: Added

components/Request/Request.jsx
howardchung: I think React takes care of updating the DOM when you change the state.

dhpaquette: That was my impression, granted I am new to it so maybe I'm likely missing something crucial here, but I could not get that to work. dispatchMatchId did not clear the text upon submission. I could just remove the line entirely until I figure out something better. Alternatively, one could keep the matchId in the TextField and prevent the floatinglabel from moving downwards? But you would lose the luster of the animation, so I do not like that as much.

bguzryanto: This is because state matchId never updated by reducer. See https://github.com/yasp-dota/ui/blob/master/reducers/request.js#L35 use `match_id` instead of `matchId` which is used by request action creator to pass action to reducer https://github.com/yasp-dota/ui/blob/master/actions/requestActions.js#L20, so state matchId won't updated to the state tree. That's why request went error. 

To overcome this problem you need to modify request reducer then modify Request.jsx to put `dispatchMatchId` after `dispatchRequest`

dhpaquette: Forgive my ignorance of redux and reducers right now, but do you mean I do something like this:
`case requestActions.MATCH_ID:
      return {
        ...state,
        matchId:` **action.matchId**`,
      };`

###### in the reducer and this in Request.jsx:

`dispatchRequest(matchId);`
 `dispatchMatchId('');`

***
Because that does indeed work in changing the state. I can see why it would not be linked up for testing purposes.

components/Request/Request.jsx
howardchung: space after :

howardchung:
push your new commit?

PR #119, opened 2016-09-01
Mike | fix issue 118 and an issue with default export change where im…

mike-robertson:
…port was still considering dependency as a named export


howardchung:
signoff

PR #129, opened 2016-09-07
Update account widget

andrkrn:


howardchung:
Thanks!

@mike-robertson any comments?

@andrkrn it would also be nice if you could provide a screenshot of what it looks like :)

howardchung:
Since users can choose their own names, can they mess up layout by having extremely long names?

components/AccountWidget/LoggedIn.jsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  return getPlayerWidget();
+};
+
+const mapStateToProps = (state, ownProps) => ({
+  loading: player.getLoading(state, ownProps.playerId),
+  error: player.getError(state, ownProps.playerId),
+  playerName: player.getPlayerName(state, ownProps.playerId),
+  playerPicture: player.getPicture(state, ownProps.playerId),
```
howardchung: what happens if the picture is not loaded?

andrkrn:
Ups, my bad. there should be `max-width` for the name

components/AccountWidget/LoggedIn.jsx
andrkrn: hmm, it's should request user on the first load if user logged in on `AccountWidget.jsx`. and when go to player profile, it will reuse data on reducer.
CMIIW

howardchung: But what about the case where the user is logged in already, then visits a page that is not their profile?  The player data won't be loaded then.

andrkrn: Just tested it and it's worked, the user data loaded when visit directly to homepage.
Do you have that issue ? Because i can't reproduce the issue on local.

howardchung: yeah, it looks like the getPlayer function call retrieves the player data.  It's an extra API call but it's probably fine.

howardchung:
signoff

@mike-robertson any comments?

components/Spinner/Spinner.jsx
```diff
@@ -1,4 +1,4 @@
 import React from 'react';
 import CircularProgress from 'material-ui/CircularProgress';
 
-export default ({ size = 1 }) => <div><CircularProgress size={size} /></div>;
+export default ({ size = 1, color = '#00BCD4' }) => <div><CircularProgress size={size} color={color} /></div>;
```
mike-robertson: i'd say that this color might need to be in or from the palette if it's a commonly used theme color (if we change theme colors, do we need to change this color?). If it's not, it's not terribly important to put it in palette.

components/palette.css
```diff
@@ -57,4 +57,10 @@
...
   justify-content: center;
-}
\ No newline at end of file
+}
+
+.verticalAlign {
+  display:flex;
+  align-items: center; /* Vertical center alignment */
+  justify-content: center; /* Horizontal center alignment */
+}
```
mike-robertson: you don't need to comment this since it's basic flexbox behavior.

components/AccountWidget/AccountWidget.jsx
```diff
@@ -48,4 +49,20 @@ const mapDispatchToProps = (dispatch) => ({
...
-export default connect(mapStateToProps, mapDispatchToProps)(AccountWidget);
+const getData = props => {
+  props.getPlayer(props.user.account_id);
+};
+
+class RequestLayer extends React.Component {
+  componentWillUpdate(nextProps) {
+    if (nextProps.user && nextProps.user.account_id) {
+      getData(nextProps);
+    }
```
mike-robertson: this is going to fire every time the component updates, even if the props are the same. ComponentWillUpdate doesn't mean that every prop changed, just that something changed. You should check that nextProps.user.account_id !== this.props.user.account_id, otherwise you will fire this async request every rerender (I think). Also, I think it makes more sense to just call this.props.getPlayer(nextProps.user.account_id) since your factored out function is just a single line. Most importantly though, this isn't getting the user. This is getting player data. Player data is used by the /players pages. Metadata stores the user info. You need to call getMetadata here, not getPlayer. Look at the main index.jsx page. It dispatches getMetadata on app load. So I don't know if you ever even need to dispatch any actions since going to log in will set the cookie and then when the user comes back, it will cause a page refresh and getMetadata. So I would remove this entire RequestLayer component as well as mapDispatchToProps.

howardchung: getPlayer is correct.  metadata only contains the user ID, but not the name/profile picture.

mike-robertson: Whoops, my bad. Updated comment.

components/AccountWidget/LoggedIn.jsx
mike-robertson: This is nice as well because it preloads player data so it should feel more responsive with this.

howardchung:
@andrkrn could you sync to latest master and resolve conflicts?

components/AccountWidget/AccountWidget.jsx
andrkrn: just checked it's will fire just once

andrkrn:
@howardchung synced to master

howardchung:
Thanks!

PR #132, opened 2016-09-08
more match tables

howardchung:
More work on match tables:

* Chat
* Ability uses/hits/damage
* Purchases counts table


components/Match/Match.jsx
```diff
@@ -79,27 +85,28 @@ class RequestLayer extends React.Component {
...
+                damage_inflictor: (p.damage_inflictor || {})[k],
+              }))}
+              columns={[{ displayName: 'Ability', field: 'name' },
+            { displayName: 'Casts', field: 'casts', sortFn: true },
+            { displayName: 'Hits', field: 'hero_hits', sortFn: true },
+            { displayName: 'Damage', field: 'damage_inflictor', sortFn: true }]}
+            />
+          </Tab>
+          ))
         }
```
mike-robertson: It would probably be worthwhile to make this a component or higher order component (similar to withPagination and PaginatedTable). I could see using this in other areas at least.

components/Match/Match.jsx
```diff
@@ -79,27 +85,28 @@ class RequestLayer extends React.Component {
...
+        </Tabs>
+        <Table data={this.props.match.chat.map(c => Object.assign({}, c, this.props.match.players[c.slot]))} columns={chatColumns} />
       </div>
```
mike-robertson: I prefer using spread over Object.assign as it's a little easier to read, but it's not necessary.

components/Match/analysis.js
```diff
@@ -0,0 +1,341 @@
+/* eslint-disable */
+import util from 'util';
+import constants from 'dotaconstants';
+
+function generatePlayerAnalysis(match, pm)
+{
+  // define condition check for each advice point
+  const advice = {};
+  const checks = {
+    // EFF@10
```
mike-robertson: could you comment or change the name (at least internal name) to something besides eff? it's not readily apparent what it is until you read the name attribute of the object returned.

components/Match/analysis.js
```diff
@@ -0,0 +1,341 @@
...
+      'bloodstone': 1,
+      'radiance': 1,
+    };
+    return (constants.items[key].desc.indexOf('Active: ') > -1 && !(key in whitelist));
+  }
+}
+export
+  {
+    generatePlayerAnalysis,
+  };
```
mike-robertson: I think we should convert this to our lint standards sooner rather than later, but if it's a direct port it's not necessary asap if you think time is better spent updating functionality. But we should keep an issue open so we don't lose track of it.

components/Match/matchColumns.jsx
```diff
@@ -249,42 +249,109 @@ const laningColumns = [
...
 
+const purchaseColumns = [
+  heroTdColumn,
+  {
+    displayName: 'TP',
+    tooltip: 'Town Portal Scrolls purchased',
+    field: 'purchase',
+    width: 1,
+    sortFn: true,
+    displayFn: (row, col, field) => field ? field.tpscroll : '',
```
mike-robertson: I think another way you could do this (without using ternary everywhere) is 
```js
(row, col, { tpscroll }) => tpscroll
// or maybe
(row, col, { tpscroll = '' }) => tpscroll
```
I'm not sure if the default value is needed or not. Not sure if it's better or worse, but something to think about.

components/Match/renderMatch.js
```diff
@@ -0,0 +1,324 @@
+/* eslint-disable */
+import { isRadiant } from '../../utility';
```
mike-robertson: Similar to the other file, I think we should make an issue for upgrading this so we don't lose track.

components/Match/renderMatch.js
```diff
@@ -0,0 +1,324 @@
...
+import constants from 'dotaconstants';
+import { generatePlayerAnalysis } from './analysis';
+
+function renderMatch(m)
+{
```
mike-robertson: part of the refactoring process should be upgrading to 'self-documenting' code with more descriptive variable names.

mike-robertson:
I didn't read over the 2 ported modules exhaustively, but do think we should create issues for upgrading them/refactor. If they work now and you think your time is better spent on mvp stuff, just let it sit in the backlog for now. There were a couple other comments but they are minor things so change at your discretion.

components/Match/Match.jsx
howardchung: Yeah, I know i'll need it at least twice, once for items and once for abilities.

components/Match/analysis.js
howardchung: probably won't do this for MVP, but can do when we ES6 this method.

components/Match/matchColumns.jsx
howardchung: interesting.  I'll try it

howardchung: Problem is that it complains when we try to destructure a property in snake_case

PR #139, opened 2016-09-15
Better player header

### LINKED ISSUE #87: display cheese/track status on player pages

blukai:
fixes #87 

image
image


howardchung:
Reviewers:
@mike-robertson 
@albertcui 
@TheCaptainDate

mike-robertson:
Damn this looks good

components/Player/PlayerHeader.jsx
```diff
@@ -1,18 +1,104 @@
...
+const mapStateToProps = (state, ownProps) => ({
+  loading: player.getLoading(state, ownProps.playerId),
+  error: player.getError(state, ownProps.playerId),
+  playerName: player.getPlayerName(state, ownProps.playerId),
+  registered: player.getLastLogin(state, ownProps.playerId),
+  picture: player.getPictureFull(state, ownProps.playerId),
+  steamLink: player.getSteamLink(state, ownProps.playerId),
+  cheese: player.getCheese(state, ownProps.playerId),
+  playerId: ownProps.playerId,
+});
```
mike-robertson: It's not necessary to pass playerId down, connect will do that automatically.

mike-robertson:
I approve pending the small change I commented on.

TheCaptainDate:
This is definitely better than mine realisation! I am like this much more.

Btw somewhere we have inline styles, somewhere we use css files. This strange for me and i think there is should be only one strategy.
Also howard asked me to move cheese icon class to palette.css, so i think this pull request also should do this since there is no more reason to finish mine one.

And the last thing - we should show that user registered but excluded from auto-parse quence because he didn't visit site for the pasn N days.
[Take a look](https://github.com/TheCaptainDate/ui/blob/6158a5345156c4dc572cbfe83f06336c4731e4ff/components/Player/PlayerName.jsx#L24-L28)

blukai:
@TheCaptainDate, @howardchung 
> And the last thing - we should show that user registered but excluded from auto-parse queue because he didn't visit site for the past N days.

How many days?
Show one more icon with tool-tip if user has visited the site not so long ago and hide if not?

howardchung:
The player API returns a field called tracked_until.  If the current timestamp is greater than it, the player is no longer tracked.

mike-robertson:
Tracked until should be available in the store, I think it may even have its own selector, but I don't remember.

components/Player/PlayerStats.jsx
```diff
@@ -1,10 +1,83 @@
...
+          style={card.general} subtitleStyle={card.title} titleStyle={card.subtitle}
+          subtitle={soloRank || 'N/A'} title="Solo MMR"
+        /> : null}
+        {partyRank ? <CardTitle
+          style={card.general} subtitleStyle={card.title} titleStyle={card.subtitle}
+          subtitle={partyRank || 'N/A'} title="Party MMR"
+        /> : null}
+        {mmrEstimate.estimate ? <CardTitle
+          style={card.general} subtitleStyle={card.title} titleStyle={card.subtitle}
+          subtitle={mmrEstimate.estimate || 'N/A'} title="estimate MMR"
```
howardchung: estimated

howardchung: at some point we'll also need a tooltip explaining how the estimate works.

TheCaptainDate:
@blukai In my realization icon just changes color to grey
And of course there is should be a tooltip for this case as i think.

blukai:
image

Any ideas how to fix errors? :flushed:

components/Player/PlayerBadges.jsx
```diff
@@ -0,0 +1,61 @@
...
+            Steam profile
+          </ReactTooltip>
+        </div>
+        {Math.round(new Date().getTime() / 1000.0) >= Number(tracked) ? null :
+          <div>
+            <div data-tip data-for="tracked">
+              <ActionVisibility className={styles.icon} />
+            </div>
+            <ReactTooltip id="tracked" place="top" type="light" effect="float">
+              This user is tracked.
```
howardchung: explain what tracking means (new matches have replays parsed)

howardchung:
```
/home/travis/build/odota/ui/components/Player/PlayerHeader.jsx
  12:8  error  Using exported name 'PlayerStats' as identifier for default export   import/no-named-as-default
  13:8  error  Using exported name 'PlayerBadges' as identifier for default export  import/no-named-as-default
```

I think eslint doesn't like it when you use the name for a default export.  I think having just a single un-named default export fixes it (and the import should be just `import PlayerStats from './PlayerStats``

blukai:
image
Someone has to make these buttons work

components/Player/PlayerActions.jsx
```diff
@@ -0,0 +1,35 @@
...
+      <div data-tip data-for="update">
+        <RaisedButton
+          disabled={true}
+          icon={<ActionUpdate />}
+          style={{ minWidth: 50 }}
+        />
+      </div>
+      <ReactTooltip id="update" place="left" type="light" effect="float">
+        <div style={{ textAlign: 'left' }}>
+          Refresh Match History:
```
howardchung: Can we make this one continuous string?  Eventually we'll move them all out to language files and it's easier to deal with it that way.

components/Player/PlayerBadges.jsx
```diff
@@ -29,6 +29,8 @@ export const PlayerBadges = ({ loading, error, cheese, tracked, steamLink }) =>
             </div>
             <ReactTooltip id="tracked" place="top" type="light" effect="float">
               This user is tracked.
+              <br />
+              New matches have replays parsed
```
howardchung: I'd use parentheses: `(New matches will have their replays automatically parsed)`

components/Player/PlayerHeader.jsx
```diff
@@ -9,8 +9,11 @@ import styles from './PlayerHeader.css';
 import Error from '../Error';
 import Spinner from '../Spinner';
 import { player } from '../../reducers';
+/* eslint-disable */
```
howardchung: We should probably solve the problem instead of working around it like this :)

components/Player/PlayerActions.jsx
blukai: Of course, it's cool, I could help in the translation into Russian and Ukrainian if necessary

howardchung:
Yeah, we can enable the buttons later.

PR #140, opened 2016-09-19
Purchases/item uses table

howardchung:
Implement more match tables.


howardchung:
any comments @mike-robertson ?

PR #141, opened 2016-09-19
Global colors, tables

blukai:
image


mike-robertson:
This looks amazing! Is this ready for review?

components/Player/Pages/playerMatchesColumns.jsx
```diff
@@ -26,43 +24,16 @@ export default [{
...
-  width: 1,
-  sortFn: true,
-}, {
-  displayName: 'P',
-  field: 'version',
-  width: 1,
-  sortFn: true,
-  displayFn: (row, col, field) => (field ? <ActionDone /> : <div />),
+  sortFn: (row) => ((row.kills + row.deaths + row.assists) / row.deaths),
+  width: 2,
```
mike-robertson: Only concern here is whether we are removing useful functionality (sorting by deaths/assists/kills separately). Definitely cleaner look and definitely frees up lots of space on the table.

blukai:
@mike-robertson, thanks, yes

components/Player/PlayerActions.jsx
```diff
@@ -1,17 +1,19 @@
 import React from 'react';
```
mike-robertson: Not a huge fan of this being called PlayerActions.jsx since it may confuse people into thinking this is somehow related to redux actions. Not a big deal but something to think about.

components/Player/PlayerTabs.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import TabBar from '../TabBar';
+import { playerPages } from '../Header/Pages';
+
+const playerPagesMapped = (accountId) => playerPages.map((e) => Object.assign({}, e, {
+  route: `/players/${accountId}/${e.name.toLowerCase()}`,
+  label: e.name,
+}));
+
```
mike-robertson: could we change this to use the spread operator since it is easier to read and matches the code style we have been using? ex:
```js
const playerPagesMapped = (accountId) => playerPages.map(page => ({
  ...page,
  route: `/players/${accountId}/${e.name.toLowerCase()}`,
  label: e.name,
}));
```

components/Table/TableContainer.jsx
```diff
@@ -1,10 +1,14 @@
...
 import styles from './TableContainer.css';
 
 const TableContainer = ({ title, style, children }) => (
   <div className={styles.container} style={{ ...style }}>
     <div className={styles.heroesContainer}>
-      <div className={styles.tableHeading}>{title}</div>
+      <div className={styles.tableHeading}>
+        <HardwareKeyboardArrowRight style={{ verticalAlign: 'top', opacity: '.6', height: 26 }} />
+        {title}
+      </div>
```
mike-robertson: Do we plan on making the table name clickable in some way? Otherwise, it feels like it might trick users into thinking they can since that kind of arrow typically signifies that there is more info or options if the item is clicked.

components/Table/TablePercent.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import ReactTooltip from 'react-tooltip';
+import styles from './TablePercent.css';
+
```
mike-robertson: If we declare the function inside the stateless component then each re-render will trigger the creation of a new function. Not really the worst thing in the world but it could become memory intensive in some situations with very large tables (assuming those tables aren't paginated in some way).

mike-robertson: The trade off is that you have to pass all the props to the function (which can be annoying), though it does follow a more functional style since it becomes a pure function.

utility/index.jsx
```diff
@@ -29,39 +33,86 @@ export function formatSeconds(input) {
...
+        src={`${API_HOST}${constants.heroes[field] ? constants.heroes[field].img : ''}`}
+        role="presentation"
+        style={{ height: 29, marginRight: 7 }}
+      />
+      <div style={{ display: 'inline-block', verticalAlign: `${row.player_slot ? 'top' : 'super'}` }}>
+        {constants.heroes[field] ? constants.heroes[field].localized_name : ''}
+        {row.match_id ?
+          <span className={styles.subText} style={{ display: 'block', marginTop: 1 }}>
+            {isRadiant(row.player_slot) ? 'Radiant' : 'Dire'}
+          </span> :
```
mike-robertson: should we color this Radiant/Dire text or is that too busy? I'm not sure it's even useful to distinguish on the player matches screen so it probably doesn't matter.

components/Player/Pages/playerMatchesColumns.jsx
```diff
@@ -1,20 +1,18 @@
...
-}, {
   displayName: 'Hero',
   field: 'hero_id',
-  width: 3,
+  width: 3.1,
   displayFn: transformations.hero_id,
 }, {
+  displayName: 'Id',
+  field: 'game_mode',
+  width: 1.8,
```
mike-robertson: I'm not sure if this is useful, so I would like your opinion, but I think we need to add a way for user to clear their sort. After merging this, there won't be a way to get the default ordering displayed unless you refresh the page or we change this from working ont he game_mode field to the id field (so that sort would work on id). The other option would be allowing there to be some sort of button users can click (which is intuitive) that would clear sorting. I'm not sure which is better (or even if we really care about it).

mike-robertson:
@blukai @howardchung I think this is ready to merge pending (minimal) action on comments.

components/AccountWidget/AccountWidget.jsx
```diff
@@ -23,7 +23,7 @@ const AccountWidget = ({ loading, error, user, style }) => (
     {error && <Error />}
     {!error && !loading && user ? (
       <div>
-        <Link className={styles.tab} to={`/players/${user.account_id}`}>{"Profile"}</Link>
+        <Link className={styles.tab} to={`/players/${user.account_id}/overview`}>{"Profile"}</Link>
```
howardchung: users entering without "overview" should also get the same page.  I assume this is done to cause the inkbar to show up on the navbar (there was a previous issue with this)

components/Header/Header.jsx
```diff
@@ -28,34 +24,15 @@ const matchPagesMapped = (matchId) => matchPages.map((e) => Object.assign({}, e,
...
-  params,
   location,
 }) => (
   <div>
     <Toolbar className={styles.header}>
       <ToolbarGroup className={styles.verticalAlign}>
-        <Link to="/">
-          <strong>{APP_NAME}</strong>
+        <Link to="/" style={{ textTransform: 'uppercase', marginRight: 10 }}>
+          <strong>&#60;{APP_NAME}&#47;&#62;</strong>
```
howardchung: Could we use template string here to make it more readable?

components/Player/Pages/playerHeroesColumns.jsx
```diff
@@ -1,28 +1,31 @@
...
+  sortFn: (row) => (row.last_played),
 }, {
-  displayName: 'Last',
-  field: 'last_played',
-  width: 2,
-  sortFn: 1,
-  displayFn: transformations.last_played,
-}, {
-  displayName: 'Played',
+  displayName: 'mp',
```
howardchung: does "Matches" work?  Or is it too long?

components/Player/Pages/playerMatchesColumns.jsx
howardchung: I agree this should sort on match ID.  You can do this by either changing the field, or the sortFn

components/Player/Pages/playerMatchesColumns.jsx
```diff
@@ -1,20 +1,18 @@
...
+  sortFn: true,
+  displayFn: transformations.match_idANDgame_mode,
```
howardchung: snake_case the whole name?  match_id_and_game_mode

components/Player/Pages/playerMatchesColumns.jsx
howardchung: The current UI has support for server-side sorting of kills/deaths/assists.  Possibly that could be used in lieu of client-side sorting.

utility/index.jsx
```diff
@@ -29,39 +33,86 @@ export function formatSeconds(input) {
   return time;
 }
 
-export const getPercentWin = (wins, games) => (games ? ((wins / games) * 100) : 0);
+export const getPercentWin = (wins, games) => (games ? (Math.round(((wins * 100) / games) * 100) / 100) : 0);
```
howardchung: If the purpose of this is just to round to 2 decimal places, use .toFixed(2)

howardchung:
I'm getting excited every time I see a PR from @blukai :)  Thank you so much!

utility/index.jsx
blukai: .toFixed(2) makes 15 = 15.00, I would like to avoid unnecessary :zero::zero:

components/Table/TableContainer.jsx
blukai: without icon it look not as I would like to :art:
changed to another

image

components/Player/Pages/playerMatchesColumns.jsx
mike-robertson: That seems wasteful, but I don't think it's really an issue. I would guess people care most about seeing where they have a positive impact vs. where they have a negative impact. The only case where that might be different is if they wanted to sort on all their matches to see their highest kills or deaths matches, but that seems like an edge case we don't need to support.

components/Table/TableContainer.jsx
mike-robertson: Agreed, it does seem to add something nice with something in front of the title. You seem to have a good eye for design so please feel free to mess around with stuff or disagree with me since I am not a designer nor do I have a natural talent for it.

components/AccountWidget/AccountWidget.jsx
mike-robertson: Yea, they are finally merging in a PR I put into material-ui, so I think that will fix that issue but I am not 100% sure.

mike-robertson:
Should we change the color of the filter button? The orange doesn't seem right with this new color scheme. Maybe it should be one of the blues or reds we are using?

utility/index.jsx
howardchung: You can wrap the whole thing in Number() to remove any unnecessary decimals.

components/Player/Pages/playerHeroesColumns.jsx
blukai: "Matches" is too long, I will add tooltips in some next pr

components/Player/Pages/playerMatchesColumns.jsx
```diff
@@ -26,14 +26,24 @@ export default [{
...
+  displayName: 'D',
+  field: 'deaths',
+  sortFn: true,
+  width: 0.8,
+}, {
+  displayName: 'A',
+  field: 'assists',
+  sortFn: true,
+  width: 0.8,
 },
```
mike-robertson: I know I made a comment about this being split, but does it remove a lot of space from the table split? Do you have a screenshot? It may be worth the loss of sortability if it means we can display our data more cleanly/fit in more stuff.

components/Player/PlayerTabs.jsx
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import TabBar from '../TabBar';
 import { playerPages } from '../Header/Pages';
 
-const playerPagesMapped = (accountId) => playerPages.map((e) => Object.assign({}, e, {
-  route: `/players/${accountId}/${e.name.toLowerCase()}`,
-  label: e.name,
+const playerPagesMapped = (accountId) => playerPages.map(page => ({
+  route: `/players/${accountId}/${page.name.toLowerCase()}`,
```
mike-robertson: you may want to do a spread at the beginning there, otherwise you will lose all the old values on the page.

components/Player/Pages/playerMatchesColumns.jsx
blukai: image

components/Player/PlayerTabs.jsx
blukai: idk what to do with these errors :laughing: 
image
may I just disable eslint here? hehe

mike-robertson: Something is wrong with eslint or babel I think... Does it run or does it fail to transpile with webpack?

howardchung: CI is passing.  It's probably the local IDE not recognizing the spread operator (one of the reasons I still use object.assign)

blukai: webpack bundle is valid

mike-robertson: Ok. I would recommend updating your linting package. I use atom and have not had this issue before. If you are curious about atom, you can get a good list of packages to download from my github, though I would recommend changing the semvar to not be absolute. Also atom is slow as shit with large files.

howardchung: It should be fine to merge this since it's valid in CI.  However if you want your local syntax highlighting to work, you'll either have to figure out your IDE (probably a spread operator plugin), or just go back to Object.assign()

blukai:
Ready to merge :four_leaf_clover:

howardchung:
Let's get @mike-robertson to sign off.  If you're interested in regular contributions, we can make you a committer as well :)

blukai:
@howardchung Yeah, I really like opendota, and contributions is a good way to improve myself I think.

components/Player/PlayerTabs.jsx
mike-robertson: I would comment that if we are making design/code style decisions because the IDE isn't working, it probably signals we shouldn't be using those IDE's instead of changing our code.

PR #146, opened 2016-09-20
add rankings player page

mike-robertson:


actions/player/playerRankingsActions.js
```diff
@@ -0,0 +1,60 @@
...
+  } else {
+    dispatch(getPlayerRankingsRequest(playerId));
+  }
+  // const modifiedOptions = getModifiedOptions(options, excludedOptions);
+
+  return fetch(`${host}${getUrl(playerId, options, url)}`, { credentials: 'include' })
+    .then(response => response.json())
+    // We want to filter out the rankings where the person is so bad they don't even have a rank
+    // for that hero
+    .then(json => dispatch(getPlayerRankingsOk(json.filter(rankings => rankings.rank), playerId)))
```
howardchung: do this in the reducer?

components/Player/Pages/Rankings.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+const PlayerRankingsTable = createTable(
+  playerRankings.getPlayerRankingsById,
+  (state, sortState, playerId) => (sortState ? sortPlayerRankings(playerId)(state) : transformPlayerRankingsById(playerId)(state)),
+  setPlayerRankingsSort
+);
+
+const Rankings = ({ playerId }) => (
+  <div>
+    <TableFilterForm submitAction={getPlayerRankings} id={playerId} page="heroes" />
```
howardchung: This page won't need the filter.

selectors/transformPlayerRankings.js
```diff
@@ -0,0 +1,11 @@
...
+import { playerRankings } from '../reducers';
+
+const getRankings = (id, numRows) => state => playerRankings.getRankingList(state, id).slice(0, numRows);
+
+const transformPlayerRankingsById = (id, numRows) => createSelector(
+  [getRankings(id, numRows)],
+  rankings => rankings
+);
+
+export default transformPlayerRankingsById;
```
howardchung: remove extraneous files

actions/player/playerRankingsActions.js
mike-robertson: reducers shouldn't be in charge of this kind of mutation. The reducer should just be pulling data from the action and returning a new portion of state unless mutation is absolutely necessary. I think though that it makes sense to also do this in the selector if we want that data in the reducer for some reason but don't want to display it.

components/Player/Pages/Rankings.jsx
mike-robertson: ok, does it not work with filters?

selectors/transformPlayerRankings.js
mike-robertson: this file is used

howardchung: See the next two files.  I couldn't comment on them since they're empty

components/Player/Pages/Rankings.jsx
howardchung: Yes, rankings and MMR do not support match filtering.

actions/player/playerRankingsActions.js
howardchung: In that case, I should update the match action to do renderMatch instead of the reducer.

selectors/transformPlayerRankings.js
mike-robertson: lol I don't even know what those are... I'll remove them and the filtering

actions/player/playerRankingsActions.js
mike-robertson: Nothing 'bad' will happen if you do that in the reducer, but I think it's considered a good practice to keep as much logic out of the reducers as possible.

mike-robertson:
I think you can close this and just do the pros branch commit (since it contains this). Then you can just review that (which has the reviewed changes)

howardchung:
OK

PR #158, opened 2016-09-22
Icons, footer

### LINKED ISSUE #157: make svg social icons
steam, twitter, discord, github, ...more?
as [example](https://github.com/odota/ui/blob/bf82d9516c1feb30dbda69d989b58e750a268df0/assets/IconCheese.jsx)
@howardchung, @mike-robertson, should I move icons, like this, to /components/Icons ?

blukai:
fixes #157


components/Icons/Discord.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+export default (props) => (
+  <svg {...props} viewBox="0 0 300 300">
```
howardchung: how did you generate these?

blukai: with Adobe Illustrator

howardchung:
CI says some import path issue.

blukai:
@howardchung, yes, I know, it's not yet done :construction:

components/Icons/Discord.jsx
mike-robertson: make sure if you are spreading props onto a plain element that you aren't putting non-html tags on it since it will cause react to give warnings.

PR #179, opened 2016-09-24
Update directory structure so that we use a /src folder

mike-robertson:
We don't want all our stuff to be at the root level of the project. This change also allows us to use absolute paths for modules in webpack.

Ex:
```js
import { something } from '../../config';
// becomes
import { something } from 'config';
```

I've already refactored everything to use absolute paths except components since it is sort of tricky to do with a regex. But going forward, we can start using 
```js
import { MyComponent } from 'components/MyComponent';
// instead of
import { MyComponents } from '../../MyComponent';
```


mike-robertson:
Reviewer: @howardchung 
Please try to get this merged before we do other PR's because it's a massive pain in the ass to merge

PR #180, opened 2016-09-24
match tabs

howardchung:
* Put match tables in tabs.  Each table/set of tables is a child of a Tab element.  I'm not sure if we want to adopt this model as well for player pages.
* Remove match selectors
* Implement additional match tables


howardchung:
requesting signoffs

src/components/Match/Match.jsx
```diff
@@ -90,22 +100,55 @@ class RequestLayer extends React.Component {
...
+        <Tabs>
+          <Tab label="Overview">
+            <MatchPlayersTable columns={overviewColumns} />
+            <MatchPlayersTable columns={abUpgradeColumns} />
+            <BuildingMap match={match} loading={this.props.loading} />
+          </Tab>
+          <Tab label="Benchmarks">
+            <MatchPlayersTable columns={benchmarksColumns(match)} />
+          </Tab>
+          <Tab label="Crosstables">
```
nicholashh: Benchmarks and Crosstables overlap on my screen

howardchung:
Yes, I think we'll want to standardize on the TabBar style established by @blukai that's currently used on player pages.

PR #181, opened 2016-09-24
Added docker

### LINKED ISSUE #176: Docker for UI
I don't use Docker for odota/ui myself, but it might be useful for people who like to run everything in Docker.

This will be a lot simpler than the configuration for odota/core.  There's only a single container and no linking needed.

perhaps @dartharaknus or @Daerdemandt might be interested.

dartharaknus:
fixes #176 

Can't use alpine-node due to issues on windows.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM node:6.6.0
+
+ENV NPM_CONFIG_LOGLEVEL warn
+
+COPY . /usr/src
+
+WORKDIR /usr/src
+
+RUN npm install
+
+ENV PATH /usr/src/node_modules/.bin:$PATH
+
+CMD ["npm", "start"]
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "webpack": "NODE_ENV=production webpack -p --bail",
     "watch": "webpack --watch",
+    "prestart": "npm install",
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ENV NPM_CONFIG_LOGLEVEL warn
+
+COPY . /usr/src
+
+WORKDIR /usr/src
+
+RUN npm install
+
+ENV PATH /usr/src/node_modules/.bin:$PATH
```
howardchung: This is needed for pm2 on odota/core.  Is it needed for anything here?

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "webpack": "NODE_ENV=production webpack -p --bail",
     "watch": "webpack --watch",
+    "prestart": "npm install",
```
howardchung: what is this used for?

Dockerfile
dartharaknus: If people want to ssh into the container and run the `webpack` command.

package.json
dartharaknus: Since, we are mounting the folder for development, we need to reinstall the node_modules. I didn't want to have a separate script which does `npm install && npm start`, so I used `prestart`

howardchung: isn't `npm install` equivalent to this?

PR #184, opened 2016-09-25
Added docker

dartharaknus:
Can't use alpine-node due to issues on windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "webpack": "NODE_ENV=production webpack -p --bail",
     "watch": "webpack --watch",
+    "prestart": "npm install",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "webpack": "NODE_ENV=production webpack -p --bail",
     "watch": "webpack --watch",
+    "prestart": "npm install",
```
howardchung: I don't see `npm run prestart` anywhere, and if this is meant to be run by the user, can't they just run `npm install` directly?

dartharaknus: No, it's like preinstall, prepublish thing for `start`.

howardchung: Ah okay, I'd forgotten that npm supports lifecycle scripts

PR #186, opened 2016-09-25
Mike | fix webpack for flexboxgrid & fix sourcemaps

mike-robertson:


webpack.config.js
```diff
@@ -28,15 +28,18 @@ const config = {
...
   },
+  devtool: 'eval-source-map',
   module: {
+    // We need to load flexboxgrid with css-modules, but others need to be loaded
+    // with regular css loader.
     loaders: [{
       test: /\.css$/,
       loader: 'style-loader!css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]!postcss-loader',
-      exclude: /node_modules/,
+      exclude: /node_modules\/?=flexboxgrid/,
```
howardchung: we can do this for now, but I'm not sure how maintainable this will be in the future (what if we find another dependency that needs this?)

howardchung:
CI failing

webpack.config.js
blukai: @howardchung https://github.com/odota/ui/pull/187/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR38 ❔

mike-robertson: yea, I'm not sure if there is a better way to do this, but we can engineer something if it ends up becoming complex.

webpack.config.js
```diff
@@ -97,6 +100,7 @@ HashBundlePlugin.prototype.apply = function f(compiler) {
 if (!isProd) {
   config.plugins.push(new DashboardPlugin());
 } else {
+  delete config.devtool;
```
howardchung: can we set config.devtool in !isProd instead of deleting it?

mike-robertson: I think it's more confusing to set things specifically for dev vs. for prod. But we can do it the other way. Let's merge this asap though and open an issue so I can use debugger :)

PR #187, opened 2016-09-25
assets structure, font fixes and smth

### LINKED ISSUE #177: Fix font in AccountWidget
Currently, @andrkrn 's AccountWidget is still using the default Roboto font, while most of the rest of the components use a new font.  We should make them match (and probably turn off uppercasing)

### LINKED ISSUE #182: avoid external google dependency
https://github.com/odota/ui/blob/master/src/components/palette.css#L1

We currently load these Google dependencies at runtime.  This will probably break in China.  Can we bundle it with webpack to avoid this issue?

blukai:
fixes #177 
fixes #182 
@howardchung, @mike-robertson review please.


src/components/Icons/Logout.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+export default (props) => (
+  <svg {...props} viewBox="0 0 300 300">
```
howardchung: should the svg go in its own file?

webpack.config.js
```diff
@@ -35,7 +35,7 @@ const config = {
       exclude: /node_modules/,
     }, {
       test: /\.css$/,
-      loader: 'style-loader!css-loader',
+      loader: 'style-loader!css-loader?modules',
```
howardchung: @mike-robertson will this work for the c3 import?

mike-robertson: I think this change should be discarded as it will break stuff such as c3

blukai: I just tried to fix flexboxgrid hehe

howardchung:
LGTM, will merge when @mike-robertson signs off

src/components/Icons/Logout.jsx
blukai: I was thinking to do this after adding propTypes

PR #188, opened 2016-09-25
Histograms

mike-robertson:
WIP. Not ready to merge but here are preliminary efforts.


mike-robertson:
ready for review @howardchung

package.json
```diff
@@ -7,7 +7,7 @@
     "webpack": "NODE_ENV=production webpack -p --bail",
     "watch": "webpack --watch",
     "prestart": "npm install",
-    "start": "webpack-dev-server -d --compress",
+    "start": "webpack-dev-server",
```
howardchung: why are the flags removed?

src/actions/player/playerHistogramActions.js
```diff
@@ -0,0 +1,66 @@
...
+    .then(json =>
+      Object.keys(json).reduce((columns, key) => {
+        columns.x.push(key);
+        columns[histogramName].push(json[key]);
+        return columns;
+      }, {
+        x: [],
+        [histogramName]: [],
+      }))
+    .then(json => bucketizeColumns(json[histogramName], json.x, histogramName))
```
howardchung: why not just pass json and histogramName here?

package.json
mike-robertson: They break source maps (I think).

src/actions/player/playerHistogramActions.js
```diff
@@ -0,0 +1,66 @@
...
+    .then(json => dispatch(getPlayerHistogramOk(json, playerId, histogramName)));
+    // .catch(error => dispatch(getPlayerHistogramError(error, playerId, histogramName)));
```
howardchung: why is the catch commented out?

src/components/Header/Pages.js
```diff
@@ -18,33 +18,33 @@ const navbarPages = [{
 }];
 
 const playerPages = [{
-  name: 'Overview',
+  name: 'Overview/',
```
howardchung: Why the trailing slash?

src/actions/player/playerHistogramActions.js
mike-robertson: I don't want to make too many assumptions about the Function signature so that it makes sense in other contexts if it's required.

src/components/Header/Pages.js
mike-robertson: I think I only needed that for histogram page. I'll change it.

src/actions/player/playerHistogramActions.js
```diff
@@ -0,0 +1,66 @@
+import fetch from 'isomorphic-fetch';
+import { API_HOST } from 'config';
+import { playerHistogram } from 'reducers';
+import { getUrl } from 'actions/utility';
+import { bucketizeColumns } from 'utility';
+
+// const excludedOptions = ['limit'];
+const url =
```
mike-robertson: getUrl expects a function to generate the url, so this carrying allows that to still work.

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -0,0 +1,74 @@
...
+  label: {
+    text: name,
+    show,
+  },
+});
+
+const selectHistogram = (router, histogramName, playerId) => {
+  router.push(`/players/${playerId}/histograms/${histogramName}`);
+};
+
```
howardchung: Does initialization work properly?  If the user directly enters a URL, does the appropriate histogram get loaded?

src/actions/player/playerHistogramActions.js
mike-robertson: Whoops. I do that during dev so that errors show up in console properly since there  is some weird js behavior with promises and errors. I'll change it.

src/components/Player/Pages/Histograms/Histograms.jsx
mike-robertson: Yep, it only works for  the url in fact.

mike-robertson: By that u mean it works when the route changes (which the button presses do)

src/components/TabBar/TabBar.jsx
```diff
@@ -16,7 +16,7 @@ const TabBar = ({ router, tabs, activeTab }) => (
     value={activeTab}
   >
     {tabs.map((tab, index) => (
-      <Tab key={index} className={styles.tab} value={tab.route} label={tab.label} onActive={tab => onActive(tab, router)}>
+      <Tab key={index} className={styles.tab} value={tab.route} label={tab.label.replace(/\//g, '')} onActive={tab => onActive(tab, router)}>
```
howardchung: I assume this is the reason the / was added to all the page names earlier, for some routing reason.

src/components/Visualizations/Graph/Graph.jsx
```diff
@@ -0,0 +1,74 @@
+import React, { Component } from 'react';
```
howardchung: rename this to histogram, if it's only one specific kind of graph

src/components/Player/Pages/Histograms/Histograms.jsx
mike-robertson: I* mean

src/components/Visualizations/Graph/Graph.jsx
mike-robertson: I'm actually going to extend this component further as more functionality is required so I don't have to deal with the library as much.

src/utility/bucketizeColumns.js
```diff
@@ -0,0 +1,47 @@
+const buckets = 40;
```
howardchung: add a comment describing what this function does

howardchung: rename to "max_buckets?"

src/components/TabBar/TabBar.jsx
mike-robertson: This just strips out the slashes for display purposes. I don't know if I actually need the slashes anymore. I'll take them out if I don't.

src/utility/bucketizeColumns.js
mike-robertson: Makes since (but camel cased probably).

src/utility/bucketizeColumns.js
```diff
@@ -0,0 +1,47 @@
...
+  if (columns.length <= buckets) {
+    return {
+      [histogramName]: columns,
+      x: xVals,
+    };
+  }
+  const max = Math.max(...xVals);
+  const bucketSize = ~~(max / buckets);
+  const newXVals = [];
+  let i = 0;
```
howardchung: a more functional solution:
`Array.from({length: buckets}, (v, k) => bucketSize * k);`

mike-robertson: Thanks, I couldn't figure out the right way to do this

src/utility/bucketizeColumns.js
```diff
@@ -0,0 +1,47 @@
...
+  for (i; i < buckets; i++) {
+    newXVals.push(bucketSize * i);
+  }
+  i = 0;
+  const newColumns = newXVals.map((val) => {
+    const newObj = { win: 0, games: 0 };
+    let enteredLoop = false;
+    while (i < xVals.length && xVals[i] <= val) {
```
howardchung: Could you achieve the same effect by using .filter() and then .reduce() for each bucket?

src/utility/bucketizeColumns.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (!enteredLoop) {
+      i++;
+    } else {
+      enteredLoop = false;
+    }
+    return newObj;
+  });
+  const removedIndices = [];
+  const filteredCols = newColumns.filter((obj, index) => {
```
howardchung: perhaps return the list of indices you want to keep, then run the same filter on both X and Columns?

src/utility/index.jsx
```diff
@@ -6,6 +6,11 @@ import { API_HOST } from 'config';
 import styles from 'components/palette.css';
 import { TableLink, TableHeroImage } from 'components/Table';
 import { KDA } from 'components/Visualizations';
+
```
howardchung: why does this need to be marked specifically as re-export?

webpack.config.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Webpack configuration file
  **/
-const DashboardPlugin = require('webpack-dashboard/plugin');
+// const DashboardPlugin = require('webpack-dashboard/plugin');
```
howardchung: remove from package.json if not considered a required part of the application.  Users who want to use it should `npm install -g`

src/utility/bucketizeColumns.js
mike-robertson: Yea I think so

src/utility/index.jsx
mike-robertson: Eh,  it doesn't. I can take it out

webpack.config.js
mike-robertson: Yea at one point I was going to add this but ran into some issues. It makes webpack cooler but makes Mac terminal messed up

src/components/Player/Pages/Histograms/Histograms.jsx
howardchung: If the user jumps directly into the SPA at a non-default route (i.e. not kills), the UI state is correct?

mike-robertson: Yep. You can test it out though to see how it behaves.
```
git checkout histograms
```

src/utility/bucketizeColumns.js
mike-robertson: I'd like to leave it as is for now, I can't think of a way of doing this right now without increasing runtime by a large amount (since this only does O(columns). If something comes to me that would be a more functional way without O(n^2) then I will reimplement.

src/utility/bucketizeColumns.js
mike-robertson: If I do it like that then I have to make an additional pass through the array. This way I can create the list as well as filter the xCols at the same time.

mike-robertson:
@howardchung I pushed changes based on comments. Let me know if anything important is missing.

package.json
howardchung: --compress should be fine.  It just gzips the bundle over the wire.

src/components/Header/Pages.js
```diff
@@ -35,6 +38,7 @@ const playerPages = [{
   name: 'Counts',
 }, {
   name: 'Histograms',
+  extra: histogramNames[0],
```
howardchung: instead of being defined here, shouldn't you set a default when you render the tabs/page?

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -0,0 +1,80 @@
...
+const Histogram = ({ histogramName, columns, router, playerId }) => (
+  <div style={{ fontSize: 10 }}>
+    <div className={styles.buttonContainer}>
```
howardchung: we will need a similar "field selector" for trends.  You might want to extract this into its own component.

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -0,0 +1,80 @@
...
+    text: name,
+    show,
+  },
+});
+
+const selectHistogram = (router, histogramName, playerId) => {
+  router.push(`/players/${playerId}/histograms/${histogramName}`);
+};
+
+const Histogram = ({ histogramName, columns, router, playerId }) => (
```
howardchung: shouldn't the default of kills be set here (as histogramNames[0])?

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -0,0 +1,80 @@
...
+    if (!this.props.histogramName) {
+      this.props.router.push(`/players/${this.props.playerId}/histograms/${histogramNames[0]}`);
+    } else {
+      getData(this.props);
+    }
+  }
+
+  componentWillUpdate(nextProps) {
+    if (!nextProps.histogramName) {
+      this.props.router.push(`/players/${this.props.playerId}/histograms/${histogramNames[0]}`);
```
howardchung: why is histogramsNames[0] hardcoded here?  Shouldn't it be the current route state?

src/components/Player/Pages/Histograms/histogramNames.js
```diff
@@ -0,0 +1,31 @@
+export default [
+  // 'kills',
```
howardchung: bug should be fixed so this is good to put back

src/components/Player/Player.jsx
```diff
@@ -14,19 +15,21 @@ import {
...
   CountsPage,
 } from './Pages';
 import TabBar from '../TabBar';
 import { playerPages } from '../Header/Pages';
 
-const playerPagesMapped = (accountId) => playerPages.map(page => ({
-  ...page,
-  route: `/players/${accountId}/${page.name.toLowerCase()}`,
-  label: page.name,
+const playerPagesMapped = (accountId) => playerPages.map(({ name, extra = '', ...rest }) => ({
```
howardchung: I don't think the "extra" param is needed here?

src/components/Player/Player.jsx
```diff
@@ -59,16 +64,20 @@ const Player = ({ playerId, info, params }) => {
...
     </div>
   );
 };
 // need to fix this
 
-const mapStateToProps = (state, { params }) => ({ playerId: params.account_id, info: params.info });
+const mapStateToProps = (state, { params }) => ({
+  playerId: params.account_id,
+  info: params.info,
+  histogramName: params.histogramName,
```
howardchung: trends is going to use a similar route pattern.  You might want to genericize the name.

src/components/Visualizations/Graph/Graph.jsx
```diff
@@ -0,0 +1,84 @@
...
+import c3 from 'c3';
+import uuid from 'node-uuid';
+import npmColor from 'color';
+
+const Graph = ({ id, height = 320 }) => (
```
howardchung: this 320 should probably be refactored into a constant.

src/components/Visualizations/Graph/Graph.jsx
```diff
@@ -0,0 +1,84 @@
...
+        ],
+        color: (color, data) => {
+          if (data.index || data.index === 0) {
+            const { index, value } = data;
+            const wins = columns[index] && columns[index].win;
+            if (!value) {
+              return npmColor().rgb(255, 255, 255);
+            }
+            const percent = wins / value;
+            const adjustedVal = percent >= 0.5 ?
```
howardchung: maybe explain what this is doing?  I'm not sure why you have to check percent === 0.5.

src/index.jsx
```diff
@@ -48,7 +49,14 @@ const store = createStore(reducer, compose(
 store.dispatch(getMetadata());
 // Create an enhanced history that syncs navigation events with the store
 const history = syncHistoryWithStore(browserHistory, store);
-history.listen(location => store.dispatch(setActiveTab(location.pathname)));
+history.listen(({ state, pathname }) => {
+  if (state && state.tabChange) {
+    store.dispatch(setActiveTab(pathname));
+  } else if (pathname.includes('histograms')) {
+    // I'd like to do this better if there is a way
+    store.dispatch(setActiveTab(`${pathname.slice(0, pathname.lastIndexOf('/'))}/${histogramNames[0]}`));
```
howardchung: I think you should just have the selector/histogram components check this.props.routeParams and render themselves accordingly.

src/index.jsx
```diff
@@ -62,7 +70,9 @@ render(
           <Route path=":info" />
         </Route>
         <Route path="players/:account_id" component={Player}>
-          <Route path=":info" />
+          <Route path=":info">
+            <Route path=":histogramName" />
```
howardchung: as mentioned previously, genericize this name

src/reducers/gotPlayer/counts.js
```diff
@@ -27,9 +27,22 @@ export const getPlayerCounts = {
     }
     return state.app.gotPlayer.counts.byId[id];
   },
-  getError: (state, id) => getPlayerCounts.getPlayerCountsById(state, id).error,
-  getLoading: (state, id) => getPlayerCounts.getPlayerCountsById(state, id).loading,
-  isLoaded: (state, id) => getPlayerCounts.getPlayerCountsById(state, id).loaded,
+  getOnlyData: (state, id) => {
```
howardchung: what is the purpose of this?

src/reducers/gotPlayer/histogram.js
```diff
@@ -0,0 +1,33 @@
...
+export const getPlayerHistogram = {
+  getPlayerHistogramById: (state, id) => {
+    if (!state.app.gotPlayer.histogram.byId[id]) {
+      return {
+        ...initialState,
+      };
+    }
+    return state.app.gotPlayer.histogram.byId[id];
+  },
+  getHistogram: histogramName => (state, id) => getPlayerHistogram.getPlayerHistogramById(state, id)[histogramName] || initialHistogramState,
```
howardchung: I think there will only ever be one histogram in state at any given time.  You can just make a request on each click and swap the histograms in state.

src/utility/bucketizeColumns.js
```diff
@@ -0,0 +1,45 @@
+const maxBuckets = 40;
```
howardchung: we don't need this function anymore?

src/components/Header/Pages.js
mike-robertson: One thing I tried earlier was routing to the first item in the histograms if the user went to /histograms, but the problem is that it breaks the back button (since pressing back goes to histograms, then reroutes them to histograms/kills). The solutions I thought of were either to modify back button behavior (which might be tricky if the go to multiple different histograms) or this one, and I thought this one was a bit simpler.

package.json
mike-robertson: ok, I'll test it out

src/components/Player/Pages/Histograms/Histograms.jsx
mike-robertson: I can do that, but then it won't be routed to the right place in the URL (it will be /histograms), and then if I do the trick I mentioned above then it will break the back button.

src/components/Player/Player.jsx
mike-robertson: see: [comment](https://github.com/odota/ui/pull/188/files/a4a36aa09aa20365261176d76b043ecf14fd2935#diff-c912dcb697b27f45c8c4cf4f06cc32f1)

src/components/Visualizations/Graph/Graph.jsx
mike-robertson: This is just changing the color scaling to be non-linear. The reason I check 0.5 is because otherwise, 50% will appear green vs. yellow due to this scaling. I'll add a comment.

src/index.jsx
mike-robertson: see: [comment](https://github.com/odota/ui/pull/188/files/a4a36aa09aa20365261176d76b043ecf14fd2935#diff-c912dcb697b27f45c8c4cf4f06cc32f1)

src/reducers/gotPlayer/histogram.js
mike-robertson: this way is better

src/reducers/gotPlayer/counts.js
mike-robertson: I had to refactor my reducerFactory method I was using for counts (which is a weird case since it is multiple tables in one request) to also work for histograms. Part of that meant changing the data structure of the reducer. This selector strips out the loading information from counts which means that I can still iterate over the object keys which are different counts tables to generate those tables.

src/components/Player/Pages/Histograms/Histograms.jsx
mike-robertson: This is checking if when the component is updating, does it have a histogram name which is truthy as a prop. So if the user goes to /histograms, they are redirected to /histograms/kills. This has the issue mentioned in this [comment](https://github.com/odota/ui/pull/188/files/a4a36aa09aa20365261176d76b043ecf14fd2935#diff-c912dcb697b27f45c8c4cf4f06cc32f1), but I think it's acceptable because it's an edge case (vs when people click on the histogram tab, which is why I refactored that to be different).

src/components/Header/Pages.js
howardchung: Instead of rerouting, what if you just treat empty string (/histograms) as the default internally without changing the URL?  (this is the current behavior)

src/components/Player/Pages/Histograms/Histograms.jsx
howardchung: I think this is acceptable.  /histograms/kills will still work and this is the current behavior.

mike-robertson: No the current site doesn't break the back button. The user needs to be able to navigate backwards

src/components/Header/Pages.js
mike-robertson: I could do that, but I will still need all this code because otherwise we can't highlight the tab in the tabbar if they click on any of the buttons. The only way to highlight the tab is to check the value, so if they click on any other histogram or route to a histogram directly, the tab in the player bar won't show as active on histograms. I think it's ok for this to be like this as I can see lots of scenarios where tabs will have routing inside them but we still want that tab to show up as highlighted.

package.json
howardchung: does --compress work ok?

src/components/Player/Player.jsx
```diff
@@ -38,6 +41,8 @@ const getPlayerSubroute = (info, playerId) => {
       return <ProsPage playerId={playerId} />;
     case 'rankings':
       return <RankingsPage playerId={playerId} />;
+    case 'histograms':
+      return <HistogramsPage playerId={playerId} histogramName={subInfo} />;
```
howardchung: do you still need this passed if the histogram components can read state from the route?

src/components/Player/Player.jsx
```diff
@@ -14,19 +16,20 @@ import {
...
 
-const playerPagesMapped = (accountId) => playerPages.map(page => ({
-  ...page,
-  route: `/players/${accountId}/${page.name.toLowerCase()}`,
-  label: page.name,
+const playerPagesMapped = (accountId) => playerPages.map(({ name, extra = '', ...rest }) => ({
+  ...rest,
+  route: `/players/${accountId}/${name.toLowerCase()}${extra ? `/${extra}` : ''}`,
+  label: name,
+  extra,
```
howardchung: this "extra" param is not needed if histogram components can access route state?

src/components/Player/Player.jsx
```diff
@@ -49,35 +54,38 @@ const getPlayerSubroute = (info, playerId) => {
   }
 };
 
-const Player = ({ playerId, info, params }) => {
-  if (!playerId) {
+const Player = ({ params: { accountId, info, subInfo } }) => {
```
howardchung: why are we wrapping the inputs in a params argument here?

src/components/Player/Player.jsx
mike-robertson: Yea I have to pass that since the player is what has access to params

src/components/Player/Player.jsx
mike-robertson: yes

src/components/Player/Player.jsx
mike-robertson: this is destructuring syntax for destructuring multiple levels (in this case, props.params.WhatIWant). It seems like only the component being given to a `<Route component={MyComponent} />` seems to be given the params as props. I haven't found a HOC that will give them to components yet. I'm sure I could make one and store the stuff in redux but it seems unnecessary for now since I'm only passing a couple levels deep. If I have to pass more levels deep then I will probably do that.

src/components/Player/Player.jsx
howardchung: Doesn't react router inject the route into every component's props?

mike-robertson: I didn't think so but I can try it out later. I'll let you know once I see.

howardchung: I think you might have to connect it to the store to get it.  If you don't want to do that, passing here is fine.

mike-robertson: It says in the docs that you shouldn't get params that way due to how react-router is async and doesn't always align with the store. It says you can use mapStateToProps in an example, but all they are doing is pulling the params from the props (which is unnecessary as they already exist as props). https://github.com/reactjs/react-router-redux#how-do-i-access-router-state-in-a-container-component

PR #205, opened 2016-09-29
Refactor routes and store and index.jsx

mike-robertson:


mike-robertson:
Please review: @howardchung

src/index.jsx
```diff
@@ -1,79 +1,21 @@
...
-        </Route>
-        <Route path="request" component={Request} />
-        <Route path="distributions" component={Distributions} />
-        <Route path="carry" />
-        <Route path="status" />
-        <Route path="*" component={FourOhFour} />
-      </Route>
-    </Router>
-  </Provider>, reactElement);
+render(<RouterWithProvider />, reactElement);
```
howardchung: maybe pass the store to the RouterWithProvider instead of re-importing it there since you already need a store reference here anyway?

PR #207, opened 2016-09-29
styles and some fixes, new cheese icon may appear 🦄

### LINKED ISSUE #202: 404 error when loading fonts
It appears that the browser is firing requests to load fonts, and it is case sensitive.  Repro: Visit any page on the production SPA

Requests:
http://beta.opendota.com/assets/fonts/PTSans/pts75f_w.woff
http://beta.opendota.com/assets/fonts/PTSans/pts75f_w.ttf

Shouldn't these be bundled with webpack?

blukai:
- [x] New cheese icon 🦄
- [x] Tooltips for table headers in player profiles

fixes #202


src/components/PercentContainer/PercentContainer.css
```diff
@@ -1,6 +1,6 @@
 @import "../palette.css";
 
-:export { progressColor: var(--accentColor) }
+:export { progressColor: var(--colorYelor) }
```
howardchung: you might just want to call this colorYellowOrange to make it clearer

blukai:
Cheese badge
image

Also badge can look like:
image

blukai:
> Do we have HTML we need to set directly? I know we have it for abilities/items, but not for table headers?

Here https://github.com/odota/ui/blob/master/src/lang/en.json#L105

howardchung:
I think we should avoid using HTML in the strings.  Would \n work?

blukai:
Unfortunately no

howardchung:
I'm fine with just not using line breaks then.  Maybe we can write it more concisely: "Approximate MMR cutoffs for the brackets are 0, 3200 and 3700"

nicholashh:
works on my machine, LGTM

PR #209, opened 2016-09-30
Remove moment dependency

howardchung:
Currently we are only using it for its .fromNow() function.  I wrote my own implementation.

* Much smaller footprint
* Easier to apply localization in the future


blukai:
Maybe this will be better? https://gist.github.com/skattyadz/1285806

howardchung:
My only concern is plurals.  I intentionally avoided plurals since I don't know in all languages if plurals can be defined just by appending a string.

PR #217, opened 2016-10-01
Pages list cleanup

dartharaknus:


src/components/Match/Match.jsx
```diff
@@ -105,12 +106,12 @@ class RequestLayer extends React.Component {
       <div>
         <MatchHeader match={match} user={this.props.user} />
         <Tabs>
-          <Tab label="Overview">
+          <Tab label={strings.tab_overview}>
```
howardchung: we are likely to move to using the same TabBar component that we use for player pages, so this structure will likely change.  We'll probably define the pages in an array (like how we do player page list).

src/components/Header/Pages.js
```diff
@@ -1,6 +1,6 @@
 import strings from 'lang';
```
howardchung: if this is just going to export navbar pages, it should probably be renamed.  Or just inline the pages in the component that uses it (if there's only one).

dartharaknus: No, there is Navbar component which is using this. I left the name as `Pages` because it is nested under `Header` and thus should be okay.

src/components/Match/Match.jsx
dartharaknus: Something for later then.

src/components/Header/Pages.js
howardchung: The NavBar component is actually not in use at this time.  You could remove it completely and inline the pages in the Header.

dartharaknus: Done. I had to remove Navbar.

PR #219, opened 2016-10-01
Started on Items UI

dartharaknus:
Need to wait for items endpoint.


howardchung:
Not sure if items will make it into MVP, because I want to change how the data is presented.  We can keep this open until we have an idea of what we want to do.

howardchung:
@nicholashh how can it work for you when the server side API isn't implemented? :P

dartharaknus:
LOL. @howardchung Can we have a simple server side API implemented which just lists 0 items for now. We can improve on that once this PR is merged.

howardchung:
thats probably equivalent to just making the action return `[]` instead of actually making the http call.

dartharaknus:
Hmm.. I think I will do that.

dartharaknus:
Items return `[]` for now.

Issue #152, opened 2016-09-22
Standardize table headings


howardchung:
I think we should establish that all Tables have a heading, with the style established by @blukai.

We can do this by adding an additional "heading" property string to the Table component, which renders the heading as part of the returned JSX.  (to keep the Table general purpose, the property can be omitted to generate no heading, but in our code we should always use it).

We should also standardize on whether we want to all caps the headings (currently it's a mix)

PR #220, opened 2016-10-02
implement match graphs

howardchung:
Reviewers:
@mike-robertson 
@nicholashh 
@albertcui


nicholashh:
lacking graph titles and gold sources table. would you like me to add?

howardchung:
I don't think we'll be doing the gold sources table.

Yes, you could add table headers but it will be part of a larger issue where we create a header component that follows @blukai 's style.

https://github.com/odota/ui/issues/152

src/components/Match/Match.jsx
```diff
@@ -29,60 +37,30 @@ import {
...
-const getMatchPlayers = (state) => state[REDUCER_KEY].match.match.players;
-const getSortState = (state) => state[REDUCER_KEY].match.sortState;
-const getSortField = (state) => state[REDUCER_KEY].match.sortField;
-const getSortFn = (state) => state[REDUCER_KEY].match.sortFn;
+const match = (state) => state.app.match;
+const getMatchPlayers = (state) => state.app.match.match.players;
+const getSortState = (state) => state.app.match.sortState;
+const getSortField = (state) => state.app.match.sortField;
+const getSortFn = (state) => state.app.match.sortFn;
 const sortMatchPlayers = (state) => defaultSort(getMatchPlayers(state), getSortState(state), getSortField(state), getSortFn(state));
```
mike-robertson: you might want to move these selectors to the reducer since the match reducer is what cares about the state shape, not the component.

src/components/Match/Match.jsx
```diff
@@ -29,60 +37,30 @@ import {
...
+  params,
+}) => ({
   matchId: params.match_id,
-  match: state[REDUCER_KEY].match.match,
-  loading: state[REDUCER_KEY].match.loading,
-  user: state[REDUCER_KEY].metadata.user,
+  match: state.app.match.match,
+  loading: state.app.match.loading,
+  user: state.app.metadata.user,
 });
```
mike-robertson: use selector if you can (not necessarily reselect, but what I do in the player reducers). same reasoning as above.

src/components/Match/renderMatch.js
```diff
@@ -250,28 +247,33 @@ function renderMatch(m) {
...
-    data.lh.push([hero].concat(p.lh_t));
-  });
-  return data;
+  if (match.players) {
+    // compute graphs
+    let goldDifference = ['Gold'];
+    let xpDifference = ['XP'];
+    goldDifference = goldDifference.concat(match.radiant_gold_adv);
+    xpDifference = xpDifference.concat(match.radiant_xp_adv);
+    const time = ['time'].concat(match.players[0].times);
```
mike-robertson: instead of doing all these concats, you could do 
```js
const goldDifference = ['Gold', ...match.radiant_gold_adv];
```
this isn't necessary but is a little more concise and easier to understand.

PR #224, opened 2016-10-02
add tabbar for matches

### LINKED ISSUE #152: Standardize table headings

howardchung:
fixes #152 

* Use TabBar component for matches
* Restructure pages for matches/players
* Externalize some strings
* Add headings to some elements

Reviewers:
@mike-robertson 
@nicholashh 
@albertcui


src/components/Player/Player.jsx
```diff
@@ -21,63 +26,82 @@ import {
...
+    info,
+    subInfo,
+  },
+}) => {
   if (!accountId) {
     return <Error />;
   }
-// Need to pass in the action into filter form, need to put that filter form into each subroute as well
+  const defInfo = info || 'overview';
+  // Need to pass in the action into filter form, need to put that filter form into each subroute as well
```
mike-robertson: use default params, don't use this syntax.

src/components/Player/Player.jsx
```diff
@@ -21,63 +26,82 @@ import {
...
   return (
     <div>
       <div className={styles.header}>
         <PlayerHeader playerId={accountId} />
         <div style={{ marginTop: 25 }}>
           <TabBar
-            playerId={accountId}
-            activeTab={info}
+            info={defInfo}
             subInfo={subInfo}
```
mike-robertson: if you use default params you don't need to do this

src/components/Player/Player.jsx
```diff
@@ -21,63 +26,82 @@ import {
...
             tabs={playerPagesMapped(accountId)}
           />
         </div>
       </div>
-      {getPlayerSubroute(info, accountId, subInfo)}
+      {playerPagesMapped(accountId).filter(page => page.name.toLowerCase() === defInfo).map(page => page.content(accountId, subInfo))}
     </div>
```
mike-robertson: This feels way more confusing to me than the switch case statement. I'd rather keep it the same way unless there is some sort of reason this is better besides personal preference.

src/components/Player/Player.jsx
```diff
@@ -21,63 +26,82 @@ import {
...
   MMRPage,
-  playerPages,
 } from './Pages';
 
-const playerPagesMapped = (accountId) => playerPages.map(({ name, ...rest }) => ({
-  ...rest,
-  route: `/players/${accountId}/${name.toLowerCase()}`,
-  label: name,
-}));
+const playerPages = [{
```
mike-robertson: putting this configration object into Player.jsx is also something I would like to avoid. I don't mind the switch case because it's essentially routing, and it's very intuitive what the code is doing. A big configuration object feels worse. Also, there is no reason to put `()` around all your return statements with jsx. It's just something that's needed for multiline jsx (and that may not even be necessary, but makes it easier to read).

src/components/Player/Player.jsx
```diff
@@ -1,10 +1,15 @@
 import React from 'react';
-import { connect } from 'react-redux';
-import { withRouter } from 'react-router';
+import {
+  connect,
+} from 'react-redux';
```
mike-robertson: please don't do this for single imports, only for multi imports.

mike-robertson:
Approved pending comments

src/components/Player/Player.jsx
howardchung: I prefer this because it removes the duplication of page names.

src/components/Player/Player.jsx
howardchung: My autoformatter automatically does this.  I'm personally okay with it because it makes it easier to add to if we ever add a second import.  It still passes eslint

PR #229, opened 2016-10-03
Match tables

### LINKED ISSUE #138: port renderMatch and analysis to ES6
Currently ESLint is disabled on these two Match methods since they are ports from core.  In the future we should remove the disable and correctly convert the functions to ES6.

howardchung:
fixes #138 

* Gold/xp reasons tables
* Log (kills + objectives)
* Heatmap POC (single player only)
* Analysis table (dumping data only, needs better styling)
* Teamfights tables (tabbed)
* Damage inflictors tables

Reviewers:
@albertcui 
@nicholashh 
@mike-robertson


src/components/Match/Heatmap.jsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+  componentWillUpdate(nextProps) {
+    drawHeatmap(nextProps, this.id);
+  }
+
+  render() {
+    return (
+      <div
+        id={this.id}
+        style={{
```
howardchung: this style probably doesn't matter (copied from buildingMap)

howardchung:
@blukai I disabled individual table column widths in this PR.  In most cases this works well, however on the player overview the hero image cell is running over into the next one without the column stretching to fit.  I tried to fix this but couldn't find a solution, maybe you can figure it out?

blukai:
I'll see what i can do

blukai:
What is the reason disabling individual table column widths? Now it looks like every column in table has the same width since `table-layout: fixed`, it can be changed to `auto `but will make another problem since `thead` and `tbody` in different tables => headings and content will have different widths.

howardchung:
Having to manually specify column widths is annoying and can break (what if someone has a really long name, etc.)  Why are thead and tbody in different tables?

blukai:
https://github.com/callemall/material-ui/blob/master/src/Table/Table.js#L270 
Thahs why.

Adding this `fixedHeader={false} style={{ tableLayout: 'auto' }}` [here](https://github.com/odota/ui/blob/master/src/components/Table/Table.jsx#L23) will solve the problem

src/components/Match/matchColumns.jsx
```diff
@@ -483,72 +444,153 @@ const unitKillsColumns = [
...
-  abilityUseColumns,
-  itemUseColumns,
-  unitKillsColumns,
-  actionsColumns,
-  lastHitsTimesColumns,
-  runesColumns,
-  cosmeticsColumns,
-  objectiveDamageColumns,
-  objectiveLogColumns,
-};
```
mike-robertson: this file is very large. Maybe we can split it up at some point?

src/components/Match/renderMatch.js
```diff
@@ -19,11 +17,124 @@ Object.keys(specific).forEach((key) => {
...
+  const computeTfData = (tf) => {
+    const newtf = {};
+    newtf.posData = [];
+    newtf.radiant_gold_delta = 0;
+    newtf.radiant_xp_delta = 0;
+    newtf.radiant_participation = 0;
+    newtf.radiant_deaths = 0;
+    newtf.dire_participation = 0;
+    newtf.dire_deaths = 0;
+    newtf.players = m.players.map(p => {
```
mike-robertson: might save you some characters if you do the object declaration more like
```js
const newtf = {
  prop: val,
  prop2: val2,
};
```

src/components/Match/renderMatch.js
```diff
@@ -19,11 +17,124 @@ Object.keys(specific).forEach((key) => {
...
+        newtf.radiant_deaths += tfplayer.deaths ? 1 : 0;
+      } else {
+        newtf.radiant_gold_delta -= tfplayer.gold_delta;
+        newtf.radiant_xp_delta -= tfplayer.xp_delta;
+        newtf.dire_participation += tfplayer.participate ? 1 : 0;
+        newtf.dire_deaths += tfplayer.deaths ? 1 : 0;
+      }
+      return {
+        ...p,
+        ...tfplayer,
```
mike-robertson: p is probably player or something, but in order for me to figure out what it is, I have to scroll all the way up the file.

src/components/Match/renderMatch.js
```diff
@@ -158,145 +189,71 @@ function renderMatch(m) {
...
-          y: 127 - (Number(y) - 64),
-          value: p[key][x][y],
-        });
-      }
     }
-    d[key] = t;
-  }
-  return d;
+    if (pm.killed) {
+      additionalProps.specific = {};
```
mike-robertson: same comments regarding variable verbosity.

src/utility/index.jsx
```diff
@@ -99,6 +90,13 @@ export const getPercentWin = (wins, games) => (games ? Number(((wins * 100) / ga
 export const camelToSnake = str =>
   str.replace(/\.?([A-Z]+)/g, (match, group) => `_${group.toLowerCase()}`).replace(/^_/, '');
 
+export const getOrdinal = (n) => {
+  // TODO localize strings
+  const s = ['th', 'st', 'nd', 'rd'];
+  const v = n % 100;
+  return n + (s[(v - 20) % 10] || s[v] || s[0]);
+};
```
mike-robertson: Same comment. I know these are basic functions, but it's hard to understand the intent of the code if I have to guess at what the variables mean.

src/actions/matchActions.js
```diff
@@ -39,7 +38,6 @@ export const getMatch = (matchId) => (dispatch) => {
   dispatch(getMatchRequest());
   return fetch(`${API_HOST}${url}${matchId}`)
     .then(response => response.json())
-    .then(json => renderMatch(json))
```
mike-robertson: This kind of transformation should happen here probably, not in the reducer.

src/reducers/match.js
```diff
@@ -25,7 +26,7 @@ export default (state = initialState, action) => {
         ...state,
         loading: false,
         error: false,
-        match: action.payload,
+        match: renderMatch(action.payload),
```
mike-robertson: I think this kind of transformation should not happen in the reducer, but rather the action. That way, if something goes wrong in renderMatch in the future, we can understand and know about that by inspecting the action it is dispatching in redux debugging tools.

Issue #228, opened 2016-10-02
trends bugfixes


howardchung:
- [ ] data needs to be reversed before computing cumulative average (oldest to newest)
- [ ] filter out undefined values (for parsed fields on unparsed matches)

Roshanjossey:
Hey @mike-robertson: I'm working on this issue. 
Currently looking in to `https://github.com/odota/ui/blob/master/src/selectors/trends/getPlayerTrend.js`

mike-robertson:
Hey Roshan. I appreciate the help but this was already an issue we solved. It simply needed to have the PR created (which I have done now). Please let me know if you have any questions about other issues!

Roshanjossey:
@mike-robertson: Thank you for your reply. I'm also looking in to #81 and #104.

I think it'll help a lot if we attach screen shots for those issues and while sending PRs.

Anyways lets take the conversation in those threads.

PR #230, opened 2016-10-03
fix issue #228

mike-robertson:
fixes issue #228


src/selectors/trends/getPlayerTrend.js
```diff
@@ -7,6 +7,8 @@ const getMatches = fieldName =>
     state =>
       playerMatches
         .getMatchList(state, id)
+        .filter(match => match[fieldName] !== undefined)
```
howardchung: should handle null too

PR #232, opened 2016-10-04
More match implementation

howardchung:
* Play with vision map
* Party indicator
* Remove unused actions/reducers
* (MERGED) airbnb12 compliance


src/components/Match/VisionMap.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+export default function BuildingMap({
```
howardchung: correct name

src/components/Match/renderMatch.js
```diff
@@ -191,7 +191,7 @@ function renderMatch(m) {
   });
   */
   const newPlayers = m.players.map((player) => {
-    const additionalProps = {
+    const newPlayer = {
```
nicholashh: yes, very much thank you

PR #240, opened 2016-10-06
import dotaconstants json directly

howardchung:
* By importing the specific files we skip the overhead of bundling the entire object into the result.
* This is basically tree shaking

### DIFF
--- a/src/components/Form/TableFilter.config.js
+++ b/src/components/Form/TableFilter.config.js
@@ -1,3 +1,4 @@
+/*
 import {
   heroes,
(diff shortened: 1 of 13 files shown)


src/components/Form/TableFilter.config.js
```diff
@@ -6,6 +7,13 @@ import {
   lobby_type as lobbyType,
   region,
 } from 'dotaconstants';
+*/
```
dartharaknus: You should remove these comments. Dead code is bad code.

PR #241, opened 2016-10-06
distributions country table

### LINKED ISSUE #196: Put country mmr distribution data in table

howardchung:
fixes #196

### DIFF
--- a/src/components/Cheese/CheeseCircle.css
+++ b/src/components/Cheese/CheeseCircle.css
@@ -19,11 +19,11 @@
 }
 /*.front svg > circle {
   stroke: var(--primaryLinkColor) !important;
-}*/
+}
 .back svg > circle {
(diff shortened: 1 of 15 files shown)


src/components/Cheese/CheeseCircle.css
```diff
@@ -19,11 +19,11 @@
 }
 /*.front svg > circle {
   stroke: var(--primaryLinkColor) !important;
-}*/
+}
 .back svg > circle {
   stroke: rgba(255, 255, 255, .06) !important;
 }
-
+*/
```
blukai: `.back` making gray background until the goal is fully achieved. Without it would seem that the ring is completely filled. `.front svg > circle` can be removed since color already set https://github.com/odota/ui/blob/master/src/components/App/App.jsx#L22

howardchung: whoops reverted this

Issue #81, opened 2016-07-19
mobile issues


howardchung:
- [ ] navbar runs off the screen instead of stacking
- [ ] home opening text runs offscreen
- [x] player header elements overflow vertically
- [x] TabBar navigation squishes together horizontally instead of wrapping
- [x] overview match/hero tables are squished together (they should stack if the viewport is too narrow?  bootstrap handles this automatically)
- [x] home three features section should stack instead of compressing into 1/3 width
- [ ] Tables are way too wide in mobile scroll view.

Roshanjossey:
![image](https://cloud.githubusercontent.com/assets/8488446/19073566/44c2c26a-8a59-11e6-8a4a-c26abab37079.png)

@mike-robertson: I've annotated the issues mentioned in the screenshot I added above.
Could you please check if I've got them right? Also have we decided on what to do about
>overview match/hero tables are squished together (they should stack if the viewport is too narrow? bootstrap handles this automatically)

or is that an open question?

mike-robertson:
Let's pull @blukai into this discussion. My thinking is 
Navbar: add in hamburger menu for mobile
Player Header: shrink picture and stack mmr/ match info under player name (that can't fit next to name) 
Tables: auto width and stack, scrolling x
Tabbar: allow to run offscreen and swipe/scroll

Roshanjossey:
Cool @blukai: I'll get started with Tabbar. Also when I run
`npm run lint`, I get
>`> opendota-ui@1.0.0 lint /Users/roshan/open/ui`
>`> eslint . --ext .js,.jsx --fix`

>`The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint->plugin-import instead.`

With no errors. Is this how it's supposed to be?

blukai:
@Roshanjossey, eslint just checks the js,  jsx errors by a config, for example if u **skip** semicolon on second line
```js
function mapStateToProps(data) {
  return { content: data.content };
}
``` 
these will be an error.

And I guess navbar items can be wrapped in dropdown.

howardchung:
added a couple of issues with homepage on mobile

howardchung:
replacing with individual issues:
#295 
#278 

The second is fixed by PR #252

PR #243, opened 2016-10-06
issues #226, #227, #155

### LINKED ISSUE #155: rankings display fixes
http://beta.opendota.com/players/88367253/rankings

- [x] last played always shows as "never".  I think this is due to not returning the last played date on this endpoint.  We should have an option on the hero cell to hide this value.
- [x] We should make the rank an ordinal number.  There's a utility function in core to do this (or just Google for it)

- [x] Current UI has context coloring based on the percentile (green/blue/dark blue/yellow/red).  We can add this to the new UI.

- [x] Hero should link to rankings page

- [x] Rank column should only add the english in the display function. Otherwise sorting is as a string and doesn't sort properly.

### LINKED ISSUE #226: Players/Pros tabs should use combined cell
Similar to how the hero table works, the player and last played time should be combined and the sort should be on the last played time.

Pros should also display image (I think backend has been updated for this)

### LINKED ISSUE #227: Records should use combined cell
I think the Set column can be removed and combined with the hero cell.  We can make the record value column link to the match instead?

blukai:
fixes #226 
fixes #227 
fixes #155
image


src/lang/en.json
```diff
@@ -320,6 +320,7 @@
   "tooltip_biggest_hit": "Largest instance of damage on a hero",
   "tooltip_damage_dealt": "Damage dealt to heroes by items/abilities",
   "tooltip_damage_received": "Damage received from heroes by items/abilities",
+  "tooltip_registered_user": "Registered user",
```
blukai: Ye, but not now, 🍀  I'll plan to make some changes to player header that fixes this https://github.com/odota/ui/issues/81#issuecomment-251370349. Then I will add strings..

blukai:
fixes #227 
image

blukai:
almost #155
image

https://github.com/odota/ui/issues/155#issuecomment-252042343

src/lang/en.json
```diff
@@ -53,6 +53,9 @@
   "error": "Error",
   "error_message": "Whoops! Something went wrong.",
   "error_four_oh_four_message": "The page you are looking for cannot be found.",
+
+  "unknown": "Unknown",
```
howardchung: these strings should probably be namespaced under `general_` and moved into alphabetical order (by namespace).

src/lang/en.json
```diff
@@ -335,5 +339,8 @@
   "teamfight_score": "Radiant kills/Dire kills",
   "teamfight_radiant_gold_adv": "Net gold advantage for Radiant",
   "teamfight_radiant_xp_adv": "Net XP advantage for Radiant",
-  "anonymous": "Anonymous"
+  "anonymous": "Anonymous",
+
+  "radiant": "Radiant",
```
howardchung: same with these strings

src/utility/index.jsx
```diff
@@ -124,30 +118,67 @@ export const getOrdinal = (n) => {
...
 const getSubtext = (row) => {
   if (row.match_id && row.player_slot !== undefined) {
-    // TODO localize
-    return isRadiant(row.player_slot) ? 'Radiant' : 'Dire';
+    return isRadiant(row.player_slot) ? strings.radiant : strings.dire;
   }
   if (row.last_played) {
     return <FromNowTooltip timestamp={row.last_played} />;
   }
+  if ((row.name && row.value) !== undefined) {
```
howardchung: why not just check start_time?  If you don't want to render any subtext, that should be determined by the parent component.

src/utility/index.jsx
```diff
@@ -124,30 +118,67 @@ export const getOrdinal = (n) => {
...
+    return <FromNowTooltip timestamp={row.start_time} />;
+  }
   return null;
 };
 
+const percentColor = (pct) => {
```
howardchung: If we are using the function for multiple types of bucketing, maybe give it a more generic name (without "Color").

For analysis, we currently use letter grades, but I like the sentiment google material icons:
https://design.google.com/icons/

blukai: Hey, we can use dota emoticons instead if letters or gm icons. I'll do it after a while 😺

src/components/Visualizations/Table/Player.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
blukai: I think yes. 🍀

blukai:
Not yet ready for merge 🐰

howardchung:
ready for merge?

howardchung:
This LGTM, I'll merge (if there were any other changes you wanted to make, please open a new PR)

PR #246, opened 2016-10-06
Divide match overview into team specific tables

dartharaknus:


dartharaknus:
There are 2 things to note here.

1. We will also provide an option to merge all the players together but default should be divided along teams.
2. I am currently thinking of making this whole table/sort thing easier and smarter. So, this PR is only a first step.

src/reducers/match.js
```diff
@@ -51,7 +51,12 @@ export default (state = initialState, action) => {
 export const getMatch = state => state.app.match;
 export const getMatchData = state => state.app.match.match;
 export const getMatchLoading = state => state.app.match.loading;
-export const getMatchPlayers = state => state.app.match.match.players;
+
+export const getMatchPlayers = (state, team = '') =>
```
howardchung: I would move this filtering to the Match component.  Also instead of using string comparison I'd probably just take an isRadiant boolean parameter.

dartharaknus: The boolean wont work since I need it for 3 things, all, radiant and dire.

Why do I move this function to Match component but not the others?

howardchung: Don't move the entire function, but move the logic dealing with team filtering.  I think the selectors here should just get and return pieces of state and not have any further logic than that.

dartharaknus:
Updated.

PR #253, opened 2016-10-09
[mobile] Overview match/hero tables

blukai:
> overview match/hero tables are squished together (they should stack if the viewport is too narrow? bootstrap handles this automatically)

@Roshanjossey, I want to know your point of view about it.


howardchung:
@blukai do you want me to merge or wait for @Roshanjossey to review?

blukai:
Let's wait a bit. Anyway I guess I made a mistake by hiding some columns. We shouldn't hide content.

howardchung:
Yeah, I would prefer it to be horizontally scrolling rather than completely remove columns.

Roshanjossey:
@blukai, @howardchung: I might need more time to get enough context on this. I'll be able to get back to you by EOD today. (Monday 10th October).
@blukai: I'm happy that you gave me this responsibility 👍
Please ping me if its an urgent change and if we need to close this sooner.

src/utility/index.jsx
```diff
@@ -224,6 +224,20 @@ export const transformations = {
...
+  skill_and_duration: row => (
+    <div>
+      {(skill[row.skill] ? skill[row.skill] : strings.general_unknown)}
+      <span className={subTextStyle.subText}>
+        {formatSeconds(row.duration)}
+      </span>
+    </div>
+  ),
+  kda_together: row => <KDA kills={row.kills} deaths={row.deaths} assists={row.assists} together />,
+  summary: (
```
blukai: Do not merge, I'll finish it later 😆

blukai:
880px:
image

700px:
image

600px:
image

howardchung:
Is the table still scrollable to view the hidden columns?

blukai:
😆 I even didn't knew, that they were scrollable. Anyway scrollable tables are awful.

howardchung:
Between hiding columns and making it scrollable, I'd prefer the latter.

src/components/Player/Pages/Matches/Matches.css
```diff
@@ -0,0 +1,16 @@
+@media only screen and (max-width: 880px) {
+  th:nth-child(4), td:nth-child(4),
+  th:nth-child(5), td:nth-child(5),
+  th:nth-child(8), td:nth-child(8),
+  th:nth-child(9), td:nth-child(9),
+  th:nth-child(10), td:nth-child(10) {
+    display: none;
```
mike-robertson: This doesn't feel like a sustainable solution to this problem

src/components/Player/Pages/Overview/Overview.css
```diff
@@ -1,19 +1,17 @@
...
+    width: 100% !important;
+    margin: 0;
+  }
 }
 
 .overviewMatches {
-  width: 75%;
-  display: inline-block;
+  width: calc(70% - 30px);
+  margin-right: 30px;
```
mike-robertson: I think we should use a media query to make the tables stack at smaller screen sizes (tablet landscape or tablet portrait?)

mike-robertson: I just realized you have that already. My bad.

src/components/Player/Pages/Overview/Overview.css
```diff
@@ -1,19 +1,17 @@
 @import "../../../palette.css";
 
-.overviewContainer {
-  display: flex;
+.overviewMatches, .overviewHeroes {
+  display: inline-block;
+  @media only screen and (max-width: 1180px) {
+    width: 100% !important;
```
mike-robertson: We shouldn't need important here.

src/components/Player/PlayerStats.jsx
```diff
@@ -31,7 +31,7 @@ export const PlayerStatsCards = ({
     return <Spinner />;
   }
   return (
-    <div style={{ position: 'absolute' }}>
+    <div>
```
mike-robertson: Can you merge in the player header fixes branch because we're going to have a ton of merge conflicts

src/components/Table/Table.css
```diff
@@ -3,15 +3,14 @@
 .container {
   min-width: 100%;
   clear: both;
-  @media only screen and (max-width: 960px) {
-    margin: 0 -25px;
+  @media only screen and (max-width: 700px) {
```
mike-robertson: Leave the negative margin (unless you reduced the overall margin at small screen sizes) , we shouldn't have such large margins in mobile view

src/components/Table/TableContainer.jsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import Heading from 'components/Heading';
 import styles from './TableContainer.css';
 
-const TableContainer = ({ title, style, children }) => (
-  <div className={styles.container} style={{ ...style }}>
+const TableContainer = ({ title, style, children, className }) => (
+  <div className={className} style={{ ...style }}>
```
mike-robertson: This should have a default style, it shouldn't be required to pass in a className. That className should just override the default values.

src/components/Visualizations/Table/KDA.jsx
```diff
@@ -5,16 +5,16 @@ import palette from 'components/palette.css';
...
 
   return (
     <div className={styles.container}>
       <div>
-        {kills}
+        {together ? `${kills}/${deaths}/${assists}` : kills}
       </div>
       <div>
-        <div className={styles.percent} data-tip data-for={`kda-${matchId}`}>
+        <div className={styles.percent} data-tip data-for={`kda-${matchId}`} style={{ width: together && '100%' }}>
```
mike-robertson: Could you do a ternary on className instead of using in line styles? So compose the existing class in the css to a class called percentTogether which overrides width to 100%,  then do a ternary on the classname

mike-robertson:
I don't think we should merge this until comments have been addressed

src/components/Table/Table.css
blukai: We should use this for everything - https://github.com/odota/ui/blob/master/src/components/App/App.css#L20 like here, not only for table

src/components/Player/Pages/Overview/Overview.css
blukai: Yes we need `!important` since width doesn't changes without it

mike-robertson: so the file could look like this:
```css
@import "../../../palette.css";

.overviewMatches {
  width: calc(70% - 30px);
  margin-right: 30px;
}
.overviewHeroes {
  width: 30%;
}

.overviewMatches, .overviewHeroes {
  display: inline-block;
  @media only screen and (max-width: 1180px) {
    width: 100%;
    margin: 0;
  }
}
```

src/components/Player/Pages/Matches/Matches.css
Roshanjossey: We could select a range of elements using something like [css-to-select-range-of-children](https://gist.github.com/rondevera/167627). Here's a fiddle http://jsfiddle.net/46XRe/

Roshanjossey:
@mike-robertson: I feel like you've caught almost everything.
Great work 👍 
I still don't have enough contest to completely understand the changes. Let's hope I get there sooner.

src/components/Player/Pages/Matches/Matches.css
blukai: I added attributes 🆎 
https://github.com/odota/ui/pull/253/files#diff-61c29e7b5d9b23af8c7174e64aa10935R3
https://github.com/odota/ui/pull/253/files#diff-cfe4802fd02a36ad920ff929a928712fR33

src/components/Player/Pages/Matches/Matches.css
```diff
@@ -0,0 +1,13 @@
+@media only screen and (max-width: 880px) {
+  th[data-mediaq$="_wide"],
+  td[data-mediaq$="_wide"] {
+    display: none;
+  }
```
mike-robertson: I think this is a good start, but I am not a fan of making predefined breakpoints since we are planning on moving the table outside of the code eventually. Instead, I would rather leverage the redux-responsive to inject the width into the Table and TableHeader components. Then, we can check to see if mediaQuery returns true at that width. I'll post an example in the columns.

src/components/Player/Pages/Matches/playerMatchesColumns.jsx
```diff
@@ -23,27 +23,46 @@ export default [{
   field: 'skill',
   sortFn: true,
   displayFn: transformations.skill,
+  mediaQ: 'skill_wide',
 }, {
```
mike-robertson: Here's the example:
```js
mediaQuery: width => width <= 880; 
```
This will give us more flexibility if we want to do stuff more granularly without locking us into these predefined widths. We can also leverage predetermined widths within redux-responsive (or we can define custom ones). Here's the documentation on that. 
https://github.com/AlecAivazis/redux-responsive#using-custom-breakpoints

src/components/Table/Table.jsx
```diff
@@ -44,7 +40,7 @@ const getTable = (data, columns, sortState, sortField, sortClick) => (
                 overflow: `${column.field === 'kills' ? 'visible' : null}`,
```
mike-robertson: @blukai 
at columns.map, do a check for mediaQuery
```js
columns.map((column, index) => {
if (!column.mediaQuery || !column.mediaQuery(width)) {
  return null;
}
```
see my pr on playerheader or the redux-responsive docs for how to bring width into the table. Do the same thing in TableHeader where we map over columns to create the header row.

PR #254, opened 2016-10-09
With sorting

mike-robertson:
Changes table api to store sorting on a per table basis in the redux store. Now the table api looks like this:
```js
<Table paginated sorted data={data} columns={columns} />
```


src/components/Table/SortedTable/withSort.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import uuid from 'node-uuid';
+import { table } from 'reducers';
+import { sortTable } from 'actions';
+// TODO - get sorting to actually work. Nothing happens when I click right now.
```
howardchung: is this still true?

PR #256, opened 2016-10-09
Add headings to split tables

dartharaknus:


src/components/Match/matchPages.jsx
```diff
@@ -45,45 +45,43 @@ const filterMatchPlayers = (players, team = '') =>
...
 
+// TODO: Add networth and xp graph to overview
 const matchPages = [{
   name: strings.tab_overview,
   content: match => (<div>
-    <Heading title={strings.heading_overview} />
-    <MatchPlayersTableSplit match={match} columns={overviewColumns(match)} />
-    <Heading title={strings.heading_ability_builds} />
-    <MatchPlayersTable match={match} columns={abilityUpgradeColumns} />
+    <TeamTable match={match} columns={overviewColumns(match)} radiant={strings.general_radiant} dire={strings.general_dire} />
```
howardchung: You shouldn't need to pass the radiant/dire strings as props if they aren't going to change.

src/components/Match/matchPages.jsx
```diff
@@ -45,45 +45,43 @@ const filterMatchPlayers = (players, team = '') =>
...
+    <TeamTable match={match} columns={abilityUpgradeColumns} radiant={strings.radiant_ability_builds} dire={strings.dire_ability_builds} />
     <Heading title={strings.heading_buildings} />
     <BuildingMap match={match} />
   </div>),
 }, {
   name: strings.tab_benchmarks,
   content: match => (<div>
     <Heading title={strings.heading_benchmarks} />
-    <MatchPlayersTable match={match} columns={benchmarksColumns(match)} />
+    <Table match={match.players} columns={benchmarksColumns(match)} />
```
howardchung: Table does not take a match prop.  You either have to make it pass data, or keep the MatchPlayersTable wrapper component around.

src/lang/en.json
```diff
@@ -49,6 +49,9 @@
   "gold_reasons_14": "Roshan",
   "gold_reasons_15": "Courier",
 
+  "radiant_ability_builds": "Radiant Ability Builds",
```
howardchung: you should probably just build the string with (Radiant|Dire) + existing tooltip.

dartharaknus: That's a good idea.

howardchung: IIRC radiant and dire already strings (in general_ section)

src/components/Match/matchPages.jsx
```diff
@@ -45,45 +45,42 @@ const filterMatchPlayers = (players, team = '') =>
...
-  columns,
-}) => (<Table data={match.players} columns={columns} />);
-
-const MatchPlayersTableSplit = ({
+const TeamTable = ({
   match,
   columns,
+  heading = '',
 }) => (<div>
+  <Heading title={`${strings.general_radiant} ${heading}`.trim()} />
```
howardchung: why do you need to .trim() here?

howardchung:
Code looks OK, but can we get some feedback on whether we want double headings for each section?

src/components/Match/matchPages.jsx
dartharaknus: In case heading is empty which we are doing.

dartharaknus:
These headings would vanish when people want to mix both the teams. I am working on implementing that switch but it will take time.

howardchung:
Consider applying this change to all the tables, so we can see how it looks on all pages?

dartharaknus:
Updated.

src/components/Match/matchPages.jsx
mike-robertson: Could we trim the title in the Heading component? This seems like something we wouldn't want to happen in lots of cases.

howardchung: agreed with mike

src/components/Match/matchPages.jsx
```diff
@@ -45,45 +45,39 @@ const filterMatchPlayers = (players, team = '') =>
...
 
+// TODO: Add networth and xp graph to overview
 const matchPages = [{
   name: strings.tab_overview,
   content: match => (<div>
-    <Heading title={strings.heading_overview} />
-    <MatchPlayersTableSplit match={match} columns={overviewColumns(match)} />
-    <Heading title={strings.heading_ability_builds} />
-    <MatchPlayersTable match={match} columns={abilityUpgradeColumns} />
+    <TeamTable match={match} columns={overviewColumns(match)} />
```
howardchung: I actually prefer the name MatchPlayersTableSplit since it's more descriptive.  This element actually consists of two separate tables, so TeamTable is a little misleading.

src/components/Match/matchPages.jsx
```diff
@@ -99,32 +93,27 @@ const matchPages = [{
     </Col>
     <Col md={12}>
       <Heading title={strings.heading_damage} />
-      <MatchPlayersTable match={match} columns={inflictorsColumns} />
+      <Table data={match.players} columns={inflictorsColumns} />
```
howardchung: This can be split too.

src/components/Match/matchPages.jsx
pksunkara: Even if you trim the heading, adding a space before it later would still give an extra space. Which is why there is a trim here.

src/components/Match/matchPages.jsx
dartharaknus: I like this name because once we implement the option to merge players, this function will produce a single table and not split tables. I wanted to name it `Table` but didn't want to mess with existing component names.

src/components/Match/matchPages.jsx
```diff
@@ -99,32 +93,27 @@ const matchPages = [{
     </Col>
     <Col md={12}>
       <Heading title={strings.heading_damage} />
-      <MatchPlayersTable match={match} columns={inflictorsColumns} />
+      <TeamTable data={match.players} columns={inflictorsColumns} />
```
howardchung: Your table takes match, not data

src/components/Match/matchPages.jsx
```diff
@@ -45,45 +45,39 @@ const filterMatchPlayers = (players, team = '') =>
...
-  columns,
-}) => (<Table data={match.players} columns={columns} />);
-
-const MatchPlayersTableSplit = ({
+const TeamTable = ({
   match,
   columns,
+  heading = '',
 }) => (<div>
+  <Heading title={`${strings.general_radiant} ${heading}`} />
```
howardchung: If you want to avoid the trailing space, you could also use an array with .join()

dartharaknus:
Updated.

PR #263, opened 2016-10-10
Fixed webpack watch in docker

dartharaknus:


howardchung:
Should we make this a non-default option (i.e., user has to specifically start it in this mode?)  Otherwise we're adding the additional cost of polling for non-Docker users.

dartharaknus:
I am not familiar with webpack. I don't even know why it's watching in CLI mode even though we are not passing `watch` anywhere. I will leave this to you.

dartharaknus:
Okay, I have updated this.

scripts/docker.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+npm install
+webpack-dev-server --compress --watch-poll
```
howardchung: This'll work, but it might be cleaner to have the docker start pass an env var, and let webpack.config.js read the env var when deciding whether to add the key to config.

dartharaknus: This PR removes the prestart too which is helpful.

PR #264, opened 2016-10-10
MatchHeader

blukai:
image

Region, skill and buttons will move to another location at the bottom (under overview table)


mike-robertson:
Looks good! I'll try to review sometime today.

howardchung:
@bguzryanto who originally implemented match header

dartharaknus:
Can we add region, skill and matchID to the header too? I don't think they should be separated.

blukai:
I do not consider it necessary to reflect this in the header since nothing prevents scroll down the page a bit. If it will be not so convenient, then we can display it in header as it always was. 🍀

src/components/Match/MatchHeader/MatchHeader.css
```diff
@@ -0,0 +1,64 @@
...
+  font-size: 28px;
+  padding-top: 10px;
+  -webkit-filter: drop-shadow(0 0 15px var(--colorYelor));
+  filter: drop-shadow(0 0 15px var(--colorYelor));
+}
+.score {
+  font-size: 48px;
+}
+
+.radiant div:first-of-type {
```
mike-robertson: same thing for all these. We can use css-modules composes for shared styles.

```css
.team {
  & svg {
    width: 32px;
    height: 32px;
    margin: 0 15px;
    vertical-align: sub;
  }
}

.dire {
  composes: team;
  color: var(--colorDanger);
  & svg {
    fill: var(--colorDanger);
  }
  & div:first-of-type {
    float: left;
  }
}

.radiant {
  composes: team;
  color: var(--colorSuccess);
  & svg {
    fill: var(--colorSuccess);
  }
  & div:first-of-type {
    float: right;
  }
}
```

src/components/Match/MatchHeader/MatchHeader.jsx
```diff
@@ -0,0 +1,72 @@
...
+import strings from 'lang';
+import Spinner from 'components/Spinner';
+import { IconRadiant, IconDire } from 'components/Icons';
+import styles from './MatchHeader.css';
+
+export default ({ match, loading }) => {
+  if (!loading) {
+    const killsRadint = [];
+    const killsDire = [];
+    let player;
```
mike-robertson: You should declare this in the for loop if it is scoped to the for loop

src/components/Match/MatchHeader/MatchHeader.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+    for (let i = 0; i < match.players.length; i += 1) {
```
mike-robertson: Actually, just do this functionally:
```js
const kills = match.players.reduce(
  (kills, player) => {
    if (isRadiant(player.player_slot) {
      kills.radiant.push(player.kills);
    } else {
      kills.dire.push(player.kills);
    }
    return kills;
  },
  {
    radiant: [],
    dire: [],
  }
);
```

src/components/Match/MatchHeader/MatchHeader.css
```diff
@@ -0,0 +1,68 @@
...
+  & svg {
+    width: 32px;
+    height: 32px;
+    margin: 0 15px;
+    vertical-align: sub;
+  }
+}
+
+.radiant {
+  composes: team;
```
mike-robertson: Can we make this responsive? Right now, it sort of breaks after 760px, and it starts to look sort of bad at ~820ish

src/components/Match/Match.jsx
```diff
@@ -48,13 +42,8 @@ class RequestLayer extends React.Component {
...
     return (
       <div>
-        <MatchHeader match={match} user={this.props.user} />
-        <div style={{ marginTop: 25 }}>
-          <TabBar
-            info={info}
-            tabs={matchPages(matchId)}
-          />
-        </div>
+        <MatchHeader match={match} loading={this.props.loading} />
```
howardchung: if we aren't going to pass user here, we can remove the selector?

src/components/Match/MatchHeader/MatchHeader.jsx
```diff
@@ -0,0 +1,72 @@
...
+import { Row, Col } from 'react-flexbox-grid';
+import { transformations, isRadiant } from 'utility';
+import strings from 'lang';
+import Spinner from 'components/Spinner';
+import { IconRadiant, IconDire } from 'components/Icons';
+import styles from './MatchHeader.css';
+
+export default ({ match, loading }) => {
+  if (!loading) {
+    const kills = match.players.reduce(
```
howardchung: I think it's cleaner to just iterate twice here and use .map and .reduce to get your final values.  Sure, you're looping twice, but number of players is not scaling.

src/components/Match/MatchHeader/MatchHeader.jsx
```diff
@@ -0,0 +1,72 @@
...
+        return kills;
+      },
+      {
+        radiant: [],
+        dire: [],
+      }
+    );
+
+    return (
+      <header className={styles.header}>
```
howardchung: Consider adding space to support team names (which is something I want to add at some point).  Perhaps just make some empty elements and test layout by putting in fake team names.

blukai:
Normal view
image

1024px
image

PR #267, opened 2016-10-10
refactor header to have button in the title

mike-robertson:
@blukai feel free to start looking around at the changes. More changes to come while I am working on making the mobile size work.

fullsize:

1100px: 

just before first breakpoint:

### DIFF
--- a/src/components/Player/Header/PlayerHeader.jsx
+++ b/src/components/Player/Header/PlayerHeader.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { CardHeader } from 'material-ui/Card';
 import Avatar from 'material-ui/Avatar';
 import Badge from 'material-ui/Badge';
 import ReactTooltip from 'react-tooltip';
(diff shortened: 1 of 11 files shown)


mike-robertson:
new screenshots
### just after first breakpoint


### just before second breakpoint


### just after second breakpoint


### 440px


### 320px (iphone 5)

howardchung:
Opinion on making the refresh button part of the icon row?  I guess as an "action" it's kind of different from the others.

blukai:
Looks good! 👍  But a few notes about `new screenshots`:
After first and before second breakpoints: left aligned buttons will look better I guess. And after second breakpoint font size of stat title may be smaller or probably it's can be done on 440px.

mike-robertson:
Do you mean keep the coach and refresh buttons on the title @blukai? I can make those changes today

blukai:
No. move it to left side 

image

Keep the coach and refresh buttons on the title also better then centered before second breakpoint, but what if the user have long personaname?

mike-robertson:
Oh ok, I understand bow. Yea, that is better. I'll push that change up

mike-robertson:
@blukai change is done. I didn't update screenshots but feel free to pull the branch down and play around with the responsive sizing.

blukai:
And in addition align vertically avatar and personaname + badges. Like this: 

image

src/components/Player/Header/PlayerHeader.jsx
```diff
@@ -52,43 +46,71 @@ const PlayerName = ({ playerName, playerId, picture, registered, loading, error,
...
+      marginRight: width > HEADER_SM_BREAK ? 30 : 0,
+    };
+
+    if (width <= HEADER_MD_BREAK) {
+      badgeStyle = {
+        ...badgeStyle,
+        marginLeft: -1 * (LARGE_IMAGE_SIZE / 2) * 0.75,
+      };
+    }
+    if (width <= HEADER_SM_BREAK) {
```
howardchung: else if?

src/components/Player/Header/PlayerHeader.jsx
```diff
@@ -52,43 +46,71 @@ const PlayerName = ({ playerName, playerId, picture, registered, loading, error,
...
+            badgeStyle={badgeStyle}
+            style={{
+              margin: 0,
+              padding: 0,
+            }}
+          >
+            <Avatar
+              src={picture}
+              style={avatarStyle}
+              size={width > HEADER_MD_BREAK || width <= HEADER_SM_BREAK ? LARGE_IMAGE_SIZE : LARGE_IMAGE_SIZE / 2}
```
howardchung: Could we control image size via media queries like:
http://stackoverflow.com/questions/22873219/resizing-images-using-media-queries

mike-robertson: I haven't tried it but I assumed the avatar thing is doing more than just image size with that size.

howardchung: We should consider not using the avatar component and just using our own then, so we have greater control.  I think all it does is add some padding and make the image round.

mike-robertson: Ok so no matter how we do this, the current implemention is the cleanest. It doesn't make sense to not pass the size of the avatar in js, but explicitly set breakpoints in the css (from a readability point of view). I did get it to work by using css, but it's messier.

mike-robertson: And by messier I mean you can't tell where avatar size is coming from unless you look at the css. So if you think that's ok @howard I can push this up.

PR #272, opened 2016-10-12
distrib/rankings fixes

### LINKED ISSUE #257: Distributions fixes
- [x] port over data bias warning
- [x] show number of players in solo mmr distribution
- [x] add rank number to each country

howardchung:
fixes #257 

* Use standard Table element
* Use standard spinner element
* Externalize some strings

Could use some more styling on the distributions warning/number of players display.

Reviewers:
@bguzryanto 
@nicholashh 
@albertcui

### DIFF
--- a/src/components/Distributions/Distributions.jsx
+++ b/src/components/Distributions/Distributions.jsx
@@ -48,17 +48,21 @@ const Distributions = ({
           <Heading title={strings[`distributions_${key}`]} />
           {(key === 'mmr') ?
             <div>
-              <div>{`${data
-              && data.mmr
-              && data.mmr.rows
-              && data.mmr.rows.map(row => row.count).reduce(sum)} ${strings.th_players}`}</div>
+              <div>
+                {`${data
+                && data.mmr
+                && data.mmr.rows
+                && data.mmr.rows.map(row => row.count).reduce(sum)} ${strings.th_players}`}
+              </div>
               <div id="mmr" />
             </div> :
               <div>
-                <div>{`${data
-              && data.country_mmr
-              && data.country_mmr.rows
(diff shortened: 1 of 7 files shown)


src/components/Heroes/Ranking.jsx
```diff
@@ -47,7 +39,7 @@ class Ranking extends Component {
     return (
       <div>
         {isLoading || isError || rankings === null ?
-          renderLoading() : renderRanking(hero, rankings)}
+          <Spinner /> : renderRanking(hero, rankings)}
```
mike-robertson: Should we show a spinner if there is an error or rankings === null? Should we instead show some sort of error message or nothing?

src/components/Heroes/Benchmark.jsx
```diff
@@ -48,7 +39,7 @@ class Benchmark extends Component {
     return (
       <div>
         {isLoading || isError || result === null ?
-          renderLoading() : renderBenchmark(hero, result)}
+          <Spinner /> : renderBenchmark(hero, result)}
```
mike-robertson: same comment as below

src/components/Distributions/Distributions.jsx
```diff
@@ -14,31 +13,58 @@ import {
...
+    <Tabs>
+      {Object.keys(data).map(key => (
+        <Tab key={key} label={strings[`distributions_${key}`]}>
+          <Heading title={strings[`distributions_${key}`]} />
+          {(key === 'mmr') ?
+            <div>
+              <div>{`${data
+              && data.mmr
+              && data.mmr.rows
+              && data.mmr.rows.map(row => row.count).reduce(sum)} ${strings.th_players}`}</div>
```
mike-robertson: spacing here is a little confusing

PR #275, opened 2016-10-12
Matches/overview

### LINKED ISSUE #103: skill build layout
The skill build layout currently looks terrible in my opinion.

When none of the heroes reach higher levels, the numbers are all clumped out without any spacing. [Example](http://imgur.com/eqUEXLW)

Secondly, the scroll bar is annoying. The user should be able to see all the levels at a glance. The issue is that the buildings minimap is taking quite a lot of space. It should be moved either above or below this data, like on dotabuff

blukai:
fixes #103


blukai:
For now it looks like this, _(not yet done)_

image

howardchung:
Some early comments based on screenshots:

* I think MMR should go in the subText rather than level (same as we are doing for rankings).
* Level can go back to its own column
* I think Items should be the rightmost column (convention)

src/components/Match/matchColumns.jsx
```diff
@@ -27,7 +27,7 @@ export const heroTd = (row, col, field, hideName) => (
     title={row.account_id ? row.personaname : strings.general_anonymous}
     registered={row.last_login}
     accountId={row.account_id}
-    subtitle={`${row.level} Lvl`}
+    subtitle={`${row.solo_competitive_rank || `${strings.general_unknown.substr(0, 4)}.`} MMR`}
```
howardchung: I'd just use an empty string for no MMR data.

src/components/Match/matchColumns.jsx
```diff
@@ -37,7 +37,7 @@ export const heroTdColumn = {
   displayName: 'Player',
   field: 'hero_id',
   displayFn: heroTd,
-  sortFn: row => (row.level),
+  sortFn: row => row.solo_competitive_rank,
```
howardchung: this sortFn needs to sort on player slot.  This restores the default order.

src/lang/en.json
```diff
@@ -250,7 +250,7 @@
   "th_abilities": "Abilities",
   "th_items": "Items",
   "th_mmr": "MMR",
-  "th_level": "LV",
+  "th_level": "Lvl",
```
howardchung: probably just "LV" is better

src/components/Match/matchColumns.jsx
blukai: Probably we should add the button that can restore the default order for table. If user used sorting, show button on right side of the table heading.

image

src/lang/en.json
howardchung: but pokemon!

http://img.gamefaqs.net/screens/c/c/6/gfs_45550_1_2.jpg

src/components/Match/matchColumns.jsx
howardchung: I see.  In that case I prefer the full text "Unknown"

src/lang/en.json
blukai: Lul

blukai:
image

blukai:
image

blukai:
I'm going offline for a week. @howardchung, @mike-robertson, You can review and push changes right here. 🍀

howardchung:
OK.  Do you want us to wait for you before merge, or just build on this and merge when we're OK with it?

src/components/Match/matchColumns.jsx
```diff
@@ -9,87 +9,87 @@ import abilityKeys from 'dotaconstants/json/ability_keys.json';
...
-  displayFn: (row, col, field, i) => {
+  displayFn: (row) => {
     if (match.parties) {
+      const i = match.players.findIndex(player => player.player_slot === row.player_slot);
       const partyPrev = match.parties[(match.players[i - 1] || {}).player_slot] === match.parties[match.players[i].player_slot];
       const partyNext = match.parties[(match.players[i + 1] || {}).player_slot] === match.parties[match.players[i].player_slot];
       const parentStyle = {
         position: 'relative',
-        width: '100%',
+        width: 0,
```
howardchung: will these work on different resolutions (e.g. mobile?)

src/components/Match/matchColumns.jsx
```diff
@@ -9,87 +9,87 @@ import abilityKeys from 'dotaconstants/json/ability_keys.json';
...
       }
       if (partyPrev && !partyNext) {
-        return (<div style={parentStyle}>
-          <div style={Object.assign({}, style, { borderBottom: borderStyle, height: '50%', top: '0%' })} />
-          <div style={Object.assign({}, style, { borderLeft: borderStyle, height: '50%', top: '0%' })} />
-        </div>);
+        return (
+          <div style={parentStyle}>
+            <div style={{ ...style, borderBottom: borderStyle, height: '50%', top: 0 }} />
+            <div style={{ ...style, borderLeft: borderStyle, height: '56%', top: -1 }} />
```
howardchung: can we have comments explaining why we use numbers like 56 and -1?

src/components/Match/matchColumns.jsx
```diff
@@ -9,87 +9,87 @@ import abilityKeys from 'dotaconstants/json/ability_keys.json';
...
+          </div>
+        );
       }
     }
-    return <div />;
+    return null;
```
howardchung: will this crash React if we return null?

src/components/Match/matchColumns.jsx
```diff
@@ -541,11 +528,7 @@ export const inflictorsColumns = [{
     field: 'damage_inflictor',
     displayFn: (row, col, field) => (field ? Object.keys(field)
       .sort((a, b) => field[b] - field[a])
-      .map((k, i) => inflictorWithValue({
-        inflictor: k,
-        value: abbreviateNumber(field[k]),
-        key: i,
-      })) : ''),
+      .map(k => inflictorWithValue(k, abbreviateNumber(field[k]))) : ''),
```
howardchung: why did we switch to non-destructured inputs?  I think it's easier to read if we can have named parameters in the invocations.

src/utility/index.jsx
```diff
@@ -33,12 +33,15 @@ export function abbreviateNumber(num) {
...
-  return time;
+  if (!isNaN(parseFloat(input)) && isFinite(input)) {
+    const absTime = Math.abs(input);
+    const minutes = Math.floor(absTime / 60);
+    const seconds = pad(Math.floor(absTime % 60), 2);
+    let time = ((input < 0) ? '-' : '');
+    time += `${minutes}:${seconds}`;
+    return time;
+  }
+  return null;
```
howardchung: should we return '' instead since this function returns a string?

src/components/Match/matchColumns.jsx
blukai: yes

src/components/Match/matchColumns.jsx
gu3st: null is "correct" to return if you don't want a component to render anything.

src/components/Match/matchColumns.jsx
blukai: It makes no sense since now we have only 2 inputs https://github.com/odota/ui/pull/275/files#diff-525931b3dcf7bf9cbd9e42512129c82dR230

howardchung: Regardless of the number of inputs, it's helpful to have the parameter name declared on invocation  so the developer doesn't have to look up the function signature.  (depending on your dev environment, your IDE might take care of this for you).

It's not a big deal to me, so if you like it better this way, we can keep it :)

blukai:
dotacoach string will come after merging #292
image

**≥1200px**
image

**≥600px**
image

**≥560px**
image

blukai:
image

src/components/Match/matchColumns.jsx
mike-robertson: I agree with Howard, it's easier to use. My only concern would be creating objects just to pass data if we were dealing with lots of rows could actually be a performance issue potentially.

howardchung:
resolve conflict

PR #276, opened 2016-10-13
filtering

### LINKED ISSUE #74: filter bugfixes
- [x] also is it possible to make the dropdowns appear without having to type anything
- [x] also URL state needs to be updated when a filter is applied
- [x] also text field should clear after a chip is selected
- [x] constrain selections to max 1 element (on most fields)
- [x] update w/l when a filter is applied
- [x] also appears to be a bug where the ? doesn't show up in the URL sometimes

### LINKED ISSUE #234: Enable filters on histograms/trends

### LINKED ISSUE #283: hydrate form state from query string
The form state always starts out empty, even if we have a filter in the query string.  The data components render correctly, but not the form.

### LINKED ISSUE #284: swapping tabs should preserve query string
Currently we just use router push to update the location without saving the query string.  We should save it so swapping tabs preserves any filtering options in the URL.

howardchung:
fixes #234 
fixes #74 
fixes #283 
fixes #284 

* Store filter state in URL and have individual pages getData based on url state.  Also applied to w/l component.
* Remove pages prop from form reducer.  All player pages should share form state.
* ~~Use select field component rather than autocomplete?~~
* ~~Use right-side drawer for filter form?~~ (not yet, maybe later)


src/components/Player/Pages/Wardmap/Wardmap.jsx
```diff
@@ -24,6 +25,7 @@ class RequestLayer extends React.Component {
 
   render() {
     return (<div>
+      <TableFilterForm submitAction={getPlayerWardmap} id={this.props.playerId} page="heroes" />
```
howardchung: remove page prop?

howardchung:
ready for review:
@mike-robertson 
@albertcui 
@nicholashh

src/actions/formActions.js
```diff
@@ -1,3 +1,4 @@
+/* global window */
```
howardchung: not needed anymore

src/components/Player/Pages/Peers/Peers.jsx
```diff
@@ -19,7 +19,7 @@ const Peers = ({ playerId, data }) => (
 );
```
howardchung: remove submitAction on all TableFilterForms

src/components/Player/Player.jsx
```diff
@@ -55,14 +29,28 @@ class RequestLayer extends React.Component {
   }
 
   componentWillUpdate(nextProps) {
-    if (this.props.params.accountId !== nextProps.params.accountId) {
+    if (this.props.params.accountId !== nextProps.params.accountId || this.props.location.key !== nextProps.location.key) {
```
howardchung: for consistency this should also be this.props.accountId.

src/components/Player/Player.jsx
```diff
@@ -55,14 +29,28 @@ class RequestLayer extends React.Component {
...
+    const routeParams = this.props.routeParams;
+    const info = this.props.routeParams.info || 'overview';
+    const page = playerPages(accountId).find(page => page.name.toLowerCase() === info);
+    return (
+      <div>
+        <div className={styles.header}>
+          <PlayerHeader playerId={accountId} location={location} />
+          <TabBar info={info} tabs={playerPages(accountId)} />
+        </div>
+        {!loading && page ? page.content(accountId, routeParams, location) : <Spinner />}
```
howardchung: This won't work since load state is determined per page rather than at player level.  Need to determine spinner state at page level.

howardchung:
would like signoff from at least @mike-robertson before merge.

src/components/Form/FormField.jsx
```diff
@@ -1,83 +1,114 @@
...
-          text: closestMatch.text,
-          value: closestMatch.value,
+          text: value,
+          value,
         },
         limit
       );
-      setFieldText(formName, name, '');
     }
+    this.input.setState({
```
mike-robertson: You might want to put a comment on this (not doing it here causes the thing not to clear, right?)

src/components/Form/FormField.jsx
```diff
@@ -1,83 +1,114 @@
...
+        ref={ref => (this.input = ref)}
+        openOnFocus
+        dataSource={dataSource}
+        dataSourceConfig={dataSourceConfig}
+        floatingLabelText={label}
+        filter={AutoComplete.fuzzyFilter}
+        maxSearchResults={maxSearchResults}
+        onNewRequest={(value, index) =>
+        this.handleRequest({ value, index, formName, name, dataSourceConfig, dataSource, strict, addChip, limit })}
+        // onUpdateInput={searchText => searchText.length === 1 && setFieldText(formName, name, searchText)}
```
mike-robertson: You can probably remove this comment

src/reducers/form.js
```diff
@@ -34,16 +34,9 @@ const field = (state = initialFieldState, action) => {
...
-  show: false,
-};
-
-const page = (state = initialPageState) => ({
-  show: !state.show,
-});
-
+// TODO Hydrate initial form state from query string
 const initialFormState = {
-  pages: {},
```
mike-robertson: You may want to consider keeping this reducer structure (since this form could be used by other kinds of forms). Then you could say that the 'page' is just hardcoded to 'global' for this form (since it shares global state). If we have forms in the future, we might not want the same form to share global state.

mike-robertson:
Let me know your opinion on the reducer comment, then I can approve

src/reducers/form.js
mike-robertson: I think that's a great idea on both accounts. I think this is ready to merge unless you want another person's take on the changes.

src/components/Player/Pages/Trends/Trends.jsx
```diff
@@ -7,34 +7,35 @@ import {
...
 import { getCumulativeDataByField } from 'selectors';
-import { withRouter } from 'react-router';
 import { deSnake } from 'utility';
 import ButtonGarden from 'components/ButtonGarden';
 import trendNames from 'components/Player/Pages/matchDataColumns';
 // import Heading from 'components/Heading';
+import { TableFilterForm } from 'components/Form';
 
 const selectTrend = (router, playerId) => (trendName) => {
   router.push(`/players/${playerId}/trends/${trendName}`);
```
howardchung: preserve query string

src/components/Heatmap/Heatmap.jsx
```diff
@@ -26,16 +26,15 @@ function scaleAndExtrema(points, scalef, max, shift) {
...
-    const adjustedData = scaleAndExtrema(points, width / 127, null, 25);
-    // console.log(adjustedData);
-    heatmap.setData(adjustedData);
-  }
+  // scale points by width/127 units to fit to size of map
+  // offset points by 25 units to increase visibility
+  const adjustedData = scaleAndExtrema(points, width / 127, null, 25);
+  // console.log(adjustedData);
+  heatmap.setData(adjustedData);
+  heatmap.repaint();
```
howardchung: not needed

PR #279, opened 2016-10-14
Started on responsive content

dartharaknus:
# Before vs After

## 1920 width desktop

1920 before
1920 after

## Ipad Landscape (1024 width)

ipad landscape before
ipad landscape after

## Ipad (768 width)

ipad before
ipad after

## Iphone (3xx width)

iphone before
iphone after

### DIFF
--- a/src/components/Player/Pages/Records/Records.jsx
+++ b/src/components/Player/Pages/Records/Records.jsx
@@ -13,7 +13,7 @@ const Records = ({ playerId, data }) => (
   <div>
     <TableFilterForm submitAction={getPlayerRecords} id={playerId} page="records" />
     <TableContainer title={strings.heading_records}>
-      <Table paginated columns={playerRecordsColumns} data={data} />
+      <Table columns={playerRecordsColumns} data={data} />
     </TableContainer>
   </div>
 );


src/components/Table/Table.css
```diff
@@ -3,15 +3,6 @@
...
   clear: both;
-  @media only screen and (max-width: 960px) {
-    overflow-x: scroll;
-  }
-}
-
-.innerContainer {
-  @media only screen and (max-width: 960px) {
-    min-width: 1325px;
-  }
```
dartharaknus: I know this looked like the best way to do responsive content for every page. But this is not. This seriously sucks. Which is why I deleted it.

For an example on how to do responsive better, I have done it for Counts page. I will be following up on other pages in new PRs.

mike-robertson: Do you have a screenshot of what your solution looks like? I don't have my computer on me so I can't see it unless you post a screenshot.

dartharaknus:
@mike-robertson I have added the screenshots.

src/components/Player/Pages/Counts/Counts.jsx
```diff
@@ -13,14 +14,18 @@ import styles from './Counts.css';
...
-        </div>
-      ))}
-    </div>
+    <Row>
+      <div className={styles.countsContainer}>
+        {Object.keys(counts).map((key, index) => (
+          <Col xs={12} sm={12} md={6} lg={4} key={index}>
+            <div className={styles.countTable}>
+              <TableContainer title={deSnake(key)} />
+              <Table paginated columns={playerCountsColumns} data={counts[key].list} />
```
mike-robertson: screenshots look great. Here's what I would ask:
1. This table should be wrapped in TableContainer. I don't know why I didn't do it like that in the first place, but it should be like this:
```js
<TableContainer title={deSnake(key)}>
  <Table paginated columns={playerCountsColumns} data={counts[key].list} />
</TableContainer>
```

2. put the row and column into TableContainer
3. Pass xs sm md and lg props into TableContainer. Make default values 12 (so we don't have to pass them in to get max width).

mike-robertson: Pending these changes I will pass this off as it seems to be pretty straight forward and easy and won't require a large amount of effort if we abstract the row/columns away to table container.

mike-robertson: I just tried this method on the matches table and it won't work though. The table breaks at around 780px width. We still need to allow scrolling on tables if they get broken or have some other solution for tables with more than 3 small columns.

dartharaknus:
> put the row and column into TableContainer

That won't work. You need the Row outside of the container. Which means, if you do the Col in the container. They get messed up.

> Pass xs sm md and lg props into TableContainer. Make default values 12 (so we don't have to pass them in to get max width).

It's not working. I think we can always do this later once we add these props to all the pages.

> We still need to allow scrolling on tables if they get broken or have some other solution for tables with more than 3 small columns.

Yes, it needs to be done on table-by-table basis and is my next step. The common scroll behavior fucks up the width even though it has scroll. So, it wasn't exactly right beforehand.

mike-robertson:
Actually, I was originally correct. The change is super easy.

Go to Table.css (keep the changes you made though):

```css

/*Override material-ui style*/
.innerContainer > div > div {
  overflow-y: hidden !important;
  overflow-x: auto !important;
}
```

dartharaknus:
Nope. You still the need the grid. The overflow is only about scrolling/

dartharaknus:
The grid is about the changes in this file.https://github.com/odota/ui/blob/995b00970939f1027aeab172af542d2806d25e15/src/components/Player/Pages/Counts/Counts.css

mike-robertson:
You don't need the grid. We just need to be smart with our css. Check out my PR for how that's possible. You may also want read up on [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), it's very powerful and doesn't require tons of media queries or grid libraries to be responsive.

PR #282, opened 2016-10-16
fixes to table scroll. Fixes to counts page in pure css.

mike-robertson:
iphone 5 (320) Table is scrollable at this size:

second 'breakpoint':

full screensize:


mike-robertson:
@dartharaknus we can achieve the same effect without even needing any media queries or extra code using flexbox properly.

@howardchung

howardchung:
Is this an alternative to #279 ?   LGTM

mike-robertson:
Yes, this would be an alternative

dartharaknus:
I confirmed. This looks good. merge away.

PR #294, opened 2016-10-20
XP graph label in format: XP - Level L

vamo89:
Much better than the PR in the old ui =)

Extracted getLevelFromXp to utility, and used it to create a new valueFormat for xp's graph.

### DIFF
--- a/src/components/Match/MatchGraph.jsx
+++ b/src/components/Match/MatchGraph.jsx
@@ -5,6 +5,7 @@ import uuid from 'node-uuid';
 import c3 from 'c3';
 import {
   formatSeconds,
+  getLevelFromXp,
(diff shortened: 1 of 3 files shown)


src/components/Match/MatchGraph.jsx
```diff
@@ -31,6 +32,7 @@ const drawGraphs = (props, id) => {
       pattern: colorArray,
     };
     const type = props.type === 'difference' ? 'area-spline' : 'spline';
+    const valueFormat = props.type === 'xp' ? xp => `${xp} - Level ${getLevelFromXp(xp)}` : undefined;
```
howardchung: use `strings.th_level`

PR #296, opened 2016-10-20
basic status page

### LINKED ISSUE #201: Status page
https://api.opendota.com/api/status

howardchung:
fixes #201 

Just dumping the data for now.

### DIFF
--- a/src/components/Heroes/index.js
+++ b/src/components/Heroes/index.js
@@ -1,9 +1,3 @@
 import Heroes from './Heroes';
-import Benchmark from './Benchmark';
-import Ranking from './Ranking';
 
-export {
-  Heroes,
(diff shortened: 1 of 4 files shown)


howardchung:
I don't envision its state being needed anywhere else and it will be non-interactive.

Perhaps we can keep it this way for MVP and move it later if we feel the need to?  For components that are fully standalone, I like keeping it modular.

blukai:
we can display some numbers on home page `user_players,
tracked_players,
matches_last_day,
parsed_matches_last_day`

howardchung:
We could, but that would come with some additional cost (having to make a /api/status call).  If we wanted that data, we should probably instead return it in the /api/metadata call that we are already making.

PR #297, opened 2016-10-20
Use browserHistory.push instead of router

### LINKED ISSUE #300: fix trends breaking when switching fields

### LINKED ISSUE #302: replace withRouter with window.history.pushState utility function

howardchung:
fixes #302
fixes #300


src/actions/formActions.js
```diff
@@ -10,24 +12,36 @@ export const formActions = {
   TOGGLE_SHOW_FORM,
 };
 
-export const toggleShowForm = formName => ({
-  type: TOGGLE_SHOW_FORM,
-  formName,
-});
+export const addChip = (router, name, input, limit) => {
```
mike-robertson: Not a fan of tightly coupling all these components to react router if we can avoid it. Keep in mind they have completely rewritten their api 2 times already. I also don't think these would belong in actions since these are just functions, not actions.

howardchung: It doesn't have to use react-router.  We can achieve the same effect directly with window.history.pushState.

mike-robertson: Mostly I mean for this component, a form shouldn't always just modify query string or something. This form might want to do that, but future forms may not want to. If we could find a way to localize this behavior in the tablefilterform component instead of the form component, I think it would be more reusable

PR #301, opened 2016-10-20
fix trends router bug

### LINKED ISSUE #300: fix trends breaking when switching fields

howardchung:
fixes #300 

RCA:
Trends had no router component (withRouter).

Fix:
* Move routing mutation logic to ButtonGarden

### DIFF
--- a/src/components/ButtonGarden/index.jsx
+++ b/src/components/ButtonGarden/index.jsx
@@ -2,13 +2,19 @@ import React from 'react';
 import FlatButton from 'material-ui/FlatButton';
 import { withRouter } from 'react-router';
 import { deSnake } from 'utility';
+import querystring from 'querystring';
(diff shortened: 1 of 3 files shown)


howardchung:
It was already using withRouter, so I thought it was OK.

We CAN make the parent component pass the router, but that's doubling the number of withRouter calls.

mike-robertson:
I think what I mean here is why would a button component only manipulate urls on button clicks. I'd say this component shouldn't be aware of what clicking does as much as possible. So maybe add a component in top of this called routingButtonGarden which does this router change inside. Idk, just seems weird that a button knows so much about what clicking it does.

mike-robertson:
Tldr: this feels like a presentational component, not such a self aware component.

howardchung:
Integrated into #297

Issue #288, opened 2016-10-17
unparsed match indicator


howardchung:
Currently all tabs are accessible, although they render nothing if there is no parsed data.

howardchung:
An alternative: Just display some text along the lines of "The replay for this match has not yet been parsed.  Not all data may be available."

PR #305, opened 2016-10-22
Issue #288

blukai:


blukai:
image

howardchung:
Actually this issue refers to match page indication.  However, if you think this is better that what we have now we can still merge this :)

blukai:
😆 I know, it's not yet done

howardchung:
OK, just making sure since you tagged the issue in the commit rather than the PR :)

blukai:
**Unparsed match**
image

src/components/Match/MatchHeader/MatchHeader.jsx
```diff
@@ -2,86 +2,89 @@ import React from 'react';
...
+            </li>
+            <li>
+              <span>{strings.match_avg_mmr}</span>
+              {(mmrPlayers.reduce(sum) / mmrPlayers.length).toFixed(0)}
+            </li>
+          </ul>
+        </Col>
+      </Row>
+      {!match.version &&
+      <div className={styles.unparsed}>
```
howardchung: could we isolate this in a component?  We might want to reuse it for the Distributions warning.

src/components/TabBar/TabBar.jsx
```diff
@@ -7,23 +7,36 @@ const onActive = (tab) => {
...
-          <Tab
-            key={index}
-            className={styles.tab}
-            value={tab.name.toLowerCase()}
-            label={tab.name}
-            onActive={() => onActive(tab, router)}
-          />
-        ))}
+        {tabs.map((tab, index) => {
+          let isDisabled = false;
```
howardchung: I think instead of doing this here, it should be determined in the Match component when we generate the TabBar array.

blukai: Yes, I tried to do it at beginning, but have no idea how 😞

PR #318, opened 2016-10-23
add loading to players

mike-robertson:
Added a ContentContainer component. Added this component into the TableContainer component, so you can pass error and loading down to that and therefore don't have to load in Heading and ContentContainer components.


src/actions/player/playerWardmapActions.js
```diff
@@ -31,7 +31,7 @@ export const getPlayerWardmapError = (payload, id) => ({
 
 export const getPlayerWardmap = (playerId, options = {}) => (dispatch, getState) => {
   if (playerWardmap.isLoaded(getState(), playerId)) {
-    dispatch(getPlayerWardmapOk(playerWardmap.getHeroList(getState(), playerId), playerId));
+    dispatch(getPlayerWardmapOk(playerWardmap.getPlayerWardmap(getState(), playerId), playerId));
```
howardchung: why is wordcloud "byid" and this one not?

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -7,17 +7,20 @@ import { HistogramGraph } from 'components/Visualizations';
...
+const Histogram = ({ routeParams, columns, playerId, error, loading }) => (
   <div style={{ fontSize: 10 }}>
     <TableFilterForm />
     <ButtonGarden
       onClick={buttonName => browserHistory.push(`/players/${playerId}/histograms/${buttonName}${window.location.search}`)}
       buttonNames={histogramNames}
       selectedButton={routeParams.subInfo || histogramNames[0]}
     />
-    <HistogramGraph columns={columns || []} />
+    <ContentContainer error={error} loading={loading}>
```
howardchung: Could we just use ContentContainer for tables instead of a separate TableContainer?

src/components/Player/Pages/MMR/MMR.jsx
```diff
@@ -4,12 +4,13 @@ import { MMRGraph } from 'components/Visualizations';
...
 import strings from 'lang';
-import Heading from 'components/Heading';
+import { TableContainer } from 'components/Table';
 
-const MMR = ({ columns }) => (
+const MMR = ({ columns, error, loading }) => (
   <div style={{ fontSize: 10 }}>
-    <Heading title={strings.heading_mmr} />
-    <MMRGraph columns={columns} />
+    <TableContainer title={strings.heading_mmr} error={error} loading={loading}>
```
howardchung: Why TableContainer here for a non-table element?

src/components/Player/Pages/Wardmap/Wardmap.jsx
```diff
@@ -24,20 +24,24 @@ class RequestLayer extends React.Component {
...
-        </Col>
-      </Row>
-    </div>);
+    const { error, loading, data } = this.props;
+    return (
+      <div>
+        <TableFilterForm />
+        <Row>
+          <Col md>
+            <TableContainer title={strings.th_ward_observer} error={error} loading={loading}>
```
howardchung: why TableContainer for non-table?

src/components/Player/Pages/Wordcloud/Wordcloud.jsx
```diff
@@ -30,14 +30,18 @@ class RequestLayer extends React.Component {
...
-      <Heading title={strings.heading_wordcloud_said} />
-      <Wordcloud counts={this.props.data.my_word_counts} />
-      <Heading title={strings.heading_wordcloud_read} />
-      <Wordcloud counts={this.props.data.all_word_counts} />
-    </div>);
+    const { error, loading, data } = this.props;
+    return (
+      <div>
+        <TableFilterForm />
+        <TableContainer title={strings.heading_wordcloud_said} error={error} loading={loading}>
```
howardchung: same TableContainer comment

src/actions/player/playerWardmapActions.js
mike-robertson: idk, you're the one who made these 2 pages :)

src/components/Player/Pages/Histograms/Histograms.jsx
mike-robertson: Not unless we want no headings on the tables

src/components/Player/Pages/Wardmap/Wardmap.jsx
mike-robertson: Tablecontaine ris a bad name

mike-robertson:
@howardchung refactored container component everywhere

src/components/Container/index.jsx
```diff
@@ -0,0 +1,43 @@
+import React, { PropTypes } from 'react';
+import Heading from 'components/Heading';
+import Spinner from 'components/Spinner';
+import Error from 'components/Error';
+
+export const AsyncContainer = ({ loading, error, children }) => {
```
howardchung: What's the purpose of the async container?

src/store/index.js
```diff
@@ -24,12 +24,14 @@ const reducer = combineReducers({
   routing,
   browser: responsiveStateReducer,
 });
+/* eslint-disable no-underscore-dangle */
```
howardchung: Could we just disable globally?  I'd prefer not to have eslint disables sprinkled around especially if they're permanent

mike-robertson: I agree in principle, but in this case, it's probably something we should avoid: however, it's a 'hack' so that we can make redux dev tools work (it's the only way to hook it into the store basically).

src/components/Container/index.jsx
mike-robertson: It's just a way of breaking down into multiple components. No one has to use this even, it's just more readable.

mike-robertson: but the async container is used by the Container (which is what we would use everywhere else).

src/store/index.js
howardchung: According to eslint docs this rule is only a stylistic choice.  We can just apply it project wide (I don't think we have a use case for underscores anyway)

src/components/Container/index.jsx
mike-robertson: It's an internal component; I don't think the name is super important. Async makes more sense anyways since it has to do with loading and errors.

PR #319, opened 2016-10-23
Issue 85

### LINKED ISSUE #85: item/ability tooltips
Apparently material-ui has them implemented, but not documented:
https://github.com/callemall/material-ui/issues/2230#issuecomment-186824491

I don't know if they support direct HTML content, which we need since the data we have has raw HTML in it.

- [x] items
- [x] abilities
- [x] table headers

blukai:
- [x] isolate `inflictorWithValue` in a component

fixes #85


blukai:
Item description very long for some items. I don't think it's should be in tooltip. May be I should add cd and manacost idk?

image

howardchung:
Yes, see the current UI code for what gets added `public/js/tooltips.js`

blukai:
image

howardchung:
Note that this currently utilizes bundled abilities/items and will increase the bundle size.

PR #320, opened 2016-10-23
Mobile sliding menu for header

recurza:
To fix:

- [x] Set drawer background color to primary color
- [x] Login button on laptop-size screen has mysteriously become invisible


src/components/App/AppLogo.jsx
```diff
@@ -3,10 +3,10 @@ import { Link } from 'react-router';
 import { APP_NAME } from 'config';
 import styles from './App.css';
 
-export default () => (
+export default ({size}) => (
   <Link to="/" className={styles.AppLogo}>
-    <big>
+    <p style={{fontSize:size || '18px'}}>
```
howardchung: should probably use default param here

src/components/Header/Header.jsx
```diff
@@ -30,32 +34,63 @@ const navbarPages = [{
...
-          <div key={page.name} className={styles.tabContainer}>
-            {page.external ?
-              <a href={page.path} className={styles.tab}>{page.name}</a> :
-                <Link to={page.path} className={styles.tab}>{page.name}</Link>}
+function browserSelector({browser}) {
+  return {browser};
+}
+
+const Header = ({location, width}) => {
+  const isTablet = width <= HEADER_MD_BREAK;
```
howardchung: can we achieve this with CSS media queries?

recurza: true, then I'll also use media queries for AppLogo size

src/components/BurgerMenu/BurgerMenu.jsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+
+  render() {
+    return (
+      <div>
+        <IconButton onTouchTap={this.handleToggle.bind(this)} style={{ left:0, margin:0 }}>
+          <MenuIcon />
+        </IconButton>
+        <Drawer
```
blukai: >Set drawer background color to primary color

u can set `className` and use
```css
.drawer {
  & > div {
    &:last-of-type {
      background-color: *COLOR* !important;
    }
  }
}
```

src/components/App/App.jsx
```diff
@@ -13,6 +13,7 @@ import styles from './App.css';
 const muiTheme = {
   fontFamily: palette.fontFamily,
   card: { fontWeight: palette.fontWeightNormal },
+  drawer: { color: palette.defaultPrimaryColor, backgroundColor: palette.defaultPrimaryColor},
```
blukai: remove since this doesn't work

src/components/Header/Header.jsx
```diff
@@ -14,6 +14,10 @@ import AccountWidget from '../AccountWidget';
 import styles from './Header.css';
 import SearchForm from '../Search/SearchForm';
 import AppLogo from '../App/AppLogo';
+import BurgerMenu from '../BurgerMenu';
+
+import { HEADER_MD_BREAK } from '../Player/Header/PlayerHeader';
+import { HEADER_SM_BREAK } from '../Player/Header/PlayerHeader';
```
blukai: U can use `comma` for import from same file

src/components/BurgerMenu/BurgerMenu.jsx
```diff
@@ -0,0 +1,48 @@
...
+          docked={false}
+          width={200}
+          open={this.state.open}
+          onRequestChange={(open) => this.setState({open})}
+        >
+          <MenuItem style={{ backgroundColor: '#1976d2' }}>{this.props.top}</MenuItem>
```
blukai: we have exported `blue` in `palette.css`, u can just use `styles.blue`

src/components/Header/Header.jsx
mike-robertson: Just keep in mind @recurza , those breakpoints were specific to how the header layout broke. I don't know if it's necessary to follow the exact same rules for the navbar. If they are the same, then I would suggest moving these out of the PlayerHeader and into some utility file or something. Also, you can do absolute paths:
```js
import { myThing } from 'components/Player/Header/PlayerHeader';
```

package.json
```diff
@@ -35,6 +35,7 @@
     "react": "15.3.2",
     "react-dom": "15.3.2",
     "react-flexbox-grid": "0.10.2",
+    "react-hamburger-menu": "0.0.4",
```
howardchung: Do we need this?

src/components/Header/Header.jsx
```diff
@@ -30,32 +31,59 @@ const navbarPages = [{
...
+
+  const SearchGroup = () => (
+    <ToolbarGroup style={{ marginLeft: 20 }} className={styles.verticalAlign}>
+      <ActionSearch style={{ marginRight: 6, opacity: '.6' }} />
+      <SearchForm location={location} />
+    </ToolbarGroup>
+  );
 
-export default connect(null, mapDispatchToProps)(Header);
+  const AccountGroup = () => (
```
mike-robertson: I'm not sure how to feel about declaring all these components within the scope of another component. Why not just make them full on components outside of the scope of this component, especially if they aren't relying on closure scope?

src/components/Header/Header.jsx
```diff
@@ -30,32 +31,59 @@ const navbarPages = [{
...
+        <LogoGroup />
+        {!isTablet && (<LinkGroup />)}
+        <SearchGroup />
+        {!isTablet && (<AccountGroup />)}
+      </Toolbar>
+    </div>
+  );
+};
+
+const mapDispatchToProps = () => ({});
```
mike-robertson: you can delete this, it's unnecessary

src/components/BurgerMenu/BurgerMenu.css
```diff
@@ -0,0 +1,21 @@
+@import '../palette.css';
+
+.drawer {
+  & > div {
+    &:last-of-type {
```
mike-robertson: this could be 
```css
.drawer > div:last-of-type {
  rule;
}
```
if you're only going to override material ui style. Also, please notate that you are overriding material ui style here for posterity.

src/components/Header/Header.jsx
howardchung: I actually don't mind this as it's useful as a reminder that this is something that can be passed, and it's less code to change if we ever add mapDispatchToProps to this component.  Fine with either, though :)

src/components/Header/Header.jsx
recurza: They are all wrapped in <ToolbarGroup>, so they likely won't be used anywhere else

mike-robertson: So as discussed in the discord it's better to just move these out of the stateless component function. We want to minimize the things we do inside the render call; of course there can and should be logic, but declaring functions inside the render call means that each rerender is unnecessarily redefining functions. It may be a small/negligible impact in one component, but if all our components are declaring and redeclaring functions that can be declared once instead, it could be taxing to performance.

recurza:
@blukai I can't figure out how to align the name to the left without modifying LoggedIn itself, any tips?

blukai:
Nope. 😆 Looks good.

howardchung:
resolve conflict @recurza

recurza:
resolved

PR #327, opened 2016-10-24
Benchmarks

### LINKED ISSUE #321: prettify match benchmarks
Colorize the percentage according to the same scheme used for rankings

howardchung:
fixes #321 

Reviewers:
@blukai 
@albertcui 
@nicholashh


src/components/Match/matchColumns.jsx
```diff
@@ -220,10 +221,15 @@ export const benchmarksColumns = (match) => {
...
-              <span>{`${Number(bm.pct * 100).toFixed(2)}%`}</span>
-              <span>{bm.raw.toFixed(2)}</span>
+            const bucket = percentile(bm.pct);
+            const percent = Number(bm.pct * 100).toFixed(2);
+            const value = Number(bm.raw.toFixed(2));
+            return (<div data-tip data-for={`benchmarks_${row.player_slot}_${key}`}>
+              <span style={{ color: styles[bucket.color] }}>{`${percent}%`}</span>
+              <small style={{ margin: '3px' }}>{value}</small>
+              <ReactTooltip id={`benchmarks_${row.player_slot}_${key}`} place="top" effect="solid">
+                {`${value} ${strings[`th_${key}`]} ${strings.benchmarks_higher_than} ${percent}% ${strings.benchmarks_recent_performances}`}
```
blukai: What if that value equals 0? As I know there can be no negative values. So how 0 can be higher than 0?

howardchung: It should say "equal or higher than".  I'll fix it.

blukai: Probably `value > 0 ? strings.benchmarks_higher_than : strings.benchmarks_equals` ❓

howardchung: No, it should always be equal or higher than (that's how Redis does the comparisons on the server side).

Issue #335, opened 2016-10-25
image load error for tidebringer


howardchung:
On this match:
http://beta.opendota.com/matches/2729043042/combat

Tidebringer shows a failed image load for what I think is the cooldown?  Doesn't seem to affect other abilities, so perhaps this is an issue with the data source.

PR #337, opened 2016-10-25
Fixes #334, #335

### LINKED ISSUE #334: tooltip for auto attack/other
Currently no tooltip appears, we should display one that just says "Auto Attack/Other"

blukai:


src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -39,7 +40,9 @@ const tooltipContainer = thing => (
       <div
         dangerouslySetInnerHTML={{
           __html: thing.cmb
-            .replace('/apps/dota2/images/tooltips/mana.png', `${API_HOST}/apps/dota2/images/tooltips/mana.png`)
+            .replace('src="/apps/dota2/images/tooltips/mana.png"', `src="${API_HOST}/apps/dota2/images/tooltips/mana.png"`)
```
howardchung: should we do this sort of transform in the `dotaconstants` code?

https://github.com/odota/dotaconstants/blob/master/tasks/updateconstants.js#L46

howardchung: Hmm, I guess we shouldn't put API_HOST in dotaconstants, but perhaps we should be stripping out cdn.dota2.com?

howardchung: On another look, shouldn't we already be handling this?

https://github.com/odota/dotaconstants/blob/master/tasks/updateconstants.js#L59

blukai: Idk, actually it's easy to replace. But if it makes a problem, I guess yes.

howardchung:
I needed to use a regex to do the replacement, otherwise it only does the first instance in each string :)

https://github.com/odota/dotaconstants/commit/2e4b8d9f2183cbaf44485931e5c6ef422d8ae28e#diff-66650d004bd1a855b2fc1d95e77d448dR61

published a new version of dotaconstants 3.3.0

howardchung:
Actually, IMO we should keep dotaconstants as it is directly from the data source (make no assumptions about how the user will use it).

Users who want to replace cdn.dota2.com with their own host (in this case, us since I think using a http source for some page content causes the https lock to become invalid), can do so in their own code.

Sorry about the flip-flop.  Could we invert the change so we do the cdn.dota2.com replacement?  Use the same logic in the above link for global replacement of cdn.dota2.com

src/components/Visualizations/inflictorWithValue.jsx
howardchung: This should now be a single line that does:
`.replace(/http:\/\/cdn\.dota2\.com/g, '');`

src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -40,10 +40,9 @@ const tooltipContainer = thing => (
       <div
         dangerouslySetInnerHTML={{
           __html: thing.cmb
-            .replace('src="/apps/dota2/images/tooltips/mana.png"', `src="${API_HOST}/apps/dota2/images/tooltips/mana.png"`)
-            .replace('src="/apps/dota2/images/tooltips/cooldown.png"', `src="${API_HOST}/apps/dota2/images/tooltips/cooldown.png"`)
-            .replace('http://cdn.dota2.com/apps/dota2/images/tooltips/mana.png', `${API_HOST}/apps/dota2/images/tooltips/mana.png`)
-            .replace('http://cdn.dota2.com/apps/dota2/images/tooltips/cooldown.png', `${API_HOST}/apps/dota2/images/tooltips/cooldown.png`),
+            .replace(/http:\/\/cdn\.dota2\.com/g, '')
```
howardchung: see my updated comment (second arg should be API_HOST).

You also don't need the next two lines, since cdn.dota2.com is now left intact.

PR #346, opened 2016-10-26
Prettify match/cosmetics, globalize tooltip style

blukai:


blukai:
image

blukai:
In #336 I fix spinner's err but in my next pr bring it back. My bad, fix it again. 🍀

src/components/Match/matchColumns.jsx
```diff
@@ -486,16 +493,57 @@ export const runesColumns = [heroTdColumn]
...
+  legendary: '#D32CE6',
+  immortal: '#E4AE33',
+  arcana: '#ADE55C',
+  ancient: '#EB4B4B',
+};
+export const cosmeticsColumns = [{
+  displayName: 'Player',
+  field: 'player_slot',
+  cosmetics: true,
+  displayFn: (row, col, field, i) => heroTd(row, col, field, i, false, 'side'),
```
howardchung: why side instead of MMR?

src/components/Match/matchPages.jsx
```diff
@@ -232,9 +232,12 @@ const matchPages = [{
 }, {
   name: strings.tab_cosmetics,
   parsed: true,
-  content: match => (<div>
-    <TeamTable match={match} columns={cosmeticsColumns} heading={strings.heading_cosmetics} />
-  </div>),
+  content: match => (
+    <div>
+      <Heading title={strings.heading_cosmetics} />
```
howardchung: shouldn't we be consistent and use split tables everywhere if we are going to use them at all?  Seems strange to break from that just for cosmetics.

src/components/Match/matchColumns.jsx
```diff
@@ -486,16 +493,57 @@ export const runesColumns = [heroTdColumn]
...
+  sortFn: true,
+}, {
   displayName: strings.th_cosmetics,
   field: 'cosmetics',
+  cosmetics: true,
```
blukai: To hide rows in cosmetics table for players that have no cosmetics.

src/components/Match/matchColumns.jsx
blukai: Nvm, I have another better idea, I'll revert this change

src/components/Match/matchPages.jsx
blukai: Actually cosmetics IMO shouldn't be splitted for teams as other tables. It's not a team info, its personal.

src/components/Table/Table.jsx
```diff
@@ -37,6 +37,9 @@ const getTable = (data, columns, sortState, sortField, sortClick) => (
           <MaterialTableRow
             key={index}
             className={styles.row}
+            style={{
```
howardchung: I think it's fine to keep empty rows.  Alternatively run the input data through a filter function?

blukai:
@howardchung I guess we should flow cosmetics images through https like for other dota images.

howardchung:
We could, although I don't know if it's needed right away.  Open an issue?

PR #356, opened 2016-10-28
Fix confirmed name

blukai:


src/components/Visualizations/Table/HeroImage.jsx
```diff
@@ -60,7 +60,7 @@ const TableHeroImage = ({
         <div data-tip data-for={`confirmed_${accountId}`} className={styles.confirmed}>
           <IconTrophy />
           <ReactTooltip id={`confirmed_${accountId}`} place="top"effect="solid">
-            {strings.app_confirmed_as} {title}
+            {strings.app_confirmed_as} {confirmed}
```
blukai: I did it only for captains mode. Probably should add captains draft and reverse captains mode. https://github.com/odota/ui/blob/master/src/components/Match/matchColumns.jsx#L32

blukai: ok

PR #367, opened 2016-10-30
Add VP banner

albertcui:


src/components/Header/Header.jsx
```diff
@@ -47,7 +48,7 @@ const LinkGroup = () => (
     {navbarPages.map(page => (
       <div key={page.name} className={styles.tabContainer}>
         {page.external ?
-          <a href={page.path} className={styles.tab}>{page.name}</a> :
+          <a href={page.path} className={styles.tab} target="_blank">{page.name}</a> :
```
howardchung: you need to do this in BurgerMenu component as well (or refactor it out somewhere since the code is duplicated)

src/components/Home/Sponsors.jsx
```diff
@@ -32,7 +29,7 @@ export default () => (
                 {strings.home_become_sponsor}
               </span>
             }
-            href={'/'/* TODO @albertcui needs to set up donation app*/}
+            href={'//carry.opendota.com'}
```
howardchung: can use double quotes directly here

PR #376, opened 2016-10-31
Issue 231

blukai:
image
image


src/components/Match/matchColumns.jsx
```diff
@@ -494,7 +494,22 @@ export const actionsColumns = [heroTdColumn, {
 
 export const runesColumns = [heroTdColumn]
   .concat(Object.keys(runes).map(runeType => ({
-    displayName: strings[`rune_${runeType}`],
+    displayName: (
```
blukai: Nope, I don't think so.

PR #378, opened 2016-11-01
Match, table improvements

blukai:


blukai:
image

src/components/Match/matchColumns.jsx
```diff
@@ -46,168 +58,148 @@ export const heroTdColumn = {
   sortFn: true,
 };
 
+const parties = (row, match) => {
+  if (match.parties && Object.values(match.parties).reduce(sum) > 0) {
```
howardchung: what's the purpose of the second check?

blukai: avoid situations with x10 party indicator
```
parties: {
0: 0,
1: 0,
2: 0,
3: 0,
4: 0,
128: 0,
129: 0,
130: 0,
131: 0,
132: 0
}
```

howardchung: You might want to add initial value.  I think it's always safer to have it unless we have a reason not to.  (I think this crashes if match.parties is `{}`)

PR #384, opened 2016-11-02
A few misc changes

dartharaknus:
1. Records page is no longer paginated. (We don't need it to be)
2. Use blue color for parsed successful status. So, that it's not confused with winning at first glance

    > Even though I know it's a parse status, I kept thinking that it was
some other indication (like winning) since it's green in color

3. In a match page, if the dire wins, move the dire victory header to the right and match meta info to the left. (Like how it is in dota2 game)


dartharaknus:
Would love some feedback from @blukai

dartharaknus:
@blukai Updated.

Issue #387, opened 2016-11-02
react-tooltip perf issue


howardchung:
I disabled the following three tooltips:
* Parse status
* FromNowTooltip
* KLA

These are heavy hitters because they load multiple times per row (on player pages).

We can investigate more performant solutions.

PR #389, opened 2016-11-02
Fix some issues

### LINKED ISSUE #377: Display additional units
ex: https://beta.opendota.com/matches/2743155083

blukai:
Fixes #377

@howardchung, I tried to find out for issue #387, check 2nd commit.


src/components/Match/matchColumns.jsx
```diff
@@ -160,6 +160,7 @@ export const overviewColumns = match => [{
   field: 'items',
   displayFn: (row) => {
     const itemArray = [];
+    const itemBearArray = [];
```
howardchung: It may not be bear only, even that's currently the case.

howardchung:
We'll have to apply this new tooltip code to a few tooltips and perf test it.

src/components/Match/matchColumns.jsx
```diff
@@ -169,8 +170,24 @@ export const overviewColumns = match => [{
           inflictorWithValue(itemKey, formatSeconds(firstPurchase))
         );
       }
+
+      if (row.hero_id === 80 && row.additional_units) {
```
blukai: http://dev.dota2.com/showthread.php?t=132401

howardchung: Gg valve

howardchung: But put a comment explaining this for posterity

blukai:
image

Currently applied for parsed status on players/overview and confirmed, registered, solo mmr on matches

howardchung:
Can we do it for from now tooltip and KLA?

blukai:
Sure, we can do it for all tooltips that doesn't require html 🍀

blukai:
If this tooltip will work fine after some testing, we should replace with it table th tooltips and others.

howardchung:
Will you do FromNowTooltip in this PR?

blukai:
Oh, yes, I forget about it, will do it now.

howardchung:
There seems to be a short delay before the tooltip shows up.  Is there some property we can override?

blukai:
https://github.com/odota/ui/pull/389/files#diff-bab05dbdb7e778b54189549d47041d33R43

I did it because it can be annoying when user hover by chance over icons and it just appear tooltips. If he want to see it, small delay will not affect bad.

howardchung:
I did some quick testing with pages of 20 matches and it looks ok

blukai:
@howardchung, If everything ok with current changes, u can merge this. Can't use new tooltip for table th since table container have `overflow: hidden`, maybe find out a bit letter with th.

PR #396, opened 2016-11-03
implement own tabbar, proptypes fixes

### LINKED ISSUE #364: short TabBars still get full 1920px width
The effect of this is that on short bars such as Distributions, no text is visible to the user without scrolling right blindly.

### LINKED ISSUE #385: TabBar elements aren't right-clickable
This breaks the ability to open in new tab.  We should make these <Link> elements.

howardchung:
fixes #385 
fixes #364 

Reviewers:
@blukai 

@blukai maybe you could help me style this better?


blukai:
Sure. 🦄 Can I push commits right here?

howardchung:
Yeah go ahead

blukai:
image

src/components/Match/Match.jsx
```diff
@@ -38,6 +39,7 @@ class RequestLayer extends React.Component {
           <TabBar
             info={info}
             tabs={matchPages(matchId, match)}
+            mediaQClass={styles.tabBar}
```
howardchung: why does the Match TabBar need this while the others don't?

blukai: Match tabBar have a lot of tabs, the others don't.

PR #398, opened 2016-11-03
Max-width and production match bugfix

dartharaknus:


dartharaknus:
After this is done, we need to fix pagination and safari issues, and I think we can move forward with release.

src/components/Wordcloud/Wordcloud.jsx
```diff
@@ -88,7 +88,7 @@ Wordcloud.defaultProps = {
 };
 
 const mapStateToProps = state => ({
-  width: state.browser.width <= 960 ? state.browser.width - 50 : state.browser.width - 100,
+  width: state.browser.width <= 960 ? state.browser.width - 50 : (state.browser.width > 1540 ? 1440 : state.browser.width - 100),
```
howardchung: Math.min()?

dartharaknus:
Updated.

src/components/App/App.css
```diff
@@ -19,9 +19,15 @@
 .body {
   padding: 50px;
   flex-grow: 1;
+
   @media only screen and (max-width: 960px) {
     padding: 50px 25px;
   }
+
+  @media only screen and (min-width: 1540px) {
```
howardchung: Does this work for homepage (should not apply)?

dartharaknus: It is for homepage too and yes, homepage looks better like this

PR #407, opened 2016-11-05
Flex tabBar, green tracked icon.

blukai:
Tabs now positioned like in material-ui tabBar.

image

@howardchung, @dartharaknus, If u want, u can push changes for castTable/teamFight right here, if needed, I'll help with styling. 🍀

### DIFF
--- a/src/components/Match/Match.css
+++ b/src/components/Match/Match.css
@@ -216,7 +216,7 @@
 
 .tabBar {
   & a {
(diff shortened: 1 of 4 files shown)


howardchung:
Does this work on narrow viewports, or does it encounter the same squishing problem we had with the original tab bar?

blukai:
On narrow viewport scroll appear. No more squishing.

Issue #436, opened 2016-11-07
Cosmetics page need to fix HTTPS-requests to cdn.dota2.com


roman-kosov:
**Problem:**
Cosmetics page was allowed to display insecure content from cdn.dota2.com:
HTTPS...
HTTPS...
HTTPS...
HTTPS...
HTTPS...
HTTP...... 
![HTTPS vs HTTP request](https://cloud.githubusercontent.com/assets/11503318/20060482/b4b66cfc-a53e-11e6-82c2-a05e35a26ee7.png)

**Example:**
https://www.opendota.com/matches/2761811730/cosmetics

howardchung:
Nah, keep this open, we should fix it.  It just require a change to odota/core too.

roman-kosov:
https://github.com/odota/core/issues/1309:
> See `web.js` section where we proxy hero/item/ability images. Need to do the same with the image base path.

 **`web.js`**:
https://github.com/odota/core/blob/master/svc/web.js#L59

howardchung:
Core change is made

PR #438, opened 2016-11-07
Fix HTTP-request to HTTPS

roman-kosov:
Cosmetics page was allowed to display insecure content from cdn.dota2.com
**Issue**: https://github.com/odota/ui/issues/436

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -552,7 +552,7 @@ export const cosmeticsColumns = [heroTdColumn, {
         rel="noopener noreferrer"
       >
         <img
-          src={`http://cdn.dota2.com/apps/570/${cosmetic.image_path}`} role="presentation"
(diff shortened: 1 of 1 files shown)


blukai:
There is an open issue https://github.com/odota/core/issues/1309. Doing it like this is wrong.

roman-kosov:
@blukai thanks you for answer. I understood, @howardchung plans to use "cosmetics images" without cdn.dota2.com?

blukai:
Almost but not 😆 https://github.com/odota/core/blob/master/svc/web.js#L59
The same should be for cosmetics.

howardchung:
Yeah unfortunately it's not quite that straightforward since the Valve CDN has an invalid HTTPS cert.  We need to set up proxying through the core and then point to it.  We can keep this PR open for the eventual fix, though.

dartharaknus:
Agree that this is a wrong implementation. Closing this PR.

Issue #417, opened 2016-11-07
reduce size of hero images in hero selector


howardchung:
This is the heroes page.

PR #439, opened 2016-11-07
Hot Reloading enabled, Filtering of Heroes enabled, more Readability on a few pages

lucidlemon:
So, after ranting a bit on reddit that I don't 100% like the style I thought I might contribute at least something.

I fixed a lot of readability Stuff on the Heroes Page in 3aea3f0, tightening the font size a bit and giving the heroes less max-height. Improves #417 a lot.

While working on this I started to miss hot reloading, so I tried to implement this in a fast way and got it quite working. Only issue: It seems to make a full page refresh upon reloading instead of just the component, but at least you don't have to alt-tab into your browser and refresh the page constantly. This was done in e8baa1f.

I then built a new Filter for the Heroes page in c3ca000 so you can just type the name of the hero you want and get instant results. I'm only focusing on the localized name so far, maybe we should broaden this up. I also rewrote the Mapping function `createHeroList` in Heroes.jsx in order to increase readability and deal with undefined heroes since I have to filter them out. It works instant here, so shouldn't be a performance problem.

I went on to increasing the max width a bit as it breaks the match display thus far, so this might need to get fixed afterwards. I would try to aim at a 960px max width for the body.

The Lannisters send their regards.

/u/karreerose

### DIFF
--- a/src/components/TabBar/TabBar.css
+++ b/src/components/TabBar/TabBar.css
@@ -15,7 +15,7 @@
   & a { /* Tab */
     text-align: center;
     font-weight: var(--fontWeightLight);
-    font-size: 15px;
+    font-size: 14px;
     color: var(--primaryTextColor);
     padding: 10px 14px 16px;
     border-bottom: 2px solid transparent;
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["stage-3", "react", "es2015"],
+  "plugins": ["react-hot-loader/babel"]
+}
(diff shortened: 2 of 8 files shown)


src/components/TabBar/TabBar.css
```diff
@@ -15,7 +15,7 @@
   & a { /* Tab */
     text-align: center;
     font-weight: var(--fontWeightLight);
-    font-size: 15px;
+    font-size: 14px;
```
blukai: var(--fontSizeMedium)

blukai:
Also fix lint.

lucidlemon:
on it.

lucidlemon:
5182499 should solve the eslint problem.

dartharaknus:
Closing this. Since #443 is open and it contains these commits.

PR #441, opened 2016-11-07
Report bug, issue template

blukai:
image

@howardchung Maybe u can supplement issue template and fix mistakes if exists 😆

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -10,6 +10,7 @@ import { Row, Col } from 'react-flexbox-grid';
 // import MenuItem from 'material-ui/MenuItem';
 // import IconMenu from 'material-ui/IconMenu';
 // import AppBar from 'material-ui/AppBar';
+import Bug from 'material-ui/svg-icons/action/bug-report';
 import strings from 'lang';
 import AccountWidget from '../AccountWidget';
 import styles from './Header.css';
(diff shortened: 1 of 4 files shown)


howardchung:
This is very cool, I didn't know about GitHub issue templates.  I'll update the text a bit later.

Does the report bug link work well on mobile (with the nav drawer?)

src/components/Header/Header.jsx
```diff
@@ -36,20 +37,31 @@ const navbarPages = [
...
 const LinkGroup = () => (
   <ToolbarGroup className={styles.verticalAlign}>
     {navbarPages.map(page => (
       <div key={page.name} className={styles.tabContainer}>
         {page.external ?
-          <a href={page.path} className={styles.tab} rel="noopener noreferrer" target="_blank">{page.name}</a> :
-          <Link to={page.path} className={styles.tab}>{page.name}</Link>}
+          <a href={page.path} className={styles.tab} rel="noopener noreferrer" target="_blank">{page.name}</a>
+          : <Link to={page.path} className={styles.tab}>{page.name}</Link>
+        }
```
dartharaknus: Why the change in style?

dartharaknus:
Checked every resolution and looks good.

src/components/Header/Header.jsx
howardchung: I think one of the linting rules changed or something

blukai:
> Does the report bug link work well on mobile (with the nav drawer?)

Yes 🦄

PR #445, opened 2016-11-08
Improve ward map visuals

mjbcopland:
* Enlarged observer wards on the map to reflect their larger AoE compared to sentries
* Replaced team colours with individual player colours

Original | Updated
--- | ---
Original | Updated

### DIFF
--- a/src/components/Match/VisionMap.jsx
+++ b/src/components/Match/VisionMap.jsx
@@ -20,7 +20,7 @@ import {
 const obsWard = (style, stroke, iconSize) => (<svg style={style} width={iconSize} height={iconSize} xmlns="http://www.w3.org/2000/svg">
   <g>
     <title>Observer</title>
-    <circle fill="#ffff00" strokeWidth="5" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.4" />
+    <circle fill="#ffff00" strokeWidth="2.5" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.3" />
   </g>
   <defs>
     <filter id="_blur">
@@ -32,7 +32,7 @@ const obsWard = (style, stroke, iconSize) => (<svg style={style} width={iconSize
 const senWard = (style, stroke, iconSize) => (<svg style={style} width={iconSize} height={iconSize} xmlns="http://www.w3.org/2000/svg">
   <g>
     <title>Sentry</title>
-    <circle fill="#0000ff" strokeWidth="5" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.4" />
+    <circle fill="#0000ff" strokeWidth="2" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.3" />
   </g>
   <defs>
     <filter id="_blur">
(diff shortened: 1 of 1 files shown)


src/components/Match/VisionMap.jsx
```diff
@@ -20,7 +20,7 @@ import {
 const obsWard = (style, stroke, iconSize) => (<svg style={style} width={iconSize} height={iconSize} xmlns="http://www.w3.org/2000/svg">
   <g>
     <title>Observer</title>
-    <circle fill="#ffff00" strokeWidth="5" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.4" />
+    <circle fill="#ffff00" strokeWidth="2.5" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.3" />
```
dartharaknus: Why is this 2.5 and not 2?

src/components/Match/VisionMap.jsx
```diff
@@ -32,7 +32,7 @@ const obsWard = (style, stroke, iconSize) => (<svg style={style} width={iconSize
 const senWard = (style, stroke, iconSize) => (<svg style={style} width={iconSize} height={iconSize} xmlns="http://www.w3.org/2000/svg">
   <g>
     <title>Sentry</title>
-    <circle fill="#0000ff" strokeWidth="5" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.4" />
+    <circle fill="#0000ff" strokeWidth="2" stroke={stroke} r={iconSize * 0.4} cy={iconSize / 2} cx={iconSize / 2} fillOpacity="0.3" />
```
dartharaknus: Can we make this a cyan color?

dartharaknus:
@howardchung I am not too keen on the player color. We already have player specific switches. I think the map should have just team color.

src/components/Match/VisionMap.jsx
dartharaknus: You have 2 for sentries. And since this is just the border, I think both of them should have the same value 2.

mjbcopland:
* When the observer radius is larger than that of sentries, both having the same stroke thickness gives the illusion that observers have a thinner outline (see the [Ebbinghaus illusion](https://en.wikipedia.org/wiki/Ebbinghaus_illusion)). Increasing the thickness negates this effect to give the appearance that they are the same. If you disagree, I can make them actually consistent with each other.

* Feel free to tweak the AoE colours; I kept them as they were. Cyan seems a bit too far in the other direction but perhaps something in between?

#0000FF | #007FFF | #00FFFF
--- | --- | ---
Blue | Azure | Cyan

* The use of individual hero colours was partially motivated by [this comment](https://github.com/odota/ui/blob/master/src/components/Match/VisionMap.jsx#L44). Colours seemed cleaner and easier than icons.

dartharaknus:
Okay, different thickness it is. TIL. 😄 

How about #0055FF?

Can you also add tooltips to each of those circles containing information about who planted the ward, when, who destroyed it if any, when if any

howardchung:
* Maybe leave a comment in the code explaining why we have different stroke widths
* I like #007FFF but we can play around with that range if we want
* Tooltips for icons are probably outside the scope of this PR.  @micaelbergeron is working an improved scrubbable vision map, perhaps he can add tooltips as well.

micaelbergeron:
I'll try to merge this into #469

dartharaknus:
@howardchung What's the plan with moving forward on this one?

howardchung:
Needs to be synced with @micaelbergeron 's merged PR.  I think we can apply the larger ward radius.  Not sure if we want to do the colors as well.

dartharaknus:
Being done in #587

PR #467, opened 2016-11-10
Adds peerDependency on fbjs

rhubarbselleven:
From a clean state, there is a missing dependency on fbjs. As this is not a strict dependency, adding to `peerDependencies` as per https://nodejs.org/en/blog/npm/peer-dependencies/

Without this PR, a clean `npm install && npm run` fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,5 +77,8 @@
   "moduleRoots": [
     "./src",
     "./node_modules"
-  ]
+  ],
+  "peerDependencies": {
+    "fbjs": "^0.8.5"
+  }
 }


howardchung:
I thought npm3+ got rid of peer dependencies.  What version are you using?

rhubarbselleven:
```
npm -version
3.10.8
```

howardchung:
Travis builds from scratch and doesn't encounter this problem.  Are you sure it's an error and not just a warning?

rhubarbselleven:
fair point, totally happy to accept as an environmental thing. However I've tried this as both standalone and using the docker image with the same result. 

Without this change and once the webapp loads, I get a missing dependency on `fbjs/lib/memoizeStringOnly`.

howardchung:
Node version/OS?

rhubarbselleven:
Opensuse 42.1 and nodejs 6.9.1

howardchung:
Yeah I think most people and Travis are using Ubuntu, so it's probably specific to your distribution.

rhubarbselleven:
Why then would I get the same experience when using the docker image?

howardchung:
@dartharaknus can you repro this with docker?  Maybe it's something from the host

dartharaknus:
AFAIK, it shouldn't happen, but I will check.

dartharaknus:
I haven't been able to reproduce this. @rhubarbselleven Can you get a new clone and try the steps `npm install && npm start` and give us the complete log?

Issue #420, opened 2016-11-07
TabBar should break onto lines instead of scrolling


howardchung:
On mobile, scrolling is OK, but on desktop, the scrollbar is very visible and doesn't look good.  We should probably just break onto multiple lines in all cases.

PR #468, opened 2016-11-11
Fix tab bar growing larger than 1440px + FFox tab bar scrolling issues

FlaminSarge:
Fixes being unable to scroll to see other tabs on certain screen sizes, as well as being unable to scroll entirely on FFox
Temporarily papers over the issue that the tab bar doesn't wrap #420

### DIFF
--- a/src/components/TabBar/TabBar.css
+++ b/src/components/TabBar/TabBar.css
@@ -47,6 +47,5 @@
     overflow-x: auto;
     overflow-y: hidden;
     -ms-overflow-style: none;
-    overflow: -moz-scrollbars-none;
   }
 }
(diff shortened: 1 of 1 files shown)


howardchung:
@blukai @dartharaknus is this acceptable?  Will it mess up layout for some resolutions?

dartharaknus:
Isnt size of match page tab bar 1440px?

FlaminSarge:
In my testing on Chrome and FFox, the tab bar elements actually space out based on the width of the window. So the minimum size is 1440px, but they tend to expand out to 1486px as you expand the window (without the @media option set). Setting this to 1439px was causing really odd behavior where it'd be expanded at 1440px, cutting off the chat log element, as in this

FlaminSarge:
While I probably should have just closed this and opened a new PR, I updated this to fix what was really going on: tabs had 14px padding at window width 1440px, rather than 12px.

FFox issue still wasn't solved until the removal of the `-moz-scrollbars-none` rule.

FlaminSarge:
I have NOT confirmed whether `overflow-y: -moz-scrollbars-none` is necessary, but I believe it is not.

PR #469, opened 2016-11-11
Match vision tab overhaul

### LINKED ISSUE #342: use ward logs for ward map
Continuation of https://github.com/odota/core/issues/502

With logged events we can add a slider that allows replaying of ward state.

micaelbergeron:
fixes #342 

This PR addresses #342 and will add a bunch of new features to the Vision Tab, including:

- Dynamic ward filtering
- Player/Ward type filtering
- Time slider (controls visible wards on the map)
- Ward log 
- Counterwarding (require odota/parser#3)

Keep in mind that this is a work in progress.
I've also refactored certain components while working on this page.

There are still refactoring/extracting that needs to be done.

Comments welcome.


src/components/Match/VisionMap.jsx
```diff
@@ -1,136 +1,112 @@
...
+const FixedTable = Fixed(Table);
 
 // TODO Hero icon on ward circles?
 class VisionMap extends React.Component {
-  componentWillMount() {
-    this.setState({
-      enabledIndex: {},
-    });
+  shouldComponentUpdate(newProps) {
+    if (newProps.wardsLog.length == this.props.wardsLog.length) return false;
```
howardchung: use ===

howardchung: This is a boolean so you can return it directly

src/components/Match/VisionMap.jsx
```diff
@@ -1,136 +1,112 @@
...
+    return true;
   }
-  super() {
-    this.updateMap = this.updateMap.bind(this);
+
+  componentDidMount() {
+    window.addEventListener("resize", () => this.forceUpdate());
```
howardchung: I think @mike-robertson will advise that you use redux-responsive's state.browser.width for this

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+       return <a key={tick} onClick={() => props.onTickClick(tick)} className={cls.join(' ')} style={{left: percent + '%'}}>{formatSeconds(tick)}</a>
+     })}
+  </div>
+);
+
+
+class PlayerFilter extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.getObserverCount = () => this.props.player.obs_log.length;
```
howardchung: i'd advise making these functions outside the class, and take props as a parameter.

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+            <Col xs={1}></Col>
+            <Col>{heroTd(player)}</Col>
+          </Row>
+        </Col>
+        <Col xs={12} sm={5}>
+          <Row>
+            <Col xs>
+              <Button {...this.getMuiThemeProps()}
+                      label={obs_count}
+                      disabled={obs_count == 0}
```
howardchung: use !obs_count

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+                      backgroundColor={this.generateFilterKey("observer") in this.props.activeFilters ? filterOff : filterOn}
+                      style={{opacity: obs_count > 0 ? opacityOn : opacityOff}}
+                      onClick={() => onFilterClick(this.generateFilterKey("observer"), player.player_slot, "observer")}
+                      icon={<Avatar size={24} src="http://a19a1164.ngrok.io/apps/dota2/images/items/ward_observer_lg.png" />}
```
howardchung: use API_HOST

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+              />
+            </Col>
+            <Col xs>
+              <Button {...this.getMuiThemeProps()}
+                      label={sen_count}
+                      disabled={sen_count == 0}
+                      backgroundColor={this.generateFilterKey("sentry") in this.props.activeFilters ? filterOff : filterOn}
+                      style={{opacity: sen_count > 0 ? opacityOn : opacityOff}}
+                      onClick={() => onFilterClick(this.generateFilterKey("sentry"), player.player_slot, "sentry")}
+                      icon={<Avatar size={24} src="http://a19a1164.ngrok.io/apps/dota2/images/items/ward_sentry_lg.png" />}
```
howardchung: same API_HOST comment

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+  }
+}
+
+const PlayersFilter = ({ activeFilters, players, onFilterClick }) => (
+  <Paper className={styles['ward-log-player-filter']}>
+    {players.map((p,i) => <PlayerFilter player={p} activeFilters={activeFilters} onFilterClick={onFilterClick} />)}
+  </Paper>
+);
+
+import Immutable from 'seamless-immutable';
```
howardchung: imports should go at top

howardchung: Is there a significant performance improvement when using an immutable library?

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+  let filtered = filters.map(f => iter.call(frozenData, f))
+                        .reduce((o, v) => o.concat(v), []);
+  return _.differenceWith((x, y) => x === y, frozenData, filtered);
+}
+
+const FixedPlayersFilter = PlayersFilter;
+
+class VisionPage extends React.Component {
+  componentWillMount() {
+    Perf.start();
```
howardchung: should this be disabled for production?

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,231 @@
...
+                                       _.sortedIndex(value))(this.state.wardsLog);
+    this.handleViewportChange = _.debounce(50, this._handleViewportChange);
+  }
+
+  computeTick() {
+    const interval = 10 * 60; // every 10 minutes interval
+    return _.rangeStep(interval, 0, this.props.match.duration);
+  }
+
+  _handleViewportChange(value) {
```
howardchung: I think eslint will complain about the use of underscores since javascript doesn't really have any idea of private properties.

src/components/Match/WardLog.jsx
```diff
@@ -0,0 +1,100 @@
...
+import tableStyle from '../Table/Table.css';
+
+import strings from 'lang';
+import _ from 'lodash/fp';
+import { findLast } from 'lodash';
+
+
+// a simple functor that will call the correct function depending on value 
+const threshold = _.curry((start, limits, values, value) => {
+  if (limits.length != values.length) throw "Limits must be the same as functions.";
```
howardchung: !==

src/utility/components.jsx
```diff
@@ -0,0 +1,16 @@
...
+export const Fixed = (Component) => class extends React.Component {
+  shouldComponentUpdate(newProps, newState) {
+    return this.props.style != newProps.style;
+  }
+  render() {
+    return <Component {...this.props} />
+  }
+};
+ 
+export const Pure = (Component) => class extends React.PureComponent {
```
howardchung: what's the difference between this and a function component?

src/utility/components.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export const Fixed = (Component) => class extends React.Component {
```
howardchung: Doesn't the Redux store take care of this kind of comparison already?

src/utility/components.jsx
micaelbergeron: There is no difference, it just reads better when using an existing component.

    const PureRow = Pure(Row);

src/utility/components.jsx
micaelbergeron: The Vision filters are in the local state r/n.

src/components/Match/WardLog.jsx
micaelbergeron: I need to extract this functor in utility

src/components/Match/VisionPage.jsx
micaelbergeron: Yeah that was to test quickly the debounce, I'll figure out a better name

src/components/Match/VisionPage.jsx
micaelbergeron: Isn't react-addons-perf already doing this under the hood?

micaelbergeron: Could be removed.

src/components/Match/VisionPage.jsx
micaelbergeron: Yes. You can compare deeply nested objects/arrays by reference.
I need to extract this though (the whole ward_log should be Immutable, i.e. in renderMatch)

micaelbergeron: Yeah.

src/components/Match/VisionPage.jsx
micaelbergeron: Prototype left-overs.

src/components/Match/VisionPage.jsx
micaelbergeron: Prototype left-overs.

src/components/Match/VisionPage.jsx
micaelbergeron: Will do

src/components/Match/VisionPage.jsx
micaelbergeron: This component will be refactored.

src/components/Match/VisionMap.jsx
micaelbergeron: Let's iterate on this once the component is shipped.

micaelbergeron:
I must lint this.

micaelbergeron:
I think this is ready to ship.

howardchung:
Reviewers:
@albertcui 
@blukai 
@mike-robertson 
@dartharaknus

src/components/Match/Match.css
```diff
@@ -1,5 +1,16 @@
 @import "../palette.css";
 
+:root {
+  --ward-log-height: 50px;
+  --slider-ticks-color: #757575;
```
blukai: I guess this colors should be defined in palette.css

package.json
```diff
@@ -29,7 +29,9 @@
     "flexboxgrid": "6.3.1",
     "heatmap.js": "2.0.5",
     "isomorphic-fetch": "2.2.1",
+    "lodash": "^4.16.4",
     "material-ui": "0.16.1",
+    "moment-duration-format": "^1.3.0",
```
blukai: Do we need this new 7 dependencies?

src/components/Match/Match.css
```diff
@@ -148,6 +159,117 @@
   vertical-align: middle;
 }
 
+.ward-log {
+  width: 100%;
+  font-size: .8em;
```
blukai: We have some defined font sizes in palette.css. I guess 0.8em is somehing like 12px so u can use `--fontSizeSmall`

micaelbergeron: I'll oblige but I think that using px as font size is a bad practice for mobility. I'd rather leave it that way or refactor to use em/rem all around.

package.json
micaelbergeron: Let me explain why those have been added:

- Lodash: I don't feel like implementing my how flow() to compose my filters. I suggest we use the 'lodash/fp' import that is more geared toward functional programming. (I also use: inRange, sortByIndex, map/filter).
- moment-duration-format: I don't think I'm using it, so I guess I'll remove it. Good catch.
- react-addons-css-transition-group: de facto transition component for React
- react-measure: This is to deal with the resizing of the page. @howardchung suggested I use the Redux store size, but I don't see how that can integrate with the flexgrid sizing. @mike-robertson you are welcome to comment.
- seamless-immutable: okay that is a subject of its own, but this manage immutable Object/Array and make it so deeply nested values can be compared with a simple reference equality check. Having immutable objects in the Redux store is considered a good practice.
- react-addons-perf: That is very helpful to dig down React rendering and find performance issues
- react-perf-tool: I think that expose it to the Chrome extension

src/components/Match/Match.css
howardchung: We should link it to --fontSizeSmall then convert that em then

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+      from: 0,
+      to: this.props.match.wards_log.length,
+      wardsLog: props.match.wards_log,
+      filters: {},
+    };
+
+    this.ticks = this.computeTick();
+    this.findPivot = value => _.flow(_.map(x => x.entered.time),
+                                     _.sortedIndex(value))(this.state.wardsLog);
+    this.handleViewportChange = _.debounce(50, this.viewportChange);
```
mike-robertson: why not just do:
```js
this.handleViewportChange = () => setTimeout(this.viewportChange, 50);
```
No reason to drag in a third party library for that, right?

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+      max: props.match.duration,
+      from: 0,
+      to: this.props.match.wards_log.length,
+      wardsLog: props.match.wards_log,
+      filters: {},
+    };
+
+    this.ticks = this.computeTick();
+    this.findPivot = value => _.flow(_.map(x => x.entered.time),
+                                     _.sortedIndex(value))(this.state.wardsLog);
```
mike-robertson: I think sortedIndex provides enough value to not need to be rewritten, but does it provide enough value to bring in all of lodash?
```js
this.findPivot = value => _.sortedIndex(this.state.wardsLog.map(x => x.entered.time), value);
```
Also, I think we should avoid using lodash when it is just a different way of doing native implementations as it doesn't require looking up how the third-party works.

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+    const interval = 10 * 60; // every 10 minutes interval
+    return _.rangeStep(interval, 0, this.props.match.duration);
+  }
+
+  viewportChange(value) {
+    this.setState({ currentTick: value, from: this.findPivot(value) });
+  }
+
+  visibleData() {
+    return PipelineFilter(_.values(this.state.filters),
```
mike-robertson: this is a useful tool lodash provides, but is it worth bringing in the library for it vs.
```js
Object.keys(this.state.filters).map(key => this.state.filters[key])
```

mike-robertson: I'd rather just right that into our own utility module so we don't need all of lodash if you think so.

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+                }
+              </Col>
+              <Col xs={12} md={6} lg={12} className={styles['ward-log-player-filter']}>
+                <Col xs className={styles['filter-header']}>
+                  {strings.general_dire}
+                </Col>
+                {<FixedPlayersFilter
+                  activeFilters={this.state.filters}
+                  onFilterClick={playerFilterClick}
+                  players={_.takeRight(5, this.props.match.players)}
```
mike-robertson:
```js
this.props.match.players.slice(-5);
```

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+            </Row>
+          </Col>
+        </Row>
+        <Row>
+          <Col xs>
+            <SliderTicks
+              value={this.state.currentTick}
+              min={this.state.min}
+              max={this.state.max}
+              className={styles['slider-ticks']}
```
mike-robertson: camelCase

src/components/Match/WardLog.jsx
```diff
@@ -0,0 +1,80 @@
...
+        <Col xs={1} className={styles.timespan} style={{ color: durationColor }}>{formatSeconds(duration)}</Col>
+        <Col xs>{wardKiller}</Col>
+      </Row>
+    </li>
+  );
+};
+const PureRowItem = Fixed(RowItem);
+
+const WardLog = (props) => {
+  const columns = ['type', 'owner', 'entered_at', 'left_at', 'duration', 'killed_by'].map(h => `ward_log_${h}`);
```
mike-robertson: can't we declare this outside of the component to prevent this executing every render?

src/components/Match/WardLog.jsx
```diff
@@ -0,0 +1,80 @@
...
+      <Row className={styles['ward-log-header']} middle="xs">
+        <Col xs={1}>{strings[columns[0]]}</Col>
+        <Col xs>{strings[columns[1]]}</Col>
+        <Col className={styles.timespan} xs={1}>{strings[columns[2]]}</Col>
+        <Col className={styles.timespan} xs={1}>{strings[columns[3]]}</Col>
+        <Col className={styles.timespan} xs={1}>{strings[columns[4]]}</Col>
+        <Col xs>{strings[columns[5]]}</Col>
+      </Row>
+      <ReactCSSTransitionGroup
+        className={styles['ward-log-list']}
```
mike-robertson: camel case

src/components/Match/renderMatch.js
```diff
@@ -92,66 +96,100 @@ function generateTeamfights(match) {
   return (match.teamfights || []).map(tf => computeTfData(tf));
 }
 
+// create a detailed history of each wards
+function generateWardLog(match) {
+  const computeWardData = (player, i) => {
+    const sameWard = _.curry((w1, w2) => w1.ehandle === w2.ehandle);
```
mike-robertson: if we're not using the benefit of this function (accepts multiple arguments or one at a time), why not just
```js
w1 => w2 => w1.ehandle === w2.ehandle;
```
I am not opposed to using this better implementation of currying but if it's never going to be used like that, why include it (unless we think it would be in the future)?

src/components/Match/renderMatch.js
```diff
@@ -92,66 +96,100 @@ function generateTeamfights(match) {
...
+           type,
+           entered: wards[0],
+           left: wards[1],
+         };
+       })
+    ;
+
+    const observers = extractWardLog('observer', player.obs_log, player.obs_left_log);
+    const sentries = extractWardLog('sentry', player.sen_log, player.sen_left_log);
+    return _.concat(observers, sentries);
```
mike-robertson:
```js
observers.concat(sentries)
```

src/components/Match/renderMatch.js
```diff
@@ -92,66 +96,100 @@ function generateTeamfights(match) {
...
+  };
+
+  const imap = _.map.convert({ cap: false }); // cap: false to keep the index
+  const wardLog = _.flow(
+    imap(computeWardData),
+    _.flatten,
+    _.sortBy(xs => xs.entered.time),
+    imap((x, i) => ({ ...x, key: i })),
+  );
```
mike-robertson: If this can avoid using lodash, I would prefer that, but I don't know exactly what 'convert' is doing here. It seems weird to me that we would need to flatten this array one level, but I don't know what the data looks like.

src/components/Match/renderMatch.js
```diff
@@ -190,8 +228,8 @@ function renderMatch(m) {
           identifier = 'fort';
         }
         newPlayer.objective_damage[identifier] = newPlayer.objective_damage[identifier] ?
-          newPlayer.objective_damage[identifier] + player.damage[key] :
-          player.damage[key];
+                                                 newPlayer.objective_damage[identifier] + player.damage[key] :
+                                                 player.damage[key];
```
mike-robertson: not sure I am a fan of this tabbing, but if others are I won't fight it.

src/utility/index.jsx
```diff
@@ -360,3 +373,12 @@ export function unpackPositionData(input) {
   }
   return input;
 }
+
+export const threshold = _.curry((start, limits, values, value) => {
```
mike-robertson: Same comment here, you could just invoke this function like threshold()()()() and manually curry the function definition but it you are against doing that (so you can do threshold(1, 2, 3)(4)), I'll defer to your judgement.

src/utility/index.jsx
```diff
@@ -360,3 +373,12 @@ export function unpackPositionData(input) {
...
+  if (limits.length !== values.length) throw new Error('Limits must be the same as functions.');
+
+  const limitsWithStart = limits.slice(0);
+  limitsWithStart.unshift(start);
+  return findLast(values, (v, i) => _.inRange(limitsWithStart[i], limitsWithStart[i + 1], value));
```
mike-robertson: useful method, but I don't know if it is worth bringing in all of lodash for. Though if we can get away with the 4kb gzipped lodash and you want to use these, I'm down with that. I'm mostly against using lodash when we could instead use native implementations at essentially no extra cost.

src/components/Match/VisionPage.jsx
micaelbergeron: This is not deboucing.

src/components/Match/VisionPage.jsx
micaelbergeron: I could use slice.

src/components/Match/WardLog.jsx
micaelbergeron: Micro-optimisation.

micaelbergeron: Will do.

src/components/Match/Vision/PlayerFilter.jsx
```diff
@@ -0,0 +1,80 @@
...
+    const {
+      player,
+      onFilterClick,
+    } = this.props;
+    const obsCount = this.getObserverCount();
+    const senCount = this.getSentryCount();
+    const [opacityOn, opacityOff] = [1, 0.4];
+    return (
+      <Row
+        className={styles['filter-row']}
```
micaelbergeron: Will do

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+import WardLog from './WardLog';
+import PlayerFilter from './Vision/PlayerFilter';
+import styles from './Match.css';
+
+const SliderTicks = props => (
+  <div {...props}>
+    {props.ticks.map((tick) => {
+      const [t, min, max] = [tick, props.min, props.max];
+      const percent = 100 * ((t - min) / (max - min));
+      const cls = [styles['slider-tick']];
```
micaelbergeron: Will do

src/components/Match/VisionPage.jsx
```diff
@@ -0,0 +1,163 @@
...
+
+// remove this component
+const PlayersFilter = ({ activeFilters, players, onFilterClick }) => (
+  <Paper>
+    {players.map(p => <PlayerFilter player={p} activeFilters={activeFilters} onFilterClick={onFilterClick} />)}
+  </Paper>
+);
+
+const PipelineFilter = (filters, data, iter = Array.prototype.filter) => {
+  const filtered = filters.map(f => iter.call(data, f))
```
micaelbergeron: I'm not 100% sure the implementation is does exactly the same thing.

src/components/Match/renderMatch.js
micaelbergeron: The fp flavor cap the iteratee to a certain arrity, this removes that cap. It is basically to have the index in the iteratee function.

howardchung:
will this integrate #445 ?

src/components/Match/VisionPage.jsx
micaelbergeron: I can remove the prototype reference.

src/components/Match/Vision/PlayerFilter.jsx
```diff
@@ -0,0 +1,80 @@
...
+    super(props);
+    this.getObserverCount = () => this.props.player.obs_log.length;
+    this.getSentryCount = () => this.props.player.sen_log.length;
+    this.getMuiThemeProps = () => ({
+      fullWidth: true,
+      disabledBackgroundColor: filterOff,
+    });
+  }
+
+  generateFilterKey(type) {
```
micaelbergeron: Doesn't seems so.

micaelbergeron:
Let's merge this first then iterate on it if you guys don't mind. #445 is easy to add after.

micaelbergeron:
I didn't remove lodash usage but I tried to use native methods when available and not changing the implementation.

src/components/Match/renderMatch.js
```diff
@@ -92,66 +96,100 @@ function generateTeamfights(match) {
...
+           player: i,
+           key: wards[0].ehandle,
+           type,
+           entered: wards[0],
+           left: wards[1],
+         };
+       })
+    ;
+
+    const observers = extractWardLog('observer', player.obs_log, player.obs_left_log);
```
howardchung: will this crash if `!player.obs_log` (unparsed match?)

micaelbergeron: I fixed this.

PR #472, opened 2016-11-11
Display permanent_buffs

### LINKED ISSUE #450: display permanent_buffs
Example JSON (LC duel):
http://api.opendota.com/api/matches/2763224835

I'll add the mappings obtained by @Crazy-Duck to dotaconstants.

blukai:
fixes #450 

image

Need some hint how to hide this column properly for unparsed matches and matches that have no data about permanent_buffs

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -17,6 +17,7 @@ import {
   transformations,
   percentile,
   sum,
+  unpackPositionData,
 } from 'utility';
 import Heatmap from 'components/Heatmap';
 import {
@@ -313,9 +314,9 @@ export const performanceColumns = [
   }, {
     displayName: strings.th_map,
     tooltip: strings.tooltip_map,
-    field: 'posData',
+    field: 'lane_pos',
     displayFn: (row, col, field) => (field ?
-      <Heatmap width={80} points={field.lane_pos} /> :
+      <Heatmap width={80} points={unpackPositionData(field)} /> :
       <div />),
   }, {
     displayName: strings.th_lane_efficiency,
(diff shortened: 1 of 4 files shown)


howardchung:
The way we've usually done is make the columns definition a function instead of a direct array, and determine whether to add this column in the returned array based on a passed match that you can do an if check on

src/components/Match/matchColumns.jsx
```diff
@@ -193,6 +194,15 @@ export const overviewColumns = match => [{
     );
   },
 }, {
+  displayName: 'Buffs',
+  field: 'permanent_buffs',
+  displayFn: row => ( // buff.permanent_buff, buff.stack_count
+    row.permanent_buffs && row.permanent_buffs.length > 0
```
howardchung: don't need the > 0?

src/components/Match/matchColumns.jsx
```diff
@@ -193,6 +194,15 @@ export const overviewColumns = match => [{
...
+      ? row.permanent_buffs.map(buff => inflictorWithValue(buffs[buff.permanent_buff], buff.stack_count, 'buff'))
+      : ' - '
```
howardchung: no spaces around `-`?

src/components/Match/matchColumns.jsx
```diff
@@ -193,6 +194,15 @@ export const overviewColumns = match => [{
     );
   },
 }, {
+  displayName: 'Buffs',
```
howardchung: Externalize the string

src/components/Match/matchColumns.jsx
```diff
@@ -196,16 +196,15 @@ export const overviewColumns = (match) => {
...
-      },
+    match.players.map(player => player.permanent_buffs && player.permanent_buffs.length).reduce(sum, 0) > 0 ? {
+      displayName: strings.th_permanent_buffs,
+      field: 'permanent_buffs',
+      displayFn: row => (
+        row.permanent_buffs && row.permanent_buffs.length > 0
+          ? row.permanent_buffs.map(buff => inflictorWithValue(buffs[buff.permanent_buff], buff.stack_count, 'buff'))
+          : '-'
+      ),
+    } : {},
```
howardchung: should be `[]` not `{}`.  concat with empty array is a no-op which is what you want.

PR #479, opened 2016-11-12
eslint: allow same precedence operators to be mix

micaelbergeron:
See http://eslint.org/docs/rules/no-mixed-operators.

`10 * (a - b) / 10 * (a - c)` will now be valid.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,6 +20,9 @@
     "import/named": 2,
     "import/no-dynamic-require": 0,
     "jsx-a11y/no-static-element-interactions": [0],
-    "react/forbid-prop-types": [2, { "forbid": ["any"] }]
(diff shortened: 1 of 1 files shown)


mike-robertson:
I'm not a fan of 
```
10 * 10 / 3
```
not being allowed, but the above expression should be done with parens IMO because it's too confusing to read what it means easily.

what it looks like: `(10 * (a - b)) / (10 * (a - c))`
what it is: `((10 * (a - b)) / 10) * (a - c)`

micaelbergeron:
I think you got the point.

howardchung:
I'm ok with this, with signoffs from 1-2 more contributors I'll merge it

howardchung:
Anyone else want to weigh in on this?

mike-robertson:
I'm cool with it as long as we are smart about when to use parens to denote the order of operations more clearly.

PR #481, opened 2016-11-13
Match timeline

### LINKED ISSUE #456: more information for 'firsts'
from odota/core#1347:

eg hovering over first blood should show who got fb, and who died. showing only the timer is very abstract.

blukai:
Improvement & fixes #456


blukai:
TODO:
- [ ] Icons
- [ ] Tooltips

blukai:
TODO2:
- [ ] Display deaths (possibly)
- [ ] Media queries

image
image
image

blukai:
Teamfight tooltip
image

blukai:
image

src/components/Match/Overview/Timeline.jsx
```diff
@@ -0,0 +1,302 @@
...
+import strings from 'lang';
+import ReactTooltip from 'react-tooltip';
+
+import heroes from 'dotaconstants/json/heroes.json';
+import barracksValue from 'dotaconstants/json/barracks_value.json';
+
+import styles from './Timeline.css';
+
+// Need this since some old matches have only `slot`, ex: 271008789
+const slotToPlayerSlot = {
```
howardchung: maybe should handle this kind of sanitation in renderMatch?

src/components/Match/Overview/Timeline.jsx
```diff
@@ -0,0 +1,302 @@
...
+                        place="right"
+                      >
+                        {obj.type === 'firstblood' &&
+                          <section>
+                            {match.players
+                              .filter(player => player.player_slot === obj.player_slot)
+                              .map(player => (
+                                <aside style={{ color: playerColors[obj.player_slot] }}>
+                                  <img
+                                    src={`${API_HOST}/apps/dota2/images/heroes/${heroes[player.hero_id].name.split('npc_dota_hero_')[1]}_icon.png`}
```
howardchung: Did you also want to swap to this in the minimap tooltips (don't remember if you ended up using the minimap icons?)

src/components/Match/Overview/Timeline.jsx
```diff
@@ -0,0 +1,302 @@
...
+                          </section>
+                        }
+                        {obj.type === 'roshan' &&
+                          match.players
+                            .filter(player => player.player_slot === aegis[0][obj.key].player_slot)
+                            .map(player => (
+                              <section key={i}>
+                                <aside style={{ color: playerColors[player.player_slot] }}>
+                                  <img
+                                    src={`${API_HOST}/apps/dota2/images/heroes/${heroes[player.hero_id].name.split('npc_dota_hero_')[1]}_icon.png`}
```
howardchung: we really should move this into some utility function, we're doing it all over the place (several places in match columns)

src/components/Match/Overview/Timeline.jsx
```diff
@@ -0,0 +1,302 @@
...
+                                    role="presentation"
+                                  />
+                                  {player.name || player.personaname || strings.general_anonymous}
+                                </aside>
+                                <span>
+                                  {!aegis[0][obj.key].act && strings.timeline_aegis_picked_up}
```
howardchung: can anything in this chain be undefined?

blukai: no

src/components/Match/Overview/Timeline.jsx
blukai: Yes, but I wont do it now :D

howardchung: open an issue for it then?

src/components/Match/renderMatch.js
```diff
@@ -264,6 +264,17 @@ function renderMatch(m) {
     }
     return newPlayer;
   });
+
+  if (m.objectives && m.objectives.length > 0) {
+    m.objectives.map(obj => (
```
howardchung: Since you're not returning an array here, I prefer forEach, with a regular block statement (so you can use if directly instead of doing the && which I find harder to read

src/components/Match/renderMatch.js
```diff
@@ -264,6 +264,17 @@ function renderMatch(m) {
...
+      obj.player_slot === undefined && (
+        obj.slot > 4
+          ? Object.defineProperty(obj, 'player_slot', { value: obj.slot + 123 })
+          : obj.slot > 0 && Object.defineProperty(obj, 'player_slot', { value: obj.slot })
```
howardchung: isn't slot = 0 a valid slot?

blukai: whoops

src/components/Match/renderMatch.js
```diff
@@ -266,13 +266,15 @@ function renderMatch(m) {
...
-      obj.player_slot === undefined && (
-        obj.slot > 4
-          ? Object.defineProperty(obj, 'player_slot', { value: obj.slot + 123 })
-          : obj.slot > 0 && Object.defineProperty(obj, 'player_slot', { value: obj.slot })
-      )
-    ));
+    m.objectives.forEach((obj) => {
+      if (obj.slot >= 0 && obj.player_slot === undefined) {
+        if (obj.slot > 4) {
+          Object.defineProperty(obj, 'player_slot', { value: obj.slot + 123 });
```
howardchung: `const newObjectives = objectives.map(obj => ({...obj, player_slot: obj.slot ? obj.slot + 123 : obj.slot}))`

blukai: > is the purpose of Object.defineProperty to avoid the eslint warning about assignment to function parameter?

yes, and I think the current solution is the most proper here since no need to define player_slot if slot = -1 that means creeps I guess 🦄

howardchung: We can put obj.slot > 0 as a check in the function, but I think it's probably better to create a new array rather than circumvent the spirit of the eslint rule by avoiding the use of =.

Object.defineProperty will mutate the input, correct?

blukai:
```js
  const newObjectives = m.objectives.map((obj) => {
    if (obj.slot > 0) {
      return {
        ...obj,
        player_slot: obj.slot > 4 ? obj.slot + 123 : obj.slot,
      }
    }
    return {
      ...obj,
    }
  });
```
is it ok now?

howardchung: yeah

howardchung:
Do you want to wait for weekend/more reviews?

blukai:
Idk, I'm not in a hurry 🍀

PR #490, opened 2016-11-15
Doesn't display match if field is null 

### LINKED ISSUE #414: Unfulfilled records show most recent game
Players with a record that they haven't completed yet (e.g. purchased rapier) seem to show their most recent match.

See https://www.opendota.com/players/1900588/records

greg737:
fixes #414

### DIFF
--- a/src/actions/player/playerRecordsActions.js
+++ b/src/actions/player/playerRecordsActions.js
@@ -40,14 +40,16 @@ export const getPlayerRecords = (playerId, options = {}) => (dispatch, getState)
 
   return fetch(`${API_HOST}${getUrl(playerId, options, url)}`, { credentials: 'include' })
     .then(response => response.json())
-    .then(json => Object.keys(json).map(key => ({
+    .then(json => Object.keys(json)
+    .map(key => ({
       name: key,
       value: json[key][key],
       hero_id: json[key].hero_id,
       start_time: json[key].start_time,
       match_id: json[key].match_id,
       game_mode: json[key].game_mode,
-    })))
+    }))
+    .filter(key => key.value !== null))
     .then(json => dispatch(getPlayerRecordsOk(json, playerId)))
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.jsx
```diff
@@ -41,6 +41,13 @@ const getTable = (data, columns, sortState, sortField, sortClick) => (
                 overflow: `${column.field === 'kills' ? 'visible' : null}`,
                 color: column.color,
               };
+              if (row.value === null && column.field !== 'name') {
```
howardchung: I think a safer fix is to apply a filter on the input array that generates the Records table.  This table is used widely and this check may cause random rows not to appear when we use it elsewhere.

howardchung: for reference:
https://github.com/odota/ui/blob/master/src/actions/player/playerRecordsActions.js

You can just add a .filter() onto the .map and remove anything where value is null or undefined

howardchung: Either is acceptable, IMO

src/actions/player/playerRecordsActions.js
```diff
@@ -40,14 +40,16 @@ export const getPlayerRecords = (playerId, options = {}) => (dispatch, getState)
...
+    .map(key => ({
       name: key,
       value: json[key][key],
       hero_id: json[key].hero_id,
       start_time: json[key].start_time,
       match_id: json[key].match_id,
       game_mode: json[key].game_mode,
-    })))
+    }))
+    .filter(key => key.value !== null))
```
howardchung: Technically this isn't a key anymore since we mapped it into an array of objects.  Maybe "record" or "element" is a better name.

Also check for undefined.

howardchung: Boolean(key.value) is more concise and works if we don't have any records that will have a value of 0

Issue #482, opened 2016-11-13
No data appears when connecting from dyanamic IP. Only CSS structure stays.


ArandomB:
My ISP only provides dynamic IP and none of the pages have any data entries. No heroes, No kills, no item, no match id nothing. Only page table borders are displayed. API call shows full data however it doesn't appear on the webpage itself. Problem appears in all browsers. Webpage only appears properly when I use a proxy/VPN with a static IP. Not an issue with adblock as results are same. Never had any problem with previous yasp/opendota on this.

- **Windows 8**:
- **1680x1050**:
- **All browsers: Chrome Firefox Opera**:
- **http://imgur.com/a/rIAg0
![yasp](https://cloud.githubusercontent.com/assets/23440244/20249443/16e400d0-aa1f-11e6-8be5-e05b5a4ed463.jpg)
**:

howardchung:
Can you verify if this endpoint returns data:
https://api.opendota.com/api

If it doesn't work, can you try the http version:
http://api.opendota.com/api

ArandomB:
My original text didn't come through earlier because of comment lines. Please read it now. Both of the API calls shows data, 1st one required captcha. .

howardchung:
This might be related to #435 .  Can you check your browser console for errors when you try to load a page?  Are you using http or https when loading the UI?

ArandomB:
I get this in console:

Failed to load resource: the server responded with a status of 403 ()
http://api.opendota.com/api/matches/2775840834 Failed to load resource: the server responded with a status of 403 (Forbidden)
2775840834:1 Fetch API cannot load http://api.opendota.com/api/matches/2775840834. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://www.opendota.com' is therefore not allowed access. The response had HTTP status code 403. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
http://api.opendota.com/api/players/117886621 Failed to load resource: the server responded with a status of 403 (Forbidden)
2775840834:1 Fetch API cannot load http://api.opendota.com/api/players/117886621. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://www.opendota.com' is therefore not allowed access. The response had HTTP status code 403. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
http://www.opendota.com/matches/2775840834 Failed to load resource: the server responded with a status of 404 (Not Found)
[...]

PR #491, opened 2016-11-15
add credentials to all requests

howardchung:
I don't think it's necessary to send credentials for requests except for metadata (which contains login state).

However, @arandomb reports interestingly ( #482 ) that the metadata endpoint appears to work, while the match one does not (and match does not send credentials).  Maybe sending the credentials makes it work, in which case we may as well send them.


howardchung:
This PR is now inverted (adding credentials to all missing requests, since that appears to resolve #482

PR #495, opened 2016-11-15
added set up for test frameworks

mike-robertson:
I will also add some basic tests that we can model other unit tests off of


mike-robertson:
@howardchung

.babelrc
```diff
@@ -0,0 +1,7 @@
+{
```
howardchung: if we're gonna use babelrc we should move all babel config out of webpack.config.js

package.json
```diff
@@ -7,9 +7,11 @@
     "webpack": "NODE_ENV=production webpack -p --bail",
     "start": "webpack-dev-server --compress",
     "startstage": "API_HOST=http://hchung.westus.cloudapp.azure.com:5000 npm start",
-    "build": "NODE_ENV=production npm run lintnofix && npm run webpack",
+    "build": "npm test && NODE_ENV=production npm run lintnofix && npm run webpack",
```
howardchung: Think the order should go install, lint, test, build?

package.json
```diff
@@ -60,24 +62,35 @@
...
-    "css-loader": "^0.25.0",
-    "eslint": "^3.10.1",
-    "eslint-config-airbnb": "^13.0.0",
-    "eslint-import-resolver-webpack": "^0.7.0",
-    "eslint-plugin-import": "^2.2.0",
-    "eslint-plugin-jsx-a11y": "^2.2.3",
-    "eslint-plugin-react": "^6.7.1",
-    "file-loader": "^0.9.0",
-    "json-loader": "^0.5.4",
+    "babel-core": "6.18.0",
```
howardchung: I moved to using caret dependencies.  Can you resync so the diff only includes the ones you added?

.eslintignore
```diff
@@ -1,2 +1,4 @@
 node_modules
-build
\ No newline at end of file
+build
+testUtils
```
howardchung: is there a reason we don't lint these files?

howardchung:
Is your eslint up to date? seems like a lot of changes that are inconsistent with what's passing master

.babelrc
mike-robertson: yea, I think that would be a good idea.

package.json
mike-robertson: yep

mike-robertson:
Which changes look wrong regarding eslint?

.eslintignore
mike-robertson: I can fix the testUtils. I am not sure about the spec files because there are things it complains about (such as not understanding where `it` and `describe` come from. I am sure there is a way around it but I haven't taken the time to figure it out yet.

package.json
mike-robertson: ahhhh ok that explains why there was a merge conflict. I didn't look closely enough. My bad!

package.json
```diff
@@ -60,23 +62,34 @@
     "wordcloud": "^1.0.5"
   },
   "devDependencies": {
-    "babel-core": "^6.18.2",
+    "babel-core": "^6.18.0",
```
howardchung: intentional downgrade?

package.json
```diff
@@ -60,23 +62,34 @@
...
     "css-loader": "^0.25.0",
-    "eslint": "^3.10.1",
-    "eslint-config-airbnb": "^13.0.0",
-    "eslint-import-resolver-webpack": "^0.7.0",
-    "eslint-plugin-import": "^2.2.0",
+    "css-modules-require-hook": "^4.0.5",
+    "dirty-chai": "^1.2.2",
+    "enzyme": "^2.6.0",
+    "eslint": "^3.8.1",
+    "eslint-config-airbnb": "^12.0.0",
```
howardchung: intentional downgrade to 12.0.0?

mike-robertson: No I'll fix this. The other one shouldn't matter due to the caret.

howardchung:
I think with #507 we can use 15.4?

mike-robertson:
yes I assume that will work. I'll update

PR #496, opened 2016-11-16
Fix Radiant bottom rax positions are up-side-down

yukidaruma:
Screenshot before fix:
chrome_2016-11-16_10-41-40


howardchung:
@blukai review?

PR #512, opened 2016-11-18
Fixes #508

### LINKED ISSUE #508: make rankings MMR display consistent with match display
On matches we use the solo icon + a number. On rankings we use {number} MMR

blukai:


src/components/Heroes/RankingTable.jsx
```diff
@@ -2,14 +2,25 @@ import React from 'react';
...
+import SocialPerson from 'material-ui/svg-icons/social/person';
 
 const rankingColumns = [{
   displayName: '#',
   displayFn: (row, col, field, index) => index + 1,
 }, {
   displayName: strings.th_name,
   displayFn: (row, col, field) => {
-    const subtitle = <span>{`${row.solo_competitive_rank || strings.general_unknown} ${strings.th_mmr}`}</span>;
+    const subtitle = (
```
howardchung: should we refactor into a function since we want the same result?

PR #516, opened 2016-11-19
Add map to teamfight so that people can click that instead of tabs

### LINKED ISSUE #415: Teamfights page Tabs need to be improved
doesn't give out much information.

### LINKED ISSUE #419: put back hero kills/map in teamfights

### LINKED ISSUE #434: Teamfights page Tabs need to fix width
Example:
over 18 fights in game
https://www.opendota.com/matches/2741286001/teamfights

Monitor is 1440px:
Teamfights fullscreen

Solution:
– 2 row and more
– Dropdown Menu

### LINKED ISSUE #519: Divide radiant/dire heroes in Teamfights

mike-robertson:
fixes #415 
fixes #434 
fixes #419 
fixes #519

WIP. Part of this PR is fixing the deaths_pos array in teamfight objects (first commit).


mike-robertson:
fullscreen:


hover state:


lots of items or abilities: 


mobile:

src/components/Match/TeamfightMap/index.jsx
```diff
@@ -0,0 +1,174 @@
...
+      <div>
+        <span className={styles.goldChange}>
+          {isRadiant(radiantGoldDelta) ? radiantGoldDelta : radiantGoldDelta * -1}
+        </span> Gold
+      </div>
+    </ReactTooltip>
+  </div>
+);
+
+const avgPosition = ({ deaths_pos: deathPositions }) => {
```
howardchung: does this have the potential for being way off if the teamfight occurs over a large area?

src/components/Match/renderMatch.js
```diff
@@ -56,7 +56,7 @@ function generateGraphData(match) {
   return {};
 }
 
-function generateTeamfights(match) {
+function generateTeamfights({ players, teamfights = [] }) {
```
howardchung: does this fail if teamfights is null?  The destructure only assigns a default value if the field is undefined?

howardchung:
Is it possible to show the individual hero deaths somehow?  That's something the old UI used to do. 

I guess if we do that we can't really use the map as a selector.

src/components/Match/TeamfightMap/index.jsx
mike-robertson: It's just an average of all the position data, so it could be off if the teamfight was that long. I was thinking about enhancing this in the next iteration to show a little tombstone or something at each death when selected, so that would sort of solve that issue.

src/components/Match/renderMatch.js
mike-robertson: Probably, I can move this back to where it was null checking before.

dartharaknus:
Can you also please fix #519 in this PR? thanks.

src/components/Match/Overview/Timeline.jsx
```diff
@@ -165,38 +167,16 @@ export default ({ match }) => {
...
-                                  ).map(player => (
-                                    <div style={{ color: playerColors[player.player_slot] }}>
-                                      {player.name || player.personaname || strings.general_anonymous}
-                                    </div>
-                                  ))
-                                }
-                              </aside>
+                              <PlayerThumb
+                                {...match.players.find(player =>
+                                  player.hero_id === heroesArr('find')(hero => hero.name === obj.key).id)}
```
howardchung: What happens if the hero can't be found?

src/components/Match/TeamfightMap/index.jsx
```diff
@@ -0,0 +1,296 @@
...
+    {isRadiant(radiantGoldDelta) ? radiantGoldDelta : radiantGoldDelta * -1}
+    <img src={`${API_HOST}/apps/dota2/images/tooltips/gold.png`} role="presentation" />
+  </div>
+);
+
+const getIconStyle = radiantGoldDelta => (isRadiant(radiantGoldDelta) ? styles.radiant : styles.dire);
+const getSelectedStyle = radiantGoldDelta =>
+  (isRadiant(radiantGoldDelta) ? styles.radiantSelected : styles.direSelected);
+
+// TODO - fix this bug where radiant is always either unstyled or gets the 'both' string here
```
howardchung: still a bug?

src/components/Match/renderMatch.js
```diff
@@ -56,44 +56,80 @@ function generateGraphData(match) {
...
   const computeTfData = (tf) => {
     const newtf = {
       ...tf,
       deaths_pos: [],
+      radiant_networth_advantage_delta: 0,
       radiant_gold_delta: 0,
+      dire_gold_delta: 0,
```
howardchung: Can't we just use radiant_gold_delta?  It's negative if the teamfight was a win for dire.

src/components/Match/renderMatch.js
```diff
@@ -56,44 +56,80 @@ function generateGraphData(match) {
   return {};
 }
 
-function generateTeamfights(match) {
+function generateTeamfights({ players, teamfights = [] }) {
   const computeTfData = (tf) => {
     const newtf = {
       ...tf,
       deaths_pos: [],
+      radiant_networth_advantage_delta: 0,
```
howardchung: What's the purpose of this new property?

src/components/Match/renderMatch.js
```diff
@@ -93,6 +102,31 @@ function generateTeamfights({ players, teamfights = [] }) {
         deaths_pos: playerDeathsPos,
       };
     });
+    // We have to do this after we process the stuff so that we will have the player in
+    // the data instead of just the 'teamfight player' which doesn't have enough data.
+    newtf.deaths_pos = newtf.deaths_pos
+      .map(death => ([{
+        ...death,
+        killer: newtf.players
+          .find(killer => killer.killed[heroes[death.player.hero_id].name]),
```
howardchung: Still needs to be addressed?

dartharaknus:
I am not convinced about the teamfight selection. How would it work when 2 of them occur at the same place?

What about using the timeline from the overview page to select the teamfight?

dartharaknus:
Actually, let's use both map and the timeline to be allowed to select the teamfight.

src/components/Match/renderMatch.js
mike-robertson: I just renamed this from radiant_gold_delta since it was confusing

src/components/Match/renderMatch.js
mike-robertson: radiant_gold_delta and dire_gold_delta are now actually the gold_delta for that fight (what the team gained/lost). the radiant_networth_advantage_delta is what was previously radiant_gold_delta. This way we have the option to display how the teams did as well as the networth advantage change.

mike-robertson:
I thought about that being a good future enhancement (selecting from the timeline). @blukai I did try using the swords for the graves, and what I learned from that is that it is pretty hard to see the crossed swords when they are close to each other. I can change it but I think it will be harder to see how many teamfights there actually were and select a teamfight when they are all close together.

src/components/Match/Overview/Timeline.jsx
mike-robertson: This is existing behavior, I just factored it out into a component. I think it selects a 'blank' icon for that hero that presumably takes up the same space or something.

src/components/Match/renderMatch.js
howardchung: You may want to avoid using the term "networth" since gold and net worth are two separate stats.  This is actually gold (we don't track net worth right now)

mike-robertson:
Also, @dartharaknus , I was thinking that we would add another way to select the teamfight. Right now where we display the time of the fight under the map, we could convert that into a dropdown. That way if there are fights close to each other, it is easier to select a specific teamfight (they would be labeled by the times the occurred and possibly the networth delta team winner (radiant or dire icon))

src/components/Match/Overview/Timeline.jsx
howardchung: OK, but we should confirm that it doesn't crash, otherwise this may break all match pages when a new hero is released.

src/components/Match/renderMatch.js
mike-robertson: ahh, I see. I'll update that name then to gold_advantage

src/components/Match/Overview/Timeline.jsx
mike-robertson: I just changed it to be safe so it shouldn't happen

mike-robertson:
I could do that, but would it be better than using an svg?

howardchung:
I personally prefer css over svg since it's human readable

PR #527, opened 2016-11-21
Distinguish team (radiant/dire) in Chat

### LINKED ISSUE #520: Distinguish team (radiant/dire) in Chat

covertPZ:
fixes #520 

Fix for issue #520 
Currently implemented by adding a new `heroTdChat` and `heroTdChatColumn` for use only in `chatColumns`, which displays Radiant/Dire icon to the left of the hero portrait.


howardchung:
Should we do this as a separate column?

covertPZ:
So I actually did do it as a separate column at first, but spoke with dartharaknus and was advised to go this direction. I can provide screenshots of both to look at and discuss.

howardchung:
I'm leaning toward separate column since it seems like less duplication but sure, screenshots would help

covertPZ:
Agreed. A new column was much simpler and could allow sorting if we wanted that in the future.

dartharaknus:
Sorting the chat doesn't make sense which is why I didn't like the separate column thing. But, yeah, we can look at the screenshots and see which one is better

covertPZ:
Only reason I said it would allow sorting is because right now you can sort the chat by hero (not sure if this is desired either). Maybe this is another issue to fix since once you sort by a hero the only way to get the chat back in order is by reloading the chat tab.

dartharaknus:
Let's disable sorting for the chat completely.

covertPZ:
teamcolumn
Updated to put the team icon in its own column. If we don't like this I can pull another screen of the icon as part of the heroTd for comparison.

dartharaknus:
It being a separate column wastes a lot of space IMO. Can you give use the other screenshot too?

covertPZ:
teamherotd
Here's what it looks like as part of the TableHeroImage. We probably want to bump up the margin if we do this method.

src/components/Match/matchColumns.jsx
```diff
@@ -51,6 +52,13 @@ export const heroTdColumn = {
   sortFn: true,
 };
 
+export const heroTdChatColumn = {
+  displayName: 'Player',
```
howardchung: externalize the string

src/components/Match/matchColumns.jsx
```diff
@@ -51,6 +52,13 @@ export const heroTdColumn = {
   sortFn: true,
 };
 
+export const heroTdChatColumn = {
```
howardchung: Maybe `{...heroTdColumn, sortFn: false}` to avoid duplication?

src/components/Match/matchColumns.jsx
```diff
@@ -428,7 +436,20 @@ export const supportColumns = [
 ];
 
 export const chatColumns = [
-  heroTdColumn, {
+  {
+    displayName: strings.filter_is_radiant,
+    field: 'team',
```
howardchung: idk if team is a valid property here.  We don't use it anyway, so maybe should leave it blank

src/components/Match/matchColumns.jsx
covertPZ: So I actually wasn't sure of a way to do that. Was not aware that you could use rest in that way. TIL

src/components/Match/matchColumns.jsx
covertPZ: This string is not localized in any of the other table headings as well. I can open as a new issue, or just fix them in here. Thoughts?

src/components/Match/matchColumns.jsx
howardchung: Note this is still babel stage 3: https://babeljs.io/docs/plugins/preset-stage-3/

It won't work in Node, you have to do Object.assign() instead

src/components/Match/matchColumns.jsx
```diff
@@ -438,18 +431,18 @@ export const supportColumns = [
...
-            <IconRadiant className={styles.iconRadiant} /> :
-            <IconDire className={styles.iconDire} />
+          <IconRadiant className={styles.iconRadiant} /> :
+          <IconDire className={styles.iconDire} />
         }
       </div>
     ,
   },
-  heroTdChatColumn, {
+  Object.assign(heroTdColumn, {sortFn: false}),
```
howardchung: you need to make {} the first argument.  Or just use the rest syntax.  I pointed out it's not officially part of ES6 yet, but we are using it all over the place anyway, and babel transpiles it for us.

src/components/Match/matchColumns.jsx
```diff
@@ -438,18 +431,18 @@ export const supportColumns = [
 export const chatColumns = [
   {
     displayName: strings.filter_is_radiant,
-    field: 'team',
-    displayFn: row =>
+    displayFn: row =>
```
howardchung: you might have to put field, just do field: ''

dartharaknus:
Okay, let's have a separate column. The "Team" header looks nice.

src/components/Match/matchColumns.jsx
covertPZ: Sure. What exactly is the field value's intended use? I assumed it was related to sorting, and excluding it worked fine on my local.

howardchung: the table will try to read row[field], but if you supply a displayFn then this behavior is overridden.

src/components/Visualizations/Table/HeroImage.css
```diff
@@ -161,3 +161,26 @@
...
+  fill: black !important;
+}
+
+.iconRadiant, .iconDire {
+  height: 26px !important;
+  width: 26px !important;
+  margin-right: 6px;
+}
+
+.teamIcon {
```
blukai: Where is it used?

src/components/Match/Match.css
```diff
@@ -291,3 +291,12 @@
   opacity: .8;
   fill: var(--textColorPrimary);
 }
+
+.chatTeamIcon svg {
+  vertical-align: top;
+  height: 26px !important;
+  width: 26px !important;
+  margin-right: 6px;
```
blukai: I think need to increase margin

covertPZ: Do we want the teamIcon to be center aligned in the Team column, or left aligned on like the rest of the columns in the table?

src/components/Visualizations/Table/HeroImage.css
covertPZ: Removing this code as well since it's not used.

PR #528, opened 2016-11-21
Fix some issues, new attack icon (from dota2), login icon

blukai:


src/components/Header/Header.jsx
```diff
@@ -21,29 +21,35 @@ import BurgerMenu from '../BurgerMenu';
...
+};
 
 const LogoGroup = ({ width }) => {
-  if (width <= mobile) {
-    navbarPages.push({
-      name: strings.app_report_bug,
-      external: true,
-      path: 'https://github.com/odota/ui/issues',
-    });
+  if (width < mobile && navbarPages.map(page => page.path === '//github.com/odota/ui/issues').filter(Boolean).length < 1) {
```
howardchung: What if we make "report bug" part of the array permanently?  That would move it to the left side, but it's cleaner code than this splicing.

blukai: Noooo 😛

howardchung: How about:

* Save the length of the array
* Push if the navBarPages.length === savedLength

blukai: Good, but how?

src/components/Header/Header.jsx
```diff
@@ -21,29 +21,38 @@ import BurgerMenu from '../BurgerMenu';
...
-    navbarPages.push({
-      name: strings.app_report_bug,
-      external: true,
-      path: 'https://github.com/odota/ui/issues',
-    });
+  if (width < mobile && navbarPagesLength === navbarPages.length) {
+    navbarPages.push(navbarBug);
+  }
+
+  // If `bug` item in navbar and it's becoming wider, remove `bug` from navbarPages
```
howardchung: Actually how about:
* Define a new const that's a copy of navbarPages in this function's scope
* Push report bug onto it if width is < mobile
* That way we avoid mutating navbarPages and having to do the check, as well as avoiding the need to splice

howardchung:
```
const extendedNavBar = [...navBarPages];
if (width < mobile) {
extendedNavBar.push(navbarBug);
}
```

blukai:
@howardchung, a bit different way, without push 🦄

PR #535, opened 2016-11-24
Fix picks_bans

blukai:


src/components/Match/Overview/Overview.jsx
```diff
@@ -14,7 +14,12 @@ export default {
   content: match => (
     <div>
       <Timeline match={match} />
-      <TeamTable players={match.players} columns={overviewColumns(match)} heading={strings.heading_overview} />
+      <TeamTable
+        players={match.players}
+        columns={overviewColumns(match)}
+        heading={strings.heading_overview}
+        picksBans={match.picks_bans}
```
howardchung: If we're only ever going to render PicksBans here why not just directly render it under TeamTable instead of passing it down?

PR #541, opened 2016-11-24
add css linting

### LINKED ISSUE #530: Add CSS linter
The current css code is not really styled well. We should have a linter to format it properly.

mike-robertson:
fixes #530


mike-robertson:
@howardchung 
@blukai

.stylelintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-css-modules"
+  ],
+  "rules": {
+    "property-case": ["lower", { "severity": "warning" }],
+    "property-no-unknown": [true, { "severity": "warning" }],
+    "shorthand-property-no-redundant-values": false,
+    "indentation": "tabs"
```
blukai: Got warning on line 9 and 10, I guess `indentation` should be 2 🍀

mike-robertson:
btw I didn't format the palette.css because it would make some things harder to read. We can fix that file later when we are sure which rules we care about and which we don't.

package.json
```diff
@@ -10,6 +10,7 @@
     "build": "npm run lintnofix && npm test && npm run webpack",
     "lint": "eslint . --ext .js,.jsx --fix",
     "lintnofix": "eslint . --ext .js,.jsx",
+    "lint:css": "stylelint \"src/**/*.css\"",
```
howardchung: i prefer single quotes to escape chars, if possible

howardchung:
Are we adding this to CI?

mike-robertson:
I think we can. I will make the change with the quotes and add it to the build process I guess.

PR #543, opened 2016-11-26
wrap tabbar elements to new line

### LINKED ISSUE #420: TabBar should break onto lines instead of scrolling

howardchung:
fixes #420 

Reviewers:
@mike-robertson 
@dartharaknus 
@blukai


mike-robertson:
Why don't we just break onto multiple lines in one screen size and scroll on others? I think breaking onto multiple lines is very bad in mobile; at least one whole screen length will be tabs before content.

src/components/TabBar/TabBar.css
```diff
@@ -41,11 +41,4 @@
     border-color: var(--primaryLinkColor);
     color: color(var(--colorMuted) lightness(85%));
   }
-
-  @media only screen and (max-width: 1439px) {
-    white-space: nowrap;
```
mike-robertson: Change this media query to be smaller,  add media queries to the overflow code. This solution won't work for mobile users.

howardchung: Honestly I'm not that big of a fan of the scrolling on mobile either.  While it's better than on desktop since the scrollbar isn't as large, the fact that it scroll isn't immediately obvious.

Currently with the 13/14 elements we have this breaks onto 4 lines (about 1/3 screen height).  I tested it in the Chrome mobile view and it looks OK to me.

mike-robertson: Let me check it out on your cloud box, maybe it's not as bad as I thought. What's the URL again?

mike-robertson: I still prefer the scrolling on mobile, but it's not as bad as I thought. I think we can do better than this, but if you want to go ahead with this merge I won't hold it back.

PR #550, opened 2016-11-29
Hero link to filtered matches by hero_id

blukai:


howardchung:
For peers this is done via a link on the number of matches.  They should be consistent?

blukai:
`transformations.hero_id` used in different places. So I just made a thing like for subtitle that determine where `heroName` should point to matches played on this hero and where not.

howardchung:
Yeah but we could put the link on the MP column and leave transformations.hero_id as is?

howardchung:
Try merging master into this, should fix the CI

src/components/Player/Pages/Heroes/playerHeroesColumns.jsx
```diff
@@ -1,13 +1,14 @@
...
 import { TablePercent } from 'components/Visualizations';
 import strings from 'lang';
 
-export const playerHeroesOverviewColumns = [{
+export const playerHeroesOverviewColumns = playerId => [{
   displayName: strings.th_hero_id,
   tooltip: strings.tooltip_hero_id,
   field: 'hero_id',
-  displayFn: transformations.hero_id,
+  displayFn: row => <Link to={`/players/${playerId}/matches?hero_id=${row.hero_id}`}>{transformations.hero_id(row)}</Link>,
```
howardchung: games column, not hero_id column?

blukai: what❔

howardchung: the filter link for peers/pros is on the column with the number of games.  For heroes it should be the same (so the column with field: 'games'?

blukai: No, I don't think so. There is no reason to add new column when we can make hero clickable.

howardchung: It's not a new column, it's the "matches played" column that already exists.

howardchung:
LGTM.  perhaps in the future we can also add the links for with/against heroes, but i guess we can do that if someone asks for it :)

PR #565, opened 2016-12-04
use template strings for time

### LINKED ISSUE #423: Translation: string templating, plurals and so on
Hello,
I've see a bunch of examples where a sentence is split into smaller parts and there is some dynamic value in between. 

An example of this are `time_` strings. I assume this ends up as `2 [hours][s] 5 [minutes][s] [ago]`. For Bulgarian, the `ago` should be in front while the words for hours and minutes have different endings for their plural form. Needless to say, this will be a problem in other languages.

One way to do it would be to have a string for plural and one for singular and finally a template `{hours} {minutes} {seconds} ago`. Or you could have a template for all possible combination. Valve does that sometimes. There are lots of other examples where a sentence is split up and would make translation difficult to impossible.

### LINKED ISSUE #433: Declension of cardinal numerals issue for russian
In other language there is problems to. Eg. in russian there is no problem with "ago", but the declension problem is very strong. Just because the phrases in Russian formed otherwise, for this type of entry is required to specify three forms of the word, which are used for some rule requires some comparison operations.
More details can be found in the [Wikipedia](https://en.wikipedia.org/wiki/Russian_declension#Declension_of_cardinal_numerals), 4th paragraph.

For now, i leave only reductions of this words in russian ("сек.", "ч.") on translation page. However, if you want to fix this issue, here the proper JS-solution for russian language (AIK, function body syntax is compatible to React): https://gist.github.com/realmyst/1262561 - but i'm not sure that it's necessary for other language. It looks like that it's the unique russian issue...

howardchung:
fixes #433 
fixes #423 

Reviewers:
@kozhilya
@rossengeorgiev


kozhilya:
Well... It may be better than previous solution, but still is not perfect. As i said in issue, in russian there is 2 prular form of word. So, it's still impossible to put full words in this variant.
Is it possible to use default locales by momentjs? [Their solution](https://github.com/moment/moment/blob/develop/src/locale/ru.js) already have all required functions and texts.

P.S. [Offtop] I'm sorry for late answer, I was lost in the endless notifications from Github...

howardchung:
We can, but the reason we stopped using it is that it's a pretty heavyweight dependency (400kb uncompressed).

kozhilya:
I found [page](http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms) showing that problem with prular forms is regular for many Slavic languages (Russian, Romanian, Czech, etc), Celtic languages (Irish, Welsh), Arabic language and some other.

May be then use default locales only for several languages with unusual sentences format, and use current solution in project for other? Just leave that strings empty, check it in condition statements and then load momentjs locale?
Or export these locales with definition only `relativeTime` field. In the [file](https://github.com/moment/moment/blob/develop/src/locale/ru.js), that I used in my previous post, `locale.relativeTime.m` defined as function `relativeTimeWithPlural` which solves the problem with prular forms. Same as for [Bosnian](https://github.com/moment/moment/blob/develop/src/locale/bs.js) or [Polish](https://github.com/moment/moment/blob/develop/src/locale/pl.js)...

PR #567, opened 2016-12-04
Show radiant/dire team name if available

### LINKED ISSUE #538: show radiant/dire team name if available

covertPZ:
fixes #538 

A first pass at making the string modifications. Would like to have some discussion of ways we may want to be passing this data to the `TeamTable` since I feel the way I've implemented may be a bit clunky.
Fix for #538


howardchung:
@blukai for review

howardchung:
I'm with this approach.  What do you think is clunky about it?

covertPZ:
Mainly the addition of the two props to the `TeamTable` component. We also pass the players value from match into the component, so would we want to just pass the entire match to the component or would that be less efficient/worse practice?

howardchung:
@mike-robertson might want to weigh in.  Originally, TeamTable took the match as input.  It was refactored later to take players only, but now we want additional data from match, so do we want to add new props or go back to passing match?

howardchung:
resolve conflict?

covertPZ:
Sorry, I've been away and haven't been able to get on my dev machine to resolve the conflict. I can update when I return at the beginning of the week.

howardchung:
No worries!  I added you to the org so you can make a branch directly on the repo.  That way it's easier for one of the other contributors to push to your branch for any last-minute cleanup like resolving conflicts.

howardchung:
Actually it looks like GitHub added a UI to support interactive merging directly from the site :)

PR #568, opened 2016-12-04
Enhancement/form filter button

mike-robertson:


src/lang/en.json
```diff
@@ -93,7 +93,9 @@
   "farm_sentries": "Sentries killed",
   "farm_roshan": "Roshans killed",
   "farm_necronomicon": "Necronomicon units killed",
-    
+
+  "filter_button_text_open": "FILTER",
```
howardchung: should we use regular casing here and use css properties to style it (uppercasing?)

mike-robertson: Yea we can do that.

PR #569, opened 2016-12-05
Add styles to tabs in CastTable

### LINKED ISSUE #416: Casts page Tabs need to be restyled
The blue color is from material UI and it doesn't look good. Maybe use TabBar component written by @howardchung and @blukai

covertPZ:
fixes #416

Fix for #416 
Implemented style fix for the time being using just the `Tabs` component. Is the rework of `TabBar` to support content something we want to look into more?


howardchung:
Yeah, tbh I don't really like the tabs component (and it doesn't work that well on mobile).  can merge this as a temporary measure though

PR #575, opened 2016-12-08
Feature/pro player search

mike-robertson:
Add pro players to search results. Only makes 1 request per user session for pro players.


src/lang/en.json
```diff
@@ -357,6 +357,8 @@
   "pagination_last": "Last",
   "pagination_of": "of",
 
+  "pro_players": "Pro Players",
```
howardchung: give this a namespace, maybe "app_pro_players"

mike-robertson: sounds good

howardchung:
You might need to sync master

PR #580, opened 2016-12-10
implement pro matches views

### LINKED ISSUE #514: proMatches/proPlayers listing
After merge of https://github.com/odota/core/pull/1362

Add two new top-level tabs showing a list of recent pro matches and a list of pro players, for easy access

howardchung:
fixes #514 

* Adds a top-level navbar element called "Matches"
* Lists last 100 pro matches
* renamed some reducer exports for consistency


src/components/Matches/index.js
```diff
@@ -0,0 +1,3 @@
+import Matches from './Matches';
```
mike-robertson: if you want to, you could just rename your Matches.jsx to index.jsx to avoid the reimport/export step.

src/components/Matches/Matches.jsx
```diff
@@ -0,0 +1,58 @@
...
+  displayName: strings.general_dire,
+  field: 'dire_name',
+}];
+
+class RequestLayer extends React.Component {
+  componentDidMount() {
+    this.props.dispatchProMatches();
+  }
+  render() {
+    const loading = this.props.loading;
```
mike-robertson: I think it might be better to use the `<Container />` component here.

mike-robertson:
I was also thinking that we could add some sort of 'latest pro matches' feed to the homepage. One thing that I think we are lacking in right now is the only way to get to content (unless you are signed in) is to search players. It doesn't need to be right now, but I think it might be a cool thing to add in at some point in the future.

PR #600, opened 2016-12-15
Fixes #599

### LINKED ISSUE #599: Fix Hero Image registered icon
It's currently messing up the height of the heroimage component which in turn messing up a lot of other components heights.

## Before

before

Sometimes 36px, sometimes 55px depending on if the user is registered

## After

after

Constant 36px

dartharaknus:


howardchung:
probably some csslint rule failing.  LGTM, will wait for @blukai 's review

dartharaknus:
@howardchung That was failing because of commits month ago. Maybe it started checking that just now. Anyway, fixed it.

PR #612, opened 2016-12-17
Revamps vision page

### LINKED ISSUE #483: New Vision page feedback
- [x] Use standard headings for Radiant and Dire
- [x] larger map, smaller table (lots of whitespace in player name column)
- [x] ward map should replay the log to get ward state at the PIT on the slider.  Default behavior can be no filtering/show all wards
- [x] ward tooltips can be styled like other tooltips
- [x] ward tooltips should give some contextual information such as placement time, owner, killed by
- [x] integrate PR #445 
- [x] on mobile the ward log gets squished together

dartharaknus:
Fixes #483


src/components/Match/Vision/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+class Vision extends React.Component {
+  static hideVisionLog(playerSlot, type) {
+    return l => l.entered.player_slot === playerSlot && l.type === type;
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.sliderMin = -90;
+    this.sliderMax = props.match.duration;
```
mike-robertson: if you set things based on props in the constructor, you need to do a componentWillReceiveProps check to update the value if the prop value changes.

src/components/Match/Vision/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+  visibleData() {
+    const time = this.state.currentTick;
+
+    return this.props.match.wards_log.filter(ward => time == -90 || (time > ward.entered.time && (!ward.left || time < ward.left.time)));
+  }
+
+  togglePlayerFilter(name, filter) {
+    if (name in this.state.filters) {
+      const oldFilters = Object.assign({}, this.state.filters);
```
mike-robertson: prefer destructuring over Object.assign unless it makes sense to use Object.assign instead.

src/components/Match/Vision/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+      delete oldFilters[name];
+      this.setState({ filters: oldFilters });
+    } else {
+      const newFilter = {};
+
+      newFilter[name] = filter;
+      this.setState({ filters: Object.assign({}, this.state.filters, newFilter) });
```
mike-robertson: Same comment about destructuring

src/components/Match/Vision/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  render() {
+    const visibleWards = this.visibleData();
+    const playerFilterClick = (filterKey, playerSlot, type) => this.togglePlayerFilter(filterKey, Vision.hideVisionLog(playerSlot, type));
+
+    return (
+      <div>
+        <VisionMap match={this.props.match} wards={visibleWards} />
```
mike-robertson: you can destructure props/state at the top of render function to save long dot notations (ex: `const { match } = this.props;`

src/components/Match/matchColumns.jsx
```diff
@@ -706,3 +660,132 @@ export const teamfightColumns = [
...
   },
 ];
+
+export const visionColumns = [
+  heroTdColumn, {
+    center: true,
+    displayName: (
+      <div style={{ display: 'inline-flex', verticalAlign: 'middle' }}>
+        <img height="15" src={`${API_HOST}/apps/dota2/images/items/ward_observer_lg.png`} role="presentation" />
+        &nbsp;PUR
```
mike-robertson: what is PUR? I'm just curious, you don't have to change it

mike-robertson: But you should probably factor this out into the strings

src/components/Match/renderMatch.js
```diff
@@ -160,19 +160,20 @@ function generateWardLog(match) {
       })
     ;
 
-    const observers = extractWardLog('observer', safePlayer.obs_log, safePlayer.obs_left_log);
-    const sentries = extractWardLog('sentry', safePlayer.sen_log, safePlayer.sen_left_log);
+    const observers = extractVisionLog('observer', safePlayer.obs_log, safePlayer.obs_left_log);
+    const sentries = extractVisionLog('sentry', safePlayer.sen_log, safePlayer.sen_left_log);
     return _.concat(observers, sentries);
```
mike-robertson: prefer native method here (observers.concat(sentries) as it also doesn't mutate, just copies.

src/components/Table/TableHeaderColumn.jsx
```diff
@@ -7,12 +7,18 @@ import styles from './Table.css';
...
+  if (column.center) {
+    style.justifyContent = 'center';
+  }
 
   return (
     <MaterialTableHeaderColumn>
       <div
         className={column.sortFn ? styles.headerCell : styles.headerCellNoSort}
         onClick={() => column.sortFn && sortClick(column.field, sortState, column.sortFn)}
+        style={style}
```
mike-robertson: I know you didn't do this, but you can clean up the className.
```css
.headerCellNoSort {
  display: flex;
  flex-direction: row;
  justify-content: space-between;
  text-align: center;
  text-transform: uppercase;
  color: var(--primaryTextColor) !important;
  font-weight: 500;
}

.headerCell {
  composes: headerCellNoSort;
}
```
So we don't need that conditional for headerCell or headerCellNoSort, they are the same. Then, instead of using the style object here, just do a ternary for your justifyContent center.

src/utility/index.jsx
```diff
@@ -39,27 +40,34 @@ export function abbreviateNumber(num) {
...
+
 export function formatSeconds(input) {
   if (!isNaN(parseFloat(input)) && isFinite(input)) {
     const absTime = Math.abs(input);
     const minutes = Math.floor(absTime / 60);
     const seconds = pad(Math.floor(absTime % 60), 2);
+
     let time = ((input < 0) ? '-' : '');
     time += `${minutes}:${seconds}`;
+
```
mike-robertson: I'm not a fan of all the space this creates (adding a space before returns). Doesn't really seem necessary to me

src/utility/index.jsx
```diff
@@ -70,14 +78,17 @@ export const calculateRelativeXY = ({ clientX, clientY, currentTarget }) => {
   // const y = clientY - bounds.top;
   let x = clientX + document.body.scrollLeft;
   let y = clientY + document.body.scrollTop;
+
```
mike-robertson: I am cool with this though (return before if/do/while/etc. blocks)

dartharaknus:
opendota dota 2 statistics

src/components/Match/Vision/VisionMap.jsx
```diff
@@ -0,0 +1,134 @@
...
+        <WardTooltipEnter player={match.players[log.player]} log={log} />
+        {log.left && <WardTooltipLeft log={log} />}
+      </ReactTooltip>
+    </div>
+  );
+};
+
+class VisionMap extends React.Component {
+  componentDidMount() {
+    window.addEventListener('resize', () => this.forceUpdate());
```
dartharaknus: I am not sure, this was already here from previous code. I thought it was needed.

src/components/Match/Vision/index.jsx
dartharaknus: This was part of old code. I pushed new code which doesn't do this anymore.

src/components/Match/matchColumns.jsx
dartharaknus: It's a shorthand for "PURCHASE"

src/components/Match/Vision/VisionMap.jsx
mike-robertson: it might be; i'd just double check cause this would cause 2 renders for window resize if measure does do that (and it seems like it would, but maybe not).

dartharaknus: Looks like I can remove this.

src/components/Table/TableHeaderColumn.jsx
dartharaknus: I don't understand this comment. The CSS you wrote is already in the css file. What exactly are you proposing?

src/components/Match/Vision/VisionFilter.jsx
```diff
@@ -0,0 +1,79 @@
...
+  handleCheckGeneratorTeam(index) {
+    const parent = this.props.parent;
+
+    return event => parent.setTeam(index === 0 ? 'radiant' : 'dire', event.target.checked);
+  }
+
+  columns(index) {
+    const { players } = this.props.match;
+
+    return [
```
howardchung: could this array be generated with a function? seems like a lot of similar code

src/components/Match/Vision/VisionItems.jsx
```diff
@@ -0,0 +1,15 @@
...
+import TeamTable from 'components/Match/TeamTable';
+import { visionColumns } from 'components/Match/matchColumns';
+import strings from 'lang';
+
+const VisionItems = ({ match }) => (
+  <TeamTable
+    players={match.players}
+    heading={strings.heading_vision}
+    columns={visionColumns}
+    radiantTeam={match.radiant_team} direTeam={match.dire_team}
```
howardchung: direTeam should go on new line

src/components/Match/matchColumns.jsx
```diff
@@ -706,3 +660,132 @@ export const teamfightColumns = [
...
+    center: true,
+    displayName: (
+      <div style={{ display: 'inline-flex', verticalAlign: 'middle' }}>
+        <img height="15" src={`${API_HOST}/apps/dota2/images/items/ward_observer_lg.png`} role="presentation" />
+        &nbsp;{strings.th_use_shorthand}
+      </div>
+    ),
+    tooltip: strings.tooltip_used_ward_observer,
+    sortFn: true,
+    displayFn: row => row.item_uses.ward_observer || '-',
```
howardchung: this will crash if item_uses is undefined (unparsed match).  The vision tab won't be clickable if this is the case, but may want to handle it for safety/if the user direct links to it.

src/components/Match/matchColumns.jsx
```diff
@@ -706,3 +660,132 @@ export const teamfightColumns = [
...
+    displayFn: (row, col, field) => field || '-',
+  }, {
+    center: true,
+    displayName: strings.th_tpscroll,
+    tooltip: strings.tooltip_purchase_tpscroll,
+    field: 'purchase_tpscroll',
+    sortFn: true,
+    displayFn: (row, col, field) => field || '-',
+  },
+  visionColumns[1],
```
howardchung: you might want to define these columns as individual variables and reference by name in both places.  Trying to track down specific indexes in the array is more difficult and prone to breakage if someone changes the order or something.

src/lang/en.json
```diff
@@ -587,6 +589,8 @@
   "tooltip_purchase_buyback": "Buyback purchases",
   "tooltip_used_ward_observer": "Number of Observer Wards placed during the game",
   "tooltip_used_ward_sentry": "Number of Sentry Wards placed during the game",
+  "tooltip_used_dust": "Number of times Dust of Appearance is used during the game",
```
howardchung: "was used".  Past tense more accurate, I think?

src/lang/en.json
```diff
@@ -587,6 +589,8 @@
...
+  "tooltip_used_smoke_of_deceit": "Number of times Smoke of Deceit is used during the game",
```
howardchung: same comment about past tense

src/lang/en.json
```diff
@@ -651,6 +655,11 @@
   "teamfight_participation": "Radiant participation/Dire participation",
   "teamfight_score": "Radiant kills/Dire kills",
   "teamfight_radiant_gold_adv": "Net gold advantage for Radiant",
-  "teamfight_radiant_xp_adv": "Net XP advantage for Radiant"
+  "teamfight_radiant_xp_adv": "Net XP advantage for Radiant",
 
+  "vision_expired": "Expired after",
+  "vision_destroyed": "Destroyed after",
+  "vision_all_time": "All time",
+  "vision_placed_observer": "placed Observer at",
```
howardchung: you might want to double check with @kozhilya or another translator but you might want to use a template string here (see time strings) to avoid word order issues in other languages.

dartharaknus:
Addressed every comment.

howardchung:
LGTM

PR #619, opened 2016-12-19
Backpack items & home space fix

blukai:


blukai:
image

howardchung:
Think we should use a separate column for backpack?

src/components/Home/Buttons.jsx
```diff
@@ -13,12 +13,8 @@ export default () => (
       <FlatButton
         label={
           <span className={styles.label}>
```
howardchung: Could we use a single span separated by space?

blukai:
> Think we should use a separate column for backpack?

In this case, items 1 - 6 will be on two lines 😆 
I think I can add backpack icon to make it recognisable

howardchung:
Check a lone druid game to see if it works well there (three rows of items?)

blukai:
image

PR #626, opened 2016-12-21
Update map and use DotaMap component instead of setting it everywhere

### LINKED ISSUE #592: 7.00 map update
- [ ] Map update
- [x] How should we display talents?  Currently @blukai 's ability display omits the attribute bonuses.  Should we start displaying everything? #594

mike-robertson:
fixes #592
WIP


src/components/DotaMap/DotaMap.css
```diff
@@ -0,0 +1,23 @@
+.map {
```
howardchung: I think we should just use detailed only, unless we have a specific need to use both map images.

Also the Dota 2 wiki has links to multiple map versions (pre 6.86, pre 6.82 etc.)  perhaps we should add those too?

src/utility/index.jsx
```diff
@@ -464,3 +464,7 @@ export const hsvToRgb = (h, s, v) => {
 
   return [r * 255, g * 255, b * 255];
 };
+
+
+// 12/13/2016 @ 5:00pm (UTC)
+export const isPost700 = unixDate => unixDate > 1481648400;
```
howardchung: You can copy the getPatchIndex function from core to make it extensible for previous patches as well:
https://github.com/odota/core/blob/master/util/utility.js#L488

howardchung:
presumably this component will also be used on:
* player wardmaps
* match vision page
* match overview

src/components/DotaMap/DotaMap.css
dartharaknus: Nah, let's leave this file as it is. We never know when we want to use both.

But yeah, agree on the other point.

src/utility/index.jsx
dartharaknus: Why this? Why not the talent information or backpack stuff to recognize 7.00?

mike-robertson: It would only work for detecting 7.00. But I think we should display older maps and stuff.

dartharaknus:
@mike-robertson What's the update on this?

mike-robertson:
I actually just spaced over the holidays about this one. It's almost done, I'll probably have it finished in the next day or so.

src/utility/index.jsx
```diff
@@ -468,3 +468,5 @@ export const hsvToRgb = (h, s, v) => {
 
 // 12/13/2016 @ 5:00pm (UTC)
 export const isPost700 = unixDate => unixDate > 1481648400;
+
+export const bindWidth = (width, maxWidth) => (width >= maxWidth ? maxWidth : width);
```
mike-robertson: good idea

mike-robertson:
So I think in the interest of getting this thing out there, we should add an issue to the backlog about supporting all historical maps which I haven't implemented yet.

howardchung:
Can we port the getPatchIndex function in this PR (even if we only decide between two maps?)

mike-robertson:
Yea I think so. The primary challenge of implementing though will be that not every patch includes a map change, so we will need some knowledge of which patches gave map changes. The alternative I can think of would be including a map for every patch, regardless of if there were actually any changes for that patch.

PR #638, opened 2017-01-02
Update to dev-server 2. Still no inplace HMR

dartharaknus:
I have tried to get the HMR working, but no dice. The page still loads completely instead of just the render


package.json
```diff
@@ -5,9 +5,9 @@
   "license": "GPL-3.0",
   "scripts": {
     "webpack": "NODE_ENV=production webpack -p --bail",
-    "start": "webpack-dev-server --compress",
-    "start:noc": "webpack-dev-server",
-    "startstage": "API_HOST=http://hchung.westus.cloudapp.azure.com:5000 npm start",
+    "start": "webpack-dashboard -- webpack-dev-server --inline --compress",
```
howardchung: what does --inline though?

Also does it matter if the executable used is webpack-dashboard rather than webpack?

mike-robertson: Webpack dashboard is just a cool console visualization for webpack. That being said, I am not sure we should make it the default npm start because it has some issues scrolling properly in iterm.

package.json
```diff
@@ -5,9 +5,9 @@
...
+    "start:noc": "webpack-dashboard -- webpack-dev-server --inline",
+    "startstage": "NODE_ENV=production API_HOST=http://hchung.westus.cloudapp.azure.com:5000 webpack-dev-server --compress",
```
mike-robertson: we should do this like `start:stage`. It's a nice convention for scripts to use the colon to extend functionality.

webpack.config.js
```diff
@@ -9,28 +9,46 @@ const postcssCssNext = require('postcss-cssnext');
 const postcssBR = require('postcss-browser-reporter');
 const postcssR = require('postcss-reporter');
 const postcssCF = require('postcss-color-function');
+const DashboardPlugin = require('webpack-dashboard/plugin');
 
 const isProd = process.env.NODE_ENV === 'production';
 
+function HashBundlePlugin() {}
```
mike-robertson: Any reason we moved this to top of file?

dartharaknus: I am planning to replace this with htmlwebpack plugin

package.json
howardchung: @dartharaknus should we use the standard webpack-dev-server for `npm start` and use webpack-dashboard in an alternate command?

dartharaknus:
@howardchung I dont agree. If we want to have the default command with no webpack-dashboard, we should remove it completely. I wanted to that as default

dartharaknus:
Updated.

PR #650, opened 2017-01-14
Enhanced TeamFight Deaths counter

### LINKED ISSUE #648: Enhance display of 'killed' in Teamfights page
On the teamfights page (e.g. https://www.opendota.com/matches/2886620915/teamfights)

It's quite difficult at a glance to tell who lived and who died. Would it be possible to make deaths > 0 stand out in some way? red+bold letters, or tinting the row?

covertPZ:
fixes #648 

Added player_death.png icon and utilize in TeamFightTable to show deaths simpler.


src/components/Match/matchColumns.jsx
```diff
@@ -622,6 +622,12 @@ export const inflictorsColumns = [
   },
 ];
 
+const colorizeDeaths = obj => (row, col, field) => (
+  field !== 0 && (<span>
```
howardchung: `>0` is better?

src/components/Match/matchColumns.jsx
```diff
@@ -622,6 +622,12 @@ export const inflictorsColumns = [
   },
 ];
 
+const colorizeDeaths = obj => (row, col, field) => (
```
howardchung: what's the purpose of `obj` here?

src/components/Match/matchColumns.jsx
covertPZ: You're right. Probably better for clarity.

src/components/Match/matchColumns.jsx
covertPZ: Initially duplicated another display function, forgot to clean up. Just wanted to open the PR before I left for the weekend.

Also need to update the text placement for the count of deaths.

covertPZ:
Sample of two skulls:
image

The other thought was to do something like the skull, with an "x2" in the bottom right corner of it, but I couldn't get a styling that was easily readable and looked good. Open to suggestions for another option though.

src/components/Match/matchColumns.jsx
```diff
@@ -644,6 +644,13 @@ export const analysisColumns = [heroTdColumn, {
   ),
 }];
 
+const playerDeaths = (row, col, field) => (
+  field > 0 && <div className={styles.playerDeath}>
```
howardchung: Feels duplicated to have the markup twice.  Could we render just the number of skulls that `field` indicates?  (In practice it should always be 0, 1, or 2, although 3 is technically possible with aegis/WK ult since those get counted as deaths)

howardchung:
LGTM, looks like just a linting issue?

covertPZ:
Yeah, I forgot to run the lint. Doing that now, then I'll push again.

PR #653, opened 2017-01-15
fix roshan timeline crash

### LINKED ISSUE #652: Games not opening on OpenDota
Some of my recent games refuse to open in OpenDota. They do not open for other people either (my friends are unable to open them) and the device I use does not matter. However, the same matches do open in Dotabuff.

The nonfunctional pages:

https://www.opendota.com/matches/2911914137
https://www.opendota.com/matches/2909997131

Please let me know if the issue is on my end. Thanks!

howardchung:
fixes #652 

Apparently there's no more key data for Roshan kills.  This is causing aegis[0][obj.key] to be undefined, and we try to read player_slot from it.

I also removed the [0] bits from the obj and aegis arrays since it seemed like there was no reason to have an array of arrays there.

Also renamed obj to events to make it clearer.

### DIFF
--- a/src/components/Match/Overview/Timeline.jsx
+++ b/src/components/Match/Overview/Timeline.jsx
@@ -32,9 +32,6 @@ const Timeline = ({
 }) => {
   const preHorn = 90; // Seconds before the battle horn
 
-  const obj = [];
-  const aegis = [];
-
   if (match.objectives && match.objectives.length > 0) {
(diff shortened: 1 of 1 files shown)


howardchung:
pushing this through since this is a hotfix causing the match page to be unavailable on matches with a roshan kill, but feel free to review anyway

PR #663, opened 2017-01-20
Remove flexbox grid

### LINKED ISSUE #315: Remove unnecessary use of flexbox grid
Adding the grid as a crutch to not properly use css isn't worth it. Unless it is absolutely better to use the grid, I will replace it with css. Most of the time we are misusing the grid where we could just do simple flexbox (which will actually be more responsive than hardset breakpoints).

masad-frost:
Fixes #315

Everything should be the same, there might be a couple of pixels difference here and there though.


masad-frost:
I have. To be safe, I'll do an extensive run tomorrow.

masad-frost:
Good to go

Issue #669, opened 2017-01-24
distributions css fix


howardchung:
@masad-frost I think this might be due to the react-flexbox removal.  Maybe one of the necessary classes got removed from the graph container?

http://hchung.westus.cloudapp.azure.com:8080/distributions

masad-frost:
I just took a quick look. It does seem that it was broken after that merge.

masad-frost:
It's probably a loader issue. I'll send a fix.

PR #671, opened 2017-01-25
Fix c3/c3.css not being loaded properly

masad-frost:
fixes odota/ui#669

@howardchung


masad-frost:
I feel like this issue will come up in the future. Maybe postcss files should have their own extension to avoid things like this.

howardchung:
@mike-robertson had to deal with this before.  idk if we had a better solution than this hacky whitelisting at the time.

I suppose renaming all the postcss files to .postcss might be ok, but not sure if there are any unintended side effects.

PR #677, opened 2017-01-28
Remove translated language name

albertcui:


howardchung:
what's the purpose of this?

howardchung:
I guess this is fine, the user will still be able to find the language since there will still be some text in the target language.  It also saves a lot of space.  @mike-robertson ?

PR #688, opened 2017-02-01
Hide text of player header buttons on mobile

### LINKED ISSUE #686: pvgna icon text overflows on mobile
@servercimen 

On mobile, the pvgna text on player pages overflows the page.  Perhaps we could use just the logo on mobile?  (that's the current behavior on match pages)

Repro:
https://www.opendota.com/players/88367253

servercimen:
fixes #686 

Like Match page, text next to all buttons are hidden on mobile screens.


howardchung:
@blukai is this the proper fix?  looks ok to me

PR #702, opened 2017-02-09
Heroes improvements

blukai:
image

image

Also color on percent line depends on percent value.

### DIFF
--- a/src/components/Heroes/columns.jsx
+++ b/src/components/Heroes/columns.jsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import { transformations } from 'utility';
+import strings from 'lang';
+import { TablePercent } from 'components/Visualizations';
+
+const percentileDisplay = (decimal, total) => <TablePercent
+  val={decimal >= 0 ? Number((decimal * 100).toFixed(2)) : 0}
+  total={total && decimal > 0 && Math.floor(total * decimal)}
+/>;
+
+const heroColumn = {
+  displayName: strings.th_hero_id,
+  tooltip: strings.tooltip_hero_id,
+  field: 'hero_id',
+  displayFn: transformations.hero_id,
+};
+
+export default {
+  pro: [heroColumn, {
+    displayName: strings.hero_pick_ban_rate,
+    field: 'pickBanRatePro',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field, row.matchCountPro),
+  }, {
+    displayName: strings.hero_pick_rate,
+    field: 'pickRatePro',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field, row.matchCountPro),
+  }, {
+    displayName: strings.hero_ban_rate,
+    field: 'banRatePro',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field, row.matchCountPro),
+  }, {
+    displayName: strings.hero_win_rate,
+    field: 'winRatePro',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field, row.pro_pick),
+  }],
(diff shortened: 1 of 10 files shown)


blukai:
I'll fix error 🦄

src/components/Heroes/columns.jsx
```diff
@@ -0,0 +1,91 @@
...
+    field: 'banRatePro',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field, row.matchCountPro),
+  }, {
+    displayName: strings.hero_win_rate,
+    field: 'winRatePro',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field, row.pro_pick),
+  }],
+  ranked: [heroColumn, {
```
howardchung: You may want to call this "public" since it includes unranked matches as well.

src/components/Heroes/columns.jsx
```diff
@@ -0,0 +1,91 @@
...
+    displayName: strings.hero_5000_pick_rate,
+    field: 'pickRate5000',
+    sortFn: true,
+    displayFn: (row, col, field) => percentileDisplay(field),
```
howardchung: should we display absolute values here (like we do for pro matches).  I left it out originally due to space concerns but there's more room now

src/components/Heroes/index.jsx
```diff
@@ -150,12 +66,46 @@ class RequestLayer extends React.Component {
     });
     processedData.sort((a, b) => b.pickBanRatePro - a.pickBanRatePro);
     // TODO add filter by month
+    const heroTabs = [{
+      name: 'Professional',
```
howardchung: extract string

src/lang/en-US.json
```diff
@@ -244,7 +244,9 @@
   "heading_region": "Region",
   "heading_patch": "Patch",
   
-  "hero_this_month": "Last 30 days",
+  "hero_pro_heading": "Picks & Bans in Professional Matches",
+  "hero_ranked_heading": "Picks & Bans in Ranked Matches",
+  "hero_this_month": "matches this month",
```
howardchung: should say in last 30 days.  This month implies since the beginning of the calendar month

src/lang/en-US.json
```diff
@@ -244,7 +244,9 @@
   "heading_region": "Region",
   "heading_patch": "Patch",
   
-  "hero_this_month": "Last 30 days",
+  "hero_pro_heading": "Picks & Bans in Professional Matches",
+  "hero_ranked_heading": "Picks & Bans in Ranked Matches",
```
howardchung: public, not ranked

dartharaknus:
Didn't look at the code, but the feature looks good.

src/lang/en-US.json
```diff
@@ -244,7 +244,11 @@
   "heading_region": "Region",
   "heading_patch": "Patch",
   
-  "hero_this_month": "Last 30 days",
+  "hero_pro_tab": "Professional",
+  "hero_public_tab": "Public",
+  "hero_pro_heading": "Picks & Bans in Professional Matches",
+  "hero_public_heading": "Picks & Bans in Public Matches",
+  "hero_this_month": "matches last 30 days",
```
howardchung: we should indicate somehow that the public match data is sampled (doesn't include every match).  It's roughly a 10% sample, but if we are showing absolute numbers I think we need to indicate that.

src/lang/en-US.json
```diff
@@ -244,7 +244,11 @@
   "heading_region": "Region",
   "heading_patch": "Patch",
   
-  "hero_this_month": "Last 30 days",
+  "hero_pro_tab": "Professional",
+  "hero_public_tab": "Public",
```
howardchung: probably just changing this to Public (Sampled) is fine

PR #712, opened 2017-02-16
add highest/lowest mmr matches

howardchung:
* splits matches list into 3 tabs: pro, high mmr, low mmr


blukai:
Is there easy way to get heroes for matches? Would be cool.

howardchung:
Like the hero IDs of the heroes in the match?

blukai:
Yes

howardchung:
We could change the ingestion to store the hero_ids in a new column, but that requires a core/schema change.

I'm not sure if it's possible to use some joins or something to fetch the hero IDs into the match row.

howardchung:
Might be possible with this: http://stackoverflow.com/questions/43870/how-to-concatenate-strings-of-a-string-field-in-a-postgresql-group-by-query

I'll give it a shot!

howardchung:
The above isn't working well since the joins to the tables with the radiant/dire team hero_ids is not subject to the WHERE clause of the main query, so it's assembling the subqueries for the entire table of public_player_matches.

howardchung:
Actually, the following seems to work, although it's kinda messy:
```
          WITH match_ids AS (SELECT match_id FROM public_matches
          WHERE start_time > ?
          ORDER BY avg_mmr ${order}
          LIMIT 100)
          SELECT * FROM
          (SELECT * FROM public_matches
          WHERE match_id IN (SELECT match_id FROM match_ids)) matches
          JOIN 
          (SELECT match_id, string_agg(hero_id::text, ',') radiant_team FROM public_player_matches WHERE match_id IN (SELECT match_id FROM match_ids) AND player_slot <= 127 GROUP BY match_id) radiant_team
          USING(match_id)
          JOIN
          (SELECT match_id, string_agg(hero_id::text, ',') dire_team FROM public_player_matches WHERE match_id IN (SELECT match_id FROM match_ids) AND player_slot > 127 GROUP BY match_id) dire_team
          USING(match_id)
```

PR #715, opened 2017-02-17
Dismissable announcement

blukai:
components/Announce/blogpost.js - blog post announcement, contain test strings, I'll clean it up after.

image
image

TODO:
  - Release announcements
@howardchung, is it possible to not just opening [release] pr, but creating releases via
https://github.com/odota/ui/releases. If so, we can fetch release notes and make automated announcements about releases through github api - https://api.github.com/repos/odota/ui/releases. But also if no, we can just set up label 'release' for each release pr's, that I created for test and use this - https://api.github.com/repos/odota/ui/issues?labels=release&state=closed


howardchung:
Is there any significant advantage to using the releases API?

The PR is made as part of the ship process (putting code into the production branch) but it would be an extra manual step to create a release as well.  It looks like both support returning notes in the "body" property.

blukai:
> Is there any significant advantage to using the releases API?
Nope, I do not think so. I've been considering options.
Then just add the label 'release' for release pr's.

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+  );
+};
+
+const Announce = () => {
+  if (Object.values(blogpost).length === Object.keys(blogpost).length) {
+    // days
+    const sinceBlogpost = Math.round((new Date() - new Date(blogpost.date)) / (1000 * 60 * 60 * 24));
+    const ttl = 7;
+
+    const isDismissed = localStorage && localStorage.getItem('dismiss') === blogpost.link;
```
howardchung: we probably want to use a more unique key, maybe date + title?

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+      </aside>
+    </div>
+  );
+};
+
+const Announce = () => {
+  if (Object.values(blogpost).length === Object.keys(blogpost).length) {
+    // days
+    const sinceBlogpost = Math.round((new Date() - new Date(blogpost.date)) / (1000 * 60 * 60 * 24));
+    const ttl = 7;
```
howardchung: just do the multiplication here to get 7 days in milliseconds and then diff them?  shouldn't be any need to do math.round then.

src/components/Announce/blogpost.js
```diff
@@ -0,0 +1,6 @@
+export default {
```
howardchung: maybe we can use the releases/pull request API to generate this and read it from github?

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import RaisedButton from 'material-ui/RaisedButton';
+import strings from 'lang';
+
+import blogpost from './blogpost';
+import styles from './styles.css';
+
+const BlogpostAnnounce = () => {
```
howardchung: are you planning to create a different-looking announcement mechanism for releases and blog posts?  maybe we can just use the same one?

blukai: I thought to open release notes in modal dialog onClick. But now I think it'll be better to make it similar.

src/components/Announce/blogpost.js
blukai: hmnnn, do u mean fetch data from api for blog repo, like for ui releases? It'll be cool. But need to add a label too.

howardchung: If possible I'd like a centralized way of updating the announcement banner, rather than having blog posts be announceed via code changes and releases via PRs.

Maybe we can make PRs in the blog repo and read the API for that repo?

src/components/Announce/index.jsx
blukai: Ok

src/components/Announce/blogpost.js
howardchung: yeah I'll make a test PR so you have something to work with

howardchung: https://github.com/odota/blog/pull/7 we can remove the release label or something to hide it later.   Maybe you can test without the closed query param since I don't actually want to merge it :)

blukai: Sure

howardchung: Note: you might want to also check the merged_at property of the PR to ensure the PR was actually merged.  I think someone could randomly open a PR, put whatever they want in the text, it gets closed, and they can still tag it with release.

howardchung: Actually looks like only contributors with write access can change tags

howardchung: I backdated the post so it doesn't show up at the top and merged it so you can test as if it were real

howardchung:
If we want to make sure only admins can update the text (rather than any contributor), we can check that the PR was actually merged (merged_at) and targets production branch.

blukai:
https://api.github.com/repos/odota/blog/pulls / https://api.github.com/repos/odota/blog/issues returns nothing, I described this in note to gh support, also bad thing is that there is no way to limit amount of returned results. /pulls doesn't support labels parameter

howardchung:
Hmm maybe we should use the release api then

blukai:
Probably we should establish markup for body of releases for patches and blog. It's can be parsed with regex or something. Maybe first line of body should be the announcement's message?
For example, like this:
```
> Short message, like summary that will be extracted from `body` via regex, maybe use block-quote for this?

And here some notes that will be ignored
```

howardchung:
Can we just use markdown?  I think there are existing markdown libraries for js

blukai:
image

image

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+      const {
+        merged_at: mergedAt,
+        title,
+        body,
+        html_url: link,
+      } = data[0];
+
+      if (mergedAt) {
+        // Fetch ui if blog's pr was merged more than 7 days ago
+        if (new Date().getDate() - new Date(mergedAt).getDate() >= 7 && this.state.repo === firstPrio) {
```
howardchung: getDate() doesn't work if going between months since it only returns the day of the month

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import RaisedButton from 'material-ui/RaisedButton';
+import strings from 'lang';
+import { connect } from 'react-redux';
+import { getGithubPulls } from 'actions/githubPulls';
+import ReactMarkdown from 'react-markdown';
+
+import styles from './styles.css';
+
+const firstPrio = 'blog';
```
howardchung: instead of priorities I think we should have two elements, one for releases and one for blog.  If there is both a recent release and a blog we want to show both?

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+        html_url: link,
+      } = data[0];
+
+      if (mergedAt) {
+        // Fetch ui if blog's pr was merged more than 7 days ago
+        if (new Date().getDate() - new Date(mergedAt).getDate() >= 7 && this.state.repo === firstPrio) {
+          this.setState({ repo: secondPrio });
+        }
+
+        if (localStorage.getItem('dismiss') !== mergedAt) {
```
howardchung: maybe use repo name + pull request ID as a key?  It's unlikely but blog and releases could have the same merge time.

howardchung:
apparently pull requests count as issues, so you can fetch pull requests by label like this:
http://stackoverflow.com/questions/36974157/github-api-is-it-possible-to-list-pull-requests-by-label

howardchung:
After some thought: Let's just centralize all banner updates on the PRs in this repo.  If we want to announce a blog post, we can make a new release (with a version bump or some token change) with a link to the blog post.  Since it's technically shipping something to UI it makes sense to have a PR tracking that change in the history.  It also makes the code simpler.

blukai:
Ok, I'll try to finish it tomorrow 🍀

src/components/Announce/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+        localStorage.setItem('dismiss', value);
+      }
+    };
+
+    this.getDate = (days) => {
+      const msPerDay = 24 * 60 * 60 * 1000;
+
+      const date = new Date(new Date() - (msPerDay * days));
+
+      // Date formatting must follow the ISO8601 standard, which is YYYY-MM-DD--that's year-month-day
```
howardchung: maybe just use this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

PR #718, opened 2017-02-20
Sort by value if items are bought at the same time

masad-frost:
fixes odota/ui#351

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -307,6 +307,14 @@ export const purchaseTimesColumns = (match) => {
       displayFn: (row, column, field) => (<div>
         {field ? field
         .filter(purchase => (purchase.time >= curTime - bucket && purchase.time < curTime))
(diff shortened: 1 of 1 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -307,6 +307,14 @@ export const purchaseTimesColumns = (match) => {
       displayFn: (row, column, field) => (<div>
         {field ? field
         .filter(purchase => (purchase.time >= curTime - bucket && purchase.time < curTime))
+        .sort((p1, p2) => {
+          if (p1.time !== p2.time) {
+            // We're only concerned with sorting by value
+            // if items are bought at the same time, time is presorted
+            return 0;
+          }
+          return items[p1.key].cost - items[p2.key].cost;
```
howardchung: add a check for items[p1.key].  It may not exist if it's a brand new item and we haven't updated the items list yet, and it'll crash the SPA if we try to read cost from null.

PR #736, opened 2017-02-23
Use FormGroup as filter for logs

### LINKED ISSUE #698: search field for logs
instead of checkboxes, it might be more user friendly to allow the user to type a keyword like "kill" or "objective" into a text field and dynamically filter the log based on the input.

This would be similar to the current pro player search implementation, which @mike-robertson did.

masad-frost:
Added filter for heroes
fixes #698

### DIFF
--- a/src/components/Form/FormGroup.jsx
+++ b/src/components/Form/FormGroup.jsx
@@ -24,6 +24,16 @@ function formField({ formSelectionState, addChip, deleteChip }) {
         strict,
         limit,
       } = this.props;
+      const selectedElements = formSelectionState[name];
+      const isSelected = index > -1
+                          ? selectedElements.includes(value.value)
+                          : selectedElements.includes(value);
+      if (isSelected) {
+        // Handle inputs that are already selected
+        this.handleUpdateInput('');
+        return;
+      }
+
       let input = null;
 
       if (index > -1) {
@@ -87,6 +97,7 @@ function formField({ formSelectionState, addChip, deleteChip }) {
           onUpdateInput={this.handleUpdateInput}
           searchText={searchText}
           errorText={errorText}
+          style={{ flex: '1 0 0' }}
           listStyle={{ textTransform: 'uppercase' }}
           floatingLabelFocusStyle={{ color: errorText ? red : blue }}
           underlineFocusStyle={{ borderColor: blue }}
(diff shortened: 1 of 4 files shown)


howardchung:
Instead of chips I was actually just thinking a simple text box.  If there's text in it (and on update), we fuzzy match each log row against the content of the box.

masad-frost:
I don't think that's intuitive. It'll be really ambiguous as to what that
text box does, there's no clear action. What are am I searching for? What
are the types I can filter? Can I search for more than one? What is the
delimiter?

I'm thinking the upside of what you're suggesting is that we can filter by
more things? Like specific runes, specific players, specific
objectives...etc

Up to you, I can change it, LMK.

On Feb 24, 2017 5:55 AM, "Howard Chung" <notifications@github.com> wrote:

Instead of chips I was actually just thinking a simple text box. If there's
text in it (and on update), we fuzzy match each log row against the content
of the box.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/odota/ui/pull/736#issuecomment-282197344>, or mute the
thread

howardchung:
You could label the text box "Search logs" and hopefully it should be apparent -- e.g. once the user starts typing "kills" the logs begin filtering after each keystroke

masad-frost:
OK, I'll take care of it tomorrow if I get time, but let me get the details
down. Would you want it to search for specifics, like "haste" or "first
blood"? Also can you have multiple criterion, comma delimited?

What's the pro player search you mention in the issue references?

howardchung:
https://www.opendota.com/search?q=arteezy

Then clear the search box and start typing.  The pro player section will fuzzy match against the input and start displaying matches.

masad-frost:
Seems good. It makes a lot of sense for it to be free form here since the
player name and team name could be anything where as in the logs column
data is preset. Also we'll lose the ability to decisively filter I.e.
someone's name contains the word kill or something that's a type.

It doesn't make sense at all for me from a UX perspective, but it might be
just me and you're the boss :P. Hopefully I'll find time tomorrow.

howardchung:
I'm fine with what's in this PR too.  Maybe prototype the other approach as well and we can decide which works better (or get some input from Discord)

masad-frost:
Will do

PR #738, opened 2017-02-24
[HOTFIX] fix trends not working

howardchung:
note @masad-frost : Apparently Container is picky about its children.  I think it has to take a single element as child and doesn't like it otherwise.


masad-frost:
Oh, didn't know that. I might've messed some things up in the tablefilterform refactor then. I'll take a look

masad-frost:
Messed up histograms as well, sending a pr.

PR #752, opened 2017-02-26
Fix slow searchbar

### LINKED ISSUE #745: Search bar is really slow when on a player or match page
Typing is so janky, probably a performance issue

masad-frost:
fixes #745

Only triggers pro player search when on search page and it's debounced to make it completely smooth.

### DIFF
--- a/src/components/Search/SearchForm.jsx
+++ b/src/components/Search/SearchForm.jsx
@@ -65,20 +65,20 @@ class SearchForm extends React.Component {
   }
 }
 
-const mapStateToProps = (state) => {
-  const { error, loading, done, query, searchResults } = state.app.search;
-  return {
-    loading,
-    done,
-    error,
-    query,
-    data: searchResults,
-  };
-};
+// const mapStateToProps = (state) => {
+//   const { error, loading, done } = state.app.search;
+//   return {
+//     loading,
+//     done,
+//     error,
+//     query,
+//     data: searchResults,
+//   };
+// };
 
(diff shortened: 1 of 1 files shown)


src/components/Search/SearchForm.jsx
```diff
@@ -1,53 +1,84 @@
...
+  constructor(props) {
+    super(props);
+
+    let query = '';
+    const location = browserHistory.getCurrentLocation();
+    if (location.pathname === '/search') {
+      // Avoid getting ?q= from other paths
+      query = location.query.q;
+    }
+    this.state = { query };
```
masad-frost: FYI, I couldn't get it from the store because it seems like it updates after the search component mounts. Doing it in componentWillReceiveProps might cause some race conditions.

src/components/Search/SearchForm.jsx
```diff
@@ -1,53 +1,84 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import { browserHistory } from 'react-router';
+import { debounce } from 'lodash';
```
howardchung: If debounce isn't too hard to implement without using lodash, I'd like to at some point remove our dependency on lodash at some point to reduce bundle size.

masad-frost: We can use 'lodash.debounce' from npm

masad-frost: Or do https://lodash.com/custom-builds if you want

PR #758, opened 2017-02-27
Add relative bars into match columns

### LINKED ISSUE #86: "showBar" column property for tables
currently implemented for percent columns.  Allow setting a property on any column to display bars relative to the maximum value of that column in the given table.

masad-frost:
resolves #86 

image

- [x] Add relative bars to columns and tables that should use them @howardchung 
- [x] Consolidate current displayFn: TablePercent columns 
  - [x] Support for displaying total field


masad-frost:
I'll work on this later tonight or tomorrow.

@howardchung can you figure out what columns are we adding this to?

howardchung:
Performances table
Farm (unit kills, last hits, gold sources, xp sources)
casts (maybe?)
objectives (objective damage, runes)
vision (bottom table, maybe)
actions table
teamfights tables (maybe)

masad-frost:
Here's a wild idea. What if we make them appear only when the user hovers over column, this way we avoid clutter and can add them to all numeric fields. It also emphasizes the column the users hovers on.

BTW I'm not gonna be able to work on this till Friday 💯. Let me know if you wanna take over or something.

masad-frost:
Rebasing in a bit

src/components/Heroes/columns.jsx
```diff
@@ -23,72 +24,128 @@ export default {
     displayName: strings.hero_pick_ban_rate,
     field: 'pickBanRatePro',
     sortFn: true,
-    displayFn: (row, col, field) => percentileDisplay(field, row.matchCountPro),
+    displayFn,
+    relativeBars: {
```
howardchung: Would it be cleaner to have relativeBars just accept a function that returns the decimal value?   It can assume a divisor of 1

src/components/Match/matchColumns.jsx
```diff
@@ -334,13 +342,17 @@ export const lastHitsTimesColumns = (match) => {
...
+    const minutes = curTime / 60;
     cols.push({
-      displayName: `${curTime / 60}'`,
+      displayName: `${minutes}'`,
       field: i,
       sortFn: row => (row.lh_t ? row.lh_t[curTime / 60] : 0),
-      displayFn: row => (<div>
-        {row.lh_t ? row.lh_t[curTime / 60] : ''}
-      </div>),
+      relativeBars: {
```
howardchung: Shouldn't this take the same value as sortFn?  This should be a relative to max

src/components/Match/matchColumns.jsx
```diff
@@ -365,61 +377,77 @@ export const performanceColumns = [
     tooltip: strings.tooltip_lane_efficiency,
     field: 'lane_efficiency',
     sortFn: true,
-    displayFn: (row, col, field) => (field ? field.toFixed(2) : '-'),
+    displayFn: (row, col, field) => (field ? `${(field * 100).toFixed(2)}%` : '-'),
+    relativeBars: {
```
howardchung: probably this should be treated as a relative to max, efficiency can go over 100 (e.g. alchemist)

src/components/Match/matchColumns.jsx
```diff
@@ -850,7 +917,10 @@ export const visionColumns = [
     ),
     tooltip: strings.tooltip_used_smoke_of_deceit,
     sortFn: true,
-    displayFn: row => (row.item_uses && row.item_uses.smoke_of_deceit) || '-',
+    displayFn: (row, column, value) => value || '-',
```
howardchung: If we didn't set the `field` property in the columnDef I think value is always undefined since I think value is row[column.field]

src/components/Player/Pages/Counts/playerCountsColumns.jsx
```diff
@@ -17,5 +15,6 @@ export default [{
   displayName: strings.th_win,
   field: 'winPercent',
   sortFn: 1,
-  displayFn: (row, column, field) => <TablePercent val={Number(field.toFixed(1))} />,
+  displayFn: (...args) => `${args[4]}%`,
```
howardchung: probably better to leave named args.  Difficult to track down what args[4] is.

masad-frost:
I just wanna put it out there, now sortFn provides the value to the field, if it doesn't exist then we looks at row[column.field]. so displayFn essentially becomes displayFn(row, column, sortFn(row)).

No breaking changes, but at some point we should rename it to valueFn or something.

Just noticed the conflict. I'm on it

src/components/Heroes/columns.jsx
```diff
@@ -23,72 +22,72 @@ export default {
     displayName: strings.hero_pick_ban_rate,
     field: 'pickBanRatePro',
     sortFn: true,
-    displayFn: (row, col, field) => percentileDisplay(field, row.matchCountPro),
+    percentBars: row => decimalToCount(row.pickBanRatePro, row.matchCountPro),
```
howardchung: shouldn't percentbars take row.pickBanRatePro since it expects a decimal?  Or just make this sortFn and then pass `true`?

src/components/Player/Pages/Heroes/playerHeroesColumns.jsx
```diff
@@ -20,8 +19,8 @@ export const playerHeroesOverviewColumns = playerId => [{
   displayName: strings.th_win,
   tooltip: strings.tooltip_win_pct_as,
   field: 'win',
-  displayFn: row => <TablePercent val={getPercentWin(row.win, row.games)} />,
   sortFn: row => row.win / row.games,
+  percentBars: { getDivisor: row => row.games },
```
howardchung: should just be "true"?

src/components/Player/Pages/Peers/playerPeersColumns.jsx
```diff
@@ -30,8 +29,9 @@ const winsWith = [{
   displayName: strings.th_with_win,
   tooltip: strings.tooltip_win_pct_with,
   field: 'with_win',
-  displayFn: row => <TablePercent val={getPercentWin(row.with_win, row.with_games)} />,
   sortFn: row => row.with_win / row.with_games,
+  displayFn: (...args) => `${args[4]}%`,
```
howardchung: should be same as sortFn with rounding and percent?

src/components/Player/Pages/Peers/playerPeersColumns.jsx
```diff
@@ -30,8 +29,9 @@ const winsWith = [{
...
+  relativeBars: { getDivisor: row => row.with_games },
```
howardchung: should be "true"?

src/components/Player/Pages/Pros/playerProsColumns.jsx
```diff
@@ -26,8 +25,9 @@ export default playerId => [{
   displayName: strings.th_with_win,
   tooltip: strings.tooltip_win_pct_with,
   field: 'with_win',
-  displayFn: row => <TablePercent val={getPercentWin(row.with_win, row.with_games)} />,
   sortFn: row => row.with_win / row.with_games,
+  displayFn: (...args) => `${args[4]}%`,
```
howardchung: same comments as for playerPeersColumns

src/components/Heroes/columns.jsx
masad-frost: I thought we agreed that the function would pass on the small/alt value.

howardchung: Isn't it: 
* relativeBars takes a count value and divides by the max in the column
* percentBars takes a decimal value (between 0 and 10 and divides by 1 always

masad-frost: They both get their values from sortFn or row[column.field]. Their functions return the small value that appears next to the main value

PR #767, opened 2017-03-03
add pagination to heroes page

howardchung:


src/components/Heroes/index.jsx
```diff
@@ -76,7 +76,7 @@ class RequestLayer extends React.Component {
           className={styles.Heading}
           icon=""
         />
-        <Table data={data} columns={columns} />
+        <Table data={data} columns={columns} paginated pageLength={10} />
```
blukai: 10 is too short, maybe 20?

src/components/Heroes/index.jsx
```diff
@@ -89,7 +89,7 @@ class RequestLayer extends React.Component {
...
```
blukai: same here

PR #768, opened 2017-03-03
Add median and n to player histograms

### LINKED ISSUE #303: Show N and median for histograms

recurza:
fixes #303

### DIFF
--- a/src/components/Player/Pages/Histograms/Histograms.jsx
+++ b/src/components/Player/Pages/Histograms/Histograms.jsx
@@ -19,14 +19,28 @@ const getMedian = (columns, midpoint) => {
   return medianCol && medianCol.x;
 };
 
+const getMax = (columns) => {
+  let max = 0;
+  let maxColumn;
+  columns.forEach((col) => {
+    max = Math.max(max, col.games);
+    if (max === col.games) {
+      maxColumn = col.x;
+    }
+  });
+  return maxColumn;
+};
+
 const getSubtitleStats = (columns) => {
   const total = columns.reduce((sum, col) => (sum + col.games), 0);
-  const median = getMedian(columns, total / 2);
-  let str = `${strings.heading_total_matches}: ${total}`;
-  if (median !== undefined) {
-    str += `, ${strings.heading_median}: ${median}`;
-  }
-  return str;
+  const stats = [
+    [strings.heading_total_matches, total],
+    [strings.heading_median, getMedian(columns, total / 2)],
+    [strings.heading_max, getMax(columns)],
+  ];
+  return stats.filter(stat => stat[1] !== undefined)
+              .map(stat => `${stat[0]}: ${stat[1]}`)
+              .join(', ');
 };
 
 const histogramNames = dataColumns.filter(col => col !== 'win_rate');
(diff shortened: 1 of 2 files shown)


src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -5,23 +5,27 @@ import { playerHistogram } from 'reducers';
...
-import dataColumns from 'components/Player/Pages/matchDataColumns';
+import histogramNames from 'components/Player/Pages/matchDataColumns';
 import Container from 'components/Container';
 import { browserHistory } from 'react-router';
 import strings from 'lang';
 
-const histogramNames = dataColumns.filter(col => col !== 'win_rate');
-
-const Histogram = ({ routeParams, columns, playerId, error, loading }) => (
+const Histogram = ({ routeParams, columns, playerId, error, loading, histogramName }) => (
```
howardchung: don't we need the filter to avoid making a win rate histogram?

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -30,6 +34,24 @@ const getData = (props) => {
...
+    sum += col.games;
+    return (sum >= midpoint);
+  });
+  return medianCol && medianCol.x;
+}
+
+const getSubtitleStats = (columns) => {
+  let total = columns.reduce((sum, col) => (sum + col.games), 0);
+  let median = getMedian(columns, total/2);
+  let str = `${strings['heading_total_matches']}: ${total}`;
```
howardchung: you could probably express this in one template string with a ternary

howardchung:
lint issue?

src/components/Player/Pages/Histograms/Histograms.jsx
recurza: accident, not sure why it's deleted. fixing

howardchung:
Actually while this is still open could we add max as well?

recurza:
the getMax function feels a bit clumsy, suggestions?

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -19,14 +19,28 @@ const getMedian = (columns, midpoint) => {
   return medianCol && medianCol.x;
 };
 
+const getMax = (columns) => {
```
howardchung: you can do this with Math.max(...columns.map(col => col.x)).

We want the highest data point, not the bin with the most data values (that'd be the mode)

src/components/Player/Pages/Histograms/Histograms.jsx
```diff
@@ -19,14 +19,28 @@ const getMedian = (columns, midpoint) => {
...
 const getSubtitleStats = (columns) => {
   const total = columns.reduce((sum, col) => (sum + col.games), 0);
-  const median = getMedian(columns, total / 2);
-  let str = `${strings.heading_total_matches}: ${total}`;
-  if (median !== undefined) {
-    str += `, ${strings.heading_median}: ${median}`;
-  }
-  return str;
+  const stats = [
+    [strings.heading_total_matches, total],
```
howardchung: I suggest using objects rather than arrays here so indexing into them is more readable and doesn't break if things are reordered.

howardchung:
Actually we don't have the exact max since histograms bucket the data.  Maybe just roll back the last commit?

recurza:
wait, I messed up git pull --rebase

PR #774, opened 2017-03-05
show ranked/normal lobby

### LINKED ISSUE #773: Unranked Matches Now Shown As Ranked Matches
Unranked matches are currently being detected/displayed as ranked in match history, likely due to unranked using ban phase now.  Dotabuff doesn't have this issue.

howardchung:
fixes #773 

renamed "Ranked All Pick" to "All Draft"


howardchung:
The name in the protobuf definitions is "All Draft".  I used to call it Ranked All Pick because that's what people are familiar with, but now it's not exclusive to Ranked with the new changes

PR #776, opened 2017-03-05
Empty graph

### LINKED ISSUE #775: Handle Player/MMR page for players who don't expose it
Should either hide the tab, or more easily just show a message

howardchung:
fixes #775 

Shows an empty graph instead of nothing


howardchung:
Whoops let me remove the first commit

PR #787, opened 2017-03-07
Refactor Announce to follow component pattern.

covertPZ:
Absolutely position on home page to not push content down.

### DIFF
--- a/src/components/Announce/Announce.css
+++ b/src/components/Announce/Announce.css
@@ -62,3 +62,13 @@
     flex-shrink: 0;
   }
 }
+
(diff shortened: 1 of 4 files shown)


howardchung:
Actually would it be bad to just use absolute position everywhere?

covertPZ:
If we use absolute position everywhere it will cover the VP ad on the other pages. Not sure if that's okay, but it may also make it seem a bit like click bait if someone double clicks the dismiss button.

howardchung:
OK let's try this out first then

PR #788, opened 2017-03-07
reshuffled vision tab components

howardchung:
Rearranged components so that they fit better on the screen.  Prior to this, the 100% width vision map was not fitting on the screen (most screens are wider than they are tall), and scrolling back and forth to use the scrub bar/filter was annoying.

Map is now 50% width with the controls (scrub/filter) on the right 50%.

Put use counts/log below in another 50/50 split.

Added pagination to ward log.

Demo:
http://hchung.westus.cloudapp.azure.com:8080/matches/3037756062/vision


dartharaknus:
I think this is the wrong approach. We definitely need a bigger map. In the current map size, you can't see details much. One way of solving the filter problem would be a floating fixed div. I am vehemently against reducing the size of the map.

dartharaknus:
And please no pagination to the ward log. This is just one game. It is important to see all the wards without clicking anything else like page numbers.

src/components/Match/Vision/VisionLog.jsx
```diff
@@ -58,7 +60,8 @@ const generateData = match => (log) => {
 
 const VisionLog = ({ match, wards }) => (
   <div className={styles.visionLog}>
```
blukai: remove this class

image

blukai:
I fully like this pr, 100% map was really annoying.

howardchung:
The 100% map doesn't fit on my screen though, which makes the controls hard to use.

dartharaknus:
What about setting the map size to 100% of the height of your screen and moving the filters to the side? This 50% map is not useful to anyone as it's a bit small. So, if scrolling is an issue, let's make it fit the height instead of cutting it in half.

src/components/Match/matchColumns.jsx
```diff
@@ -770,7 +770,7 @@ export const teamfightColumns = [
     displayFn: inflictorRow(items),
   },
 ];
-
+/*
 const purchaseObserverColumn = {
```
dartharaknus: Why are these commented out?

howardchung: To make the tables fit better I removed the purchase columns (since this page is more about actual usage).

Purchase data remains on the purchase page.

howardchung:
Do you know the proper CSS for getting to 100% height?  I tried height: 100% but it doesn't seem to be working.

On my screen the current 50% width is also taking up about 90% of the height (once browser, taskbar are included).  My screen resolution is 2560x1440 -- what resolution are you using that it's unusable at 50%?

dartharaknus:
I am not saying it's unusable. I am just saying the smaller map is not good for detail oriented users (like me). What if we add an option and let the user decide?

Hmm.. Why do we need to split the below tables into 50%? They look good enough as they are now. Regarding moving them to purchases page, we should have them at both places. They are both purchases and vision related. Users should click once on vision and see everything vision related.

blukai:
> Regarding moving them to purchases page, we should have them at both places. They are both purchases and vision related. Users should click once on vision and see everything vision related.

I agree

> I am just saying the smaller map is not good for detail oriented users (like me). What if we add an option and let the user decide?

Probably the best decision would be to add switcher - http://www.material-ui.com/#/components/toggle

howardchung:
I'd like to push this as-is, and we can think about solutions for larger maps later?

PR #789, opened 2017-03-08
return top 10 matches for records

howardchung:
Records page:

* Instead of just returning a single row for each record, this now uses a ButtonGarden selector and returns the top 10 matches for each record.

* This allows users to discover more matches that are high in a particular stat, and helps mitigate issues where a record is affected by a fluky data point.

* Uses the /matches endpoint + sort field, so we could retire the /records endpoint from core.

### DIFF
--- a/src/components/Player/Pages/Records/Records.jsx
+++ b/src/components/Player/Pages/Records/Records.jsx
@@ -12,7 +12,8 @@ import dataColumns from 'components/Player/Pages/matchDataColumns';
 import ButtonGarden from 'components/ButtonGarden';
 import playerRecordsColumns from './playerRecordsColumns';
 
-const recordsColumns = dataColumns.filter(col => col !== 'win_rate');
+const excludedColumns = ['win_rate', 'level'];
+const recordsColumns = dataColumns.filter(col => !excludedColumns.includes(col));
 
 const Records = ({ routeParams, data, error, loading, playerId }) => {
   const selected = routeParams.subInfo || recordsColumns[0];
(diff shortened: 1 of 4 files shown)


dartharaknus:
Any screenshots?

blukai:
@dartharaknus 

image

blukai:
@howardchung I think record called **Level** is definitely useless and should not be

covertPZ:
How difficult would it be to maintain the same list we have now as "Top" tab in this layout as the default view? Would we be able to just pull it as a subset from the data so we can still retire the /records endpoint? 

Only thing I'm not a huge fan of with this new layout is that I feel like I wouldn't explore all the data since I can't get a quick benchmark for each category.

howardchung:
We'd have to make a separate API call for each record--so it would be pretty expensive since there like 20 fields.

src/components/Player/Pages/Records/Records.jsx
```diff
@@ -3,20 +3,36 @@ import { connect } from 'react-redux';
...
+  return (<div style={{ fontSize: 10 }}>
+    <ButtonGarden
+      onClick={(buttonName) => {
+        browserHistory.push(`/players/${playerId}/records/${buttonName}${window.location.search}`);
+      }}
+      buttonNames={recordsColumns}
+      selectedButton={selected}
+    />
+    <Container title={strings.heading_records} error={error} loading={loading}>
+      <Table columns={playerRecordsColumns.concat({ displayName: strings[`th_${selected}`], field: selected })} data={data} />
```
masad-frost: Some records don't have headers. Namely Throw, Comeback, Stomp, Loss. I think it's safe to just add `|| 'strings.th_record'` to display name

masad-frost: Also, my stuns records has values with 6 decimal places. If you can add a displayFn to solve that. If you don't want to add it to everything, you can use `Math.round(x * 100) / 100` instead of the usual `toFixed` to maintain whole numbers.

PR #795, opened 2017-03-10
autocomplete support for explorer

howardchung:
Still WIP, could probably use further improvements.

Autocompletes:
* SQL keywords (SELECT, JOIN, etc.)
* table names (matches, player_matches, etc.)
* column names (kills, deaths, assists, etc.)
* postgresql functions (to_timestamp, etc.)

Should we:
* Do longer phrases?  Like `JOIN player_matches USING(match_id)`
* Add more function names?
* Determine what set of completions to show based on the editor state?

### DIFF
--- a/src/components/Explorer/Explorer.jsx
+++ b/src/components/Explorer/Explorer.jsx
@@ -41,6 +41,7 @@ import querystring from 'querystring';
 import queryTemplate from './queryTemplate';
 import ExplorerFormField from './ExplorerFormField';
 import fields from './fields';
+import autocomplete from './autocomplete';
 import styles from './Explorer.css';
 
 // TODO omnibox search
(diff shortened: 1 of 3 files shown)


masad-frost:
> Do longer phrases? Like JOIN player_matches USING(match_id)
I think that's a bit too much.

> Add more function names?
Sure, why not.

> Determine what set of completions to show based on the editor state?
Would be really cool. Could be implemented using a suffix tree. 

I think the main thing is we should add a visual representation for the schema; bonus if it's interactive (i.e click on column/table it adds it to the query, for now at least have a link to the schema.

howardchung:
Maybe something like:
https://data.stackexchange.com/stackoverflow/query/new

PR #796, opened 2017-03-11
Implement counts record on Overview Tab.

### LINKED ISSUE #771: counts summary section on overview
I'm thinking we can display some of the counts data (abbreviated) on the Overview page.  Category data like:
* radiant/dire and winrate on each
* most played gamemode and winrate
* winrate and matches played in most recent patch
* top 3 regions and winrate

We can make the counts API call (like we do for peers, heroes, and matches) and then just display a subset of the data.

covertPZ:
fixes #771 
Increased number of heroes in order to make the page alignment look cleaner.
Logic may not be the best, so definitely looking for some review/optimization. (had some difficulty understanding the transforms that were occurring to the data and why certain fields weren't accessible at certain stages in processing)
Container title probably deserves a better name, or at least it's own variable if we decide to change it in the future.

### DIFF
--- a/src/components/Player/Pages/Counts/playerCountsColumns.jsx
+++ b/src/components/Player/Pages/Counts/playerCountsColumns.jsx
@@ -18,3 +18,20 @@ export default [{
   sortFn: row => row.winPercent / 100, // percentBars expects decimal value
   percentBars: true,
 }];
+
+export const playerCountsOverviewColumns = [{
+  displayName: strings.th_category,
+  field: 'category',
+  sortFn: true,
+  displayFn: transformations.category,
+}, {
+  displayName: strings.th_matches,
+  field: 'matches',
+  sortFn: true,
+  displayFn: transformations.matches,
+}, {
+  displayName: strings.th_win,
+  field: 'winPercent',
+  sortFn: row => row.winPercent / 100, // percentBars expects decimal value
+  percentBars: true,
+}];
--- a/src/components/Player/Pages/Overview/Overview.jsx
+++ b/src/components/Player/Pages/Overview/Overview.jsx
@@ -2,12 +2,14 @@ import React from 'react';
 import { connect } from 'react-redux';
 import strings from 'lang';
 import {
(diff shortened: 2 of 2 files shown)


howardchung:
tip: If you make your branch on this repo it's easier for the rest of us to clone without having to add your repo to remote :)

howardchung:
General comments:
* Currently seems a bit buried (user has to scroll down a lot).  Could we maybe move to a three column layout?
* Probably should separate the sections (radiant/dire, patch, region, mode)
* Context links to the matches filter page when number is clicked on (like heroes/players?)  More generally this could be done for the rest of the counts

covertPZ:
Ah, okay. I'll create the branches on the main repo from now on. Will make my workflow easier too :)

The three column layout makes things look super busy and will only get more filled if the counts get split into individual tables. This layout is also pushing it to show content without horizontal scrolling (below 1540px the counts would need to be pushed down) 
image

Maybe we want to represent this data similar to the "Summary" section?

blukai:
> Currently seems a bit buried (user has to scroll down a lot). Could we maybe move to a three column layout?

I think that 3 columns is bad

> - Context links to the matches filter page when number is clicked on (like heroes/players?) More generally this could be done for the rest of the counts
> - Instead of a table, maybe something a little more eye-catching? c3 has gauge charts: http://c3js.org/samples/chart_gauge.html or maybe donut charts: http://c3js.org/samples/chart_donut.html

👍

howardchung:
Yeah, maybe another row under the summary section would be good

covertPZ:
Still working out some parts of this, but before I go much further I'd like some feedback.
image

howardchung:
Maybe we could shorten the last 20 games summary and put them both on one line?  IMO having two additional rows on top is pushing down the "primary" content a little.

howardchung:
I wonder if it would make sense to put this section under the existing content, rather than over.

howardchung:
replaced by #1635

PR #807, opened 2017-03-15
Performances improvement

### LINKED ISSUE #522: Should be able to see a larger lane map
It's very small currently.

blukai:
fixes #522 

image

**TODO:**
- [ ] Responsive map


howardchung:
Hmm, I actually kind of liked being able to see multiple players at once (easier to do direct comparisons)

howardchung:
Actually I think what we can do with this is make a separate tab called laning and just display a subset of data like stats@10.  We can keep the performance table as is (but remove the small map)

PR #843, opened 2017-03-26
improvement on docker

yangli1990:


yangli1990:
This should make better use of the docker cache

PR #845, opened 2017-03-27
Finish simplified Chinese translation.

kxxoling:
Finish simplified Chinese translation; fix some little typo; fix punctuation.


howardchung:
The recommended way to translate is to do it through http://translate.opendota.com .  This probably works too, though.

PR #852, opened 2017-03-29
Story Tab: fixed firstblood bug and added draw for laning

### LINKED ISSUE #837: Match story contains "Unrecognized Hero"
https://www.opendota.com/matches/3071310248/story

As you can see, on the first line, the match story says "First blood was drawn when Keeper of the Light killed Unrecognized Hero at 1:23." I'm fairly sure it was Slardar that died. That's the only place where the hero isn't recognized in the story, though, everywhere else it displays all the names of every hero correctly. I haven't tried to re-parse because I figured it would be a good idea to let devs know about this bug and I didn't want the example to disappear.

mdiller:
fixes #837


mdiller:
fixes #837

src/components/Match/MatchStory.jsx
```diff
@@ -203,14 +203,16 @@ const localizedLane = {
...
+const getLaneScore = players => (Math.max(...players.map(player => player.gold_t[10] || 0)) || 0);
+const laneScoreDraw = 350;
 
 class LaneStory {
   constructor(match, lane) {
     this.radiant_players = match.players.filter(player => player.lane === parseInt(lane, 10) && player.isRadiant && (!player.is_roaming));
     this.dire_players = match.players.filter(player => player.lane === parseInt(lane, 10) && !player.isRadiant && (!player.is_roaming));
     this.lane = lane;
     this.winning_team = getLaneScore(this.radiant_players) > getLaneScore(this.dire_players);
+    this.is_draw = getLaneScore(this.radiant_players) - getLaneScore(this.dire_players) <= laneScoreDraw;
```
howardchung: Don't you need to use absolute value here?

PR #856, opened 2017-04-01
Make burger menu links fill available space

recurza:


howardchung:
do you have a screenshot of what this looks like?

recurza:
The appearance didn't change. It's so that users (on mobile) can click anywhere on the menu item to go to the page, instead of having to click directly on the text

PR #861, opened 2017-04-02
adds proptypes to some components

howardchung:
Questions:
* We seem to have several different ways of referencing the react proptypes.  Is one recommended?  I mostly used `React.PropTypes.string` but that's kind of scary because if I mistype `React.proptypes.string`, the lint will pass but the component will crash on run when it tries to access a nested nonexistent property.

### DIFF
--- a/src/components/Table/SortedTable/withSort.jsx
+++ b/src/components/Table/SortedTable/withSort.jsx
@@ -23,6 +23,13 @@ export default (Table, id = uuid.v4()) => {
     return <span />;
   };
 
+  SortedTable.propTypes = {
+    data: React.PropTypes.array,
+    sortState: React.PropTypes.string,
+    sortField: React.PropTypes.string,
+    sortTable: React.PropTypes.func,
+  };
+
   const mapStateToProps = (state, { data }) => ({
     sortState: table.getSortState(state, id),
     sortField: table.getSortField(state, id),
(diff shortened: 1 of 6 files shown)


covertPZ:
I think we should be fine to do it with the `React.PropTypes.string`. If the variable is referenced incorrectly and the component crashes that should be pretty obvious before the push and even if it does get pushed it should be a quick fix. There's other reasons a component could crash that the lint wouldn't pick up so I don't think we should be too worried with that as a reason.

gu3st:
Why not add the proptypes/defaultProps as a static get method on the class?    It's equivalent, but then it's encapsulated in the class itself.

howardchung:
@gu3st example?

gu3st:
Inside the class, you should be able to do something like this:

`
    static get propTypes(){ 
        return { <prop type object contents here> }
    }
`

It's defining a static (bound to class, not instance) getter method for propTypes, which accomplishes the same as ClassName.propTypes

It's worth trying out, I find it makes my classes nicer, without having propTypes/defaultProps just kinda.. hanging out outside of my Class

howardchung:
I just followed the existing convention, not sure if @mike-robertson had a reason for doing it this way.

howardchung:
React has moved proptypes to a separate package.

https://www.npmjs.com/package/prop-types

We should use this going forward, and migrate all old components.

mike-robertson:
Yea, I only did it that way because of conventions I have seen. I don't think there is any significant difference except preference. Encapsulated is probably more readable.

howardchung:
We'll still need to use the assignment style for pure function components, so maybe we should just be consistent and do that everywhere.

PR #866, opened 2017-04-06
Added Assistant search

sheldonlai:
- added a function for user to say "search [player name]" and it will direct to the search page with the query.


howardchung:
So, this is actually an April Fool's joke and wasn't intended to be a feature.  I can merge this, but we'll probably be removing it in the next couple of days.

sheldonlai:
Oh okay. It was my first time seeing the site today,  thought it was just an incomplete page.

PR #874, opened 2017-04-09
implement mass-select/deselect for vision checkboxes

### LINKED ISSUE #867: Allow single selection on vision
Allow single click to select only one on vision page.

Maybe we can implement this just by deselecting all of the boxes within a team when the box in the corner is checked.  Then it's three clicks, one to deselect everything, and then one more click to turn on just one user's wards.

@dartharaknus ?

howardchung:
fixes #867

### DIFF
--- a/src/components/Match/Vision/VisionFilter.jsx
+++ b/src/components/Match/Vision/VisionFilter.jsx
@@ -19,30 +19,30 @@ const data = [
   },
 ];
 
-export default class VisionFilter extends React.PureComponent {
-  handleCheckGenerator(player, type) {
-    const parent = this.props.parent;
-
-    return event => parent.setPlayer(player, type, event.target.checked);
-  }
-
-  handleCheckGeneratorTeam(index) {
-    const parent = this.props.parent;
-
-    return event => parent.setTeam(index === 0 ? 'radiant' : 'dire', event.target.checked);
-  }
-
+export default class VisionFilter extends React.Component {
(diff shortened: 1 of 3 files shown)


src/components/Match/Vision/VisionFilter.jsx
```diff
@@ -19,30 +19,30 @@ const data = [
...
-    const parent = this.props.parent;
-
-    return event => parent.setPlayer(player, type, event.target.checked);
-  }
-
-  handleCheckGeneratorTeam(index) {
-    const parent = this.props.parent;
-
-    return event => parent.setTeam(index === 0 ? 'radiant' : 'dire', event.target.checked);
-  }
```
dartharaknus: Please don't remove this code. You can still use it.

howardchung: These functions don't seem to be reused, and I find the new version more readable.  Is there a reason to have them defined separately and returning functions?

Behavior is the same.

dartharaknus: I find that having functions like this makes it more readable and maintainable.

howardchung: My concern is that splitting off the function like this causes the developer to have to jump around the code to find the values that are getting passed.  In the new version, the variables playerNumber, row, and  event are all in the same contiguous block.

If the function needed to be reused, then we'd have to split it off, but that currently isn't the case.

PR #891, opened 2017-04-14
Story Tab Updates

### LINKED ISSUE #886: Insert time markers at intervals in story
We should put in something like a subheading and maybe an `hr` element at intervals in the story (maybe every 10 minutes?)

This will help with linking story events to specific match times, since most story events right now don't have the time context (besides the laning summary and the first item over 4000 events).

@mdiller ?

mdiller:
- Rapier Purchases
- Time Markers
fixes #886
This seems to be the kind of thing we're looking for. Not sure if this is the right style / wording.

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -13,6 +13,16 @@ import styles from './Match.css';
 
 const heroesArr = jsonFn(heroes);
 
+const itemQualityColor = {
+  rare: '#1a87f9',
+  artifact: '#e29b01',
+  secret_shop: '#ffffff',
+  consumable: '#1d80e7',
+  common: '#2bab01',
+  epic: '#b812f9',
+  component: '#ffffff',
+};
+
 // can be used in conjunction with is_radiant
 const TEAM = {
   radiant: true,
@@ -69,7 +79,11 @@ const PlayerSpan = (player) => {
 
 // Modified version of PlayerThumb
 const ItemSpan = item => (
-  <span key={`item_${item}`} className={styles.storySpan}>
+  <span
+    key={`item_${item}`}
+    className={styles.storySpan}
+    style={{ color: itemQualityColor[items[item].qual] }}
+  >
     <img
       width="26px"
       src={items[item]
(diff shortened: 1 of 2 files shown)


mdiller:
The last commit adds item quality to color conversion. Should this be added in a different place to make it more globally accessible, instead of just in the story tab?
By the way, colors were retrieved from https://www.dota2.com/public/css/heropedia.css

src/components/Match/MatchStory.jsx
```diff
@@ -13,6 +13,16 @@ import styles from './Match.css';
...
 
+const itemQualityColor = {
+  rare: '#1a87f9',
+  artifact: '#e29b01',
+  secret_shop: '#ffffff',
+  consumable: '#1d80e7',
+  common: '#2bab01',
+  epic: '#b812f9',
+  component: '#ffffff',
+};
```
mdiller: Should this be added in a different place to make it more globally accessible, instead of just in the story tab?
I got these colors from https://www.dota2.com/public/css/heropedia.css

howardchung: If you think it might be useful to other developers/that this might be reused somewhere, you can add it as a json file in https://github.com/odota/dotaconstants

If you do that, then I'll publish a new version of dotaconstants and you can do `npm update` to update it locally, then reference this blob like we do heroes, items, etc.

src/components/Match/MatchStory.jsx
```diff
@@ -562,6 +625,21 @@ const generateStory = (match) => {
     }
   }
 
+  // Remove any unneeded Time Markers
```
mdiller: Yeah, but what that part does is remove any markers that come right after another marker. That way if nothing happens between minute 30 and minute 40, instead of this:
**20min Marker**
event1
event2
**30min Marker**
**40min Marker**
event3
event4
event5

you get this:
**20min Marker**
event1
event2
**40min Marker**
event3
event4
event5

howardchung: This might be a more elegant solution:
http://stackoverflow.com/questions/30716829/how-to-remove-repeated-entries-from-an-array-while-preserving-non-consecutive-du

You can just check if the element before is a TimeMarketEvent

mdiller: cool, ill do that. thanks

PR #907, opened 2017-04-18
Story updates

### LINKED ISSUE #901: "intro" line for match stories
Lead into the story with a line like:
`On {date}, two teams decided to play a {game mode} game of Dota 2 in {region}`

mdiller:
fixes #901 
Added an intro and moved a couple of things around

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -5,6 +5,7 @@ import {
   formatSeconds,
   jsonFn,
   transformations,
+  formatTemplate,
 } from 'utility';
 import { IconRadiant, IconDire } from 'components/Icons';
 import heroes from 'dotaconstants/build/heroes.json';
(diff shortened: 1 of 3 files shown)


src/components/Match/MatchStory.jsx
```diff
@@ -152,6 +140,29 @@ class StoryEvent {
...
 }
 
+class IntroEvent extends StoryEvent {
+  constructor(match) {
+    super(-90);
+    this.game_mode = match.game_mode;
+    this.region = match.region;
+    this.date = new Date(match.start_time * 1000);
+  }
+  get localizedRegion() {
```
howardchung: If we want to use a nicer looking string for region, you may want to consider doing this in odota/dotaconstants instead, then it would show up the same way e.g. on match pages.

mdiller: Where would we do it there? Just by changing the json file?

howardchung: By updating the updateconstants file to write the prettified text string.  Putting strings in the language files is another option, something like `region_1`, etc.

howardchung:
The generalizing of the region string can be done separately, I'll merge this now

PR #934, opened 2017-05-01
gg gold and track

evaldobratti:
new column in performance tab

I still need to change game log, to show which deaths were tracked

tell what you guys think about it

### DIFF
--- a/src/components/Match/MatchLog.jsx
+++ b/src/components/Match/MatchLog.jsx
@@ -96,11 +96,7 @@ const logColumns = [heroTdColumn, {
     switch (row.type) {
       case 'kills': {
         const hero = heroNames[row.detail] || {};
-        const comp = [<img key="hero" src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />];
-        if (row.tracked_death) {
-          comp.push('Tracked');
-        }
-        return <div>{ comp }</div>;
+        return <img src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />;
       }
       case 'runes': {
         const runeType = row.detail;
--- a/src/components/Match/MatchLog.jsx
+++ b/src/components/Match/MatchLog.jsx
@@ -96,7 +96,11 @@ const logColumns = [heroTdColumn, {
     switch (row.type) {
       case 'kills': {
         const hero = heroNames[row.detail] || {};
-        return <img src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />;
+        const comp = [<img key="hero" src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />];
+        if (row.tracked_death) {
+          comp.push('Tracked');
+        }
+        return <div>{ comp }</div>;
       }
       case 'runes': {
         const runeType = row.detail;
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -466,6 +466,31 @@ export const performanceColumns = [
       }
       return <div />;
     },
+  }, {
+    displayName: strings.th_others,
+    tooltip: strings.tooltip_others,
+    field: 'performance_others',
+    sortFn: true,
+    displayFn: (row, col, field) => {
+      const comp = [];
+      if (field) {
+        if (field.tracked_deaths) {
+          comp.push(<div>
+            { inflictorWithValue('bounty_hunter_track', abbreviateNumber(field.tracked_deaths), '',
+              `${field.tracked_deaths} ${strings.tooltip_others_total_gold} 
+              ${field.track_gold} ${strings.tooltip_others_greevils_gold}`) }
+          </div>);
+        }
+        if (field.greevils_greed_gold) {
+          comp.push(<div>
+            { inflictorWithValue('alchemist_goblins_greed', abbreviateNumber(field.greevils_greed_gold), '',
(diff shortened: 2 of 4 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -466,6 +466,31 @@ export const performanceColumns = [
...
+    tooltip: strings.tooltip_others,
+    field: 'performance_others',
+    sortFn: true,
+    displayFn: (row, col, field) => {
+      const comp = [];
+      if (field) {
+        if (field.tracked_deaths) {
+          comp.push(<div>
+            { inflictorWithValue('bounty_hunter_track', abbreviateNumber(field.tracked_deaths), '',
+              `${field.tracked_deaths} ${strings.tooltip_others_total_gold}
```
howardchung: Maybe put these two elements in an array and do `.join('\n')` so it's more apparent

src/components/Match/matchColumns.jsx
```diff
@@ -466,6 +466,31 @@ export const performanceColumns = [
...
+              ${field.track_gold} ${strings.tooltip_others_greevils_gold}`) }
+          </div>);
+        }
+        if (field.greevils_greed_gold) {
+          comp.push(<div>
+            { inflictorWithValue('alchemist_goblins_greed', abbreviateNumber(field.greevils_greed_gold), '',
+              `${field.greevils_greed_gold} ${strings.tooltip_others_greevils_gold}`) }
+          </div>);
+        }
+        return <div> { comp } </div>;
```
howardchung: no need for the extra spaces in the div

src/lang/en-US.json
```diff
@@ -702,6 +702,7 @@
   "th_permanent_buffs": "Buffs",
   "th_winner": "Winner",
   "th_played_with": "My Record With",
+  "th_others": "Others",
```
howardchung: Just "Other" is better I think

src/lang/en-US.json
```diff
@@ -841,6 +842,8 @@
   "tooltip_autoattack_other": "Auto Attack/Other",
   "tooltip_estimated_mmr": "MMR estimate based on the mean visible MMR of the recent matches played by this user",
   "tooltip_backpack": "Backpack",
+  "tooltip_others": "",
```
howardchung: what is this empty string for?

src/lang/en-US.json
```diff
@@ -862,5 +865,9 @@
   "vision_all_time": "All time",
   "vision_placed_observer": "placed Observer at",
   "vision_placed_sentry": "placed Sentry at",
-  "vision_ward_log": "Ward Log"
+  "vision_ward_log": "Ward Log",
+
+  "tooltip_others_tracked_deaths": "tracked deaths",
+  "tooltip_others_total_gold": "gold earnt from Track",
```
howardchung: I think "earned" instead of "earnt"

src/components/Match/MatchLog.jsx
```diff
@@ -96,7 +96,11 @@ const logColumns = [heroTdColumn, {
     switch (row.type) {
       case 'kills': {
         const hero = heroNames[row.detail] || {};
-        return <img src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />;
+        const comp = [<img key="hero" src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />];
+        if (row.tracked_death) {
+          comp.push('Tracked');
+        }
```
evaldobratti: I don't know how to improve it, something like [this](http://imgur.com/kZLjsHb) maybe (that I have no idea how to do it), or remove it from log at all.

src/components/Match/MatchLog.jsx
```diff
@@ -96,7 +96,11 @@ const logColumns = [heroTdColumn, {
     switch (row.type) {
       case 'kills': {
         const hero = heroNames[row.detail] || {};
-        return <img src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />;
+        const comp = [<img key="hero" src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />];
+        if (row.tracked_death) {
+          comp.push('Tracked');
```
howardchung: move to string file?

src/components/Match/MatchLog.jsx
```diff
@@ -96,7 +96,11 @@ const logColumns = [heroTdColumn, {
     switch (row.type) {
       case 'kills': {
         const hero = heroNames[row.detail] || {};
-        return <img src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />;
+        const comp = [<img key="hero" src={`${API_HOST}${hero.img}`} className={styles.imgSmall} role="presentation" />];
```
howardchung: Actually, I'mt not sure what you're trying to do here.  If you're trying to show some string if row.tracked_death is true, you should be able to draw another element in the div.  Having an array of an img element and a string seems kind of inconsistent

src/lang/en-US.json
```diff
@@ -862,5 +864,9 @@
   "vision_all_time": "All time",
   "vision_placed_observer": "placed Observer at",
   "vision_placed_sentry": "placed Sentry at",
-  "vision_ward_log": "Ward Log"
+  "vision_ward_log": "Ward Log",
+
+  "tooltip_others_tracked_deaths": "tracked deaths",
```
howardchung: move to the tooltip section with the other tooltip_ entries

evaldobratti:
I'll try to think in something better to match log tracked kills.
I think this PR is done.

PR #935, opened 2017-05-01
Lang file values from d2vpkr

mdiller:
I made a script to automatically get some of the values (`game_mode_n`, `rune_n`, `npc_dota_x`, and the new `region_n`) from d2vpkr's lang files, and save them to our lang files.
This script is probably a one time use, so I'll probably remove it from this PR before we're done. If you think we should keep it in, let me know where it should be.


src/lang/bg-BG.json
```diff
@@ -174,20 +174,20 @@
   "game_mode_6":"Intro",
   "game_mode_7":"Diretide",
   "game_mode_8":"Reverse Captains Mode",
-  "game_mode_9":"Greeviling",
-  "game_mode_10":"Tutorial",
+  "game_mode_9":"The Greeviling",
```
howardchung: You might want to backport these string updates to the English file, otherwise English will not be consistent with the translated files.

howardchung:
You can create a `dev` directory in the root and put the script there (like what we do in core)

src/lang/en-US.json
```diff
@@ -187,20 +187,20 @@
...
   "game_mode_16": "Captains Draft",
   "game_mode_17": "Balanced Draft",
   "game_mode_18": "Ability Draft",
   "game_mode_19": "Event",
-  "game_mode_20": "All Random Death Match",
-  "game_mode_21": "1v1 Mid",
-  "game_mode_22": "All Draft",
+  "game_mode_20": "All Random Deathmatch",
+  "game_mode_21": "1v1 Solo Mid",
+  "game_mode_22": "Ranked All Pick",
```
howardchung: I think we should keep this as All Draft, since even unranked all pick matches use this mode now.

mdiller: sounds good

PR #944, opened 2017-05-06
experiment with card layout

### LINKED ISSUE #909: Border/drop shadow on Container element
Suggestion:

Perhaps we should try having the Container element (which we use quite a bit) give the enclosed elements a border and a drop shadow to make them stand out more.  The result I'm envisioning would look kind of  like Material Cards: 
http://www.material-ui.com/#/components/card

@blukai

howardchung:
fixes #909 

Playing with borders/box shadow to more clearly delineate sections.

Feedback/advice very much appreciated!


howardchung:
I removed padding and the border, left in background color and the box shadow.  Still could use some improvements, I think.

PR #946, opened 2017-05-06
predictions added to match story

### LINKED ISSUE #911: New story line: Who predicted they would win the game?
@howardchung mentioned in the discord chat that the replays do include "player has predicted!" data. It'd be neat to see that added to the story view as well.

Not sure if it's parsed and available in the JSON yet.

Would be super cool if streak data was available too. But that's not important. Not sure how universal it is, since charms don't really have a "streak" per se.

Would also be amusing to mention at the end of the story to mention how the predictions panned out.

villith:
fixes #911

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -226,7 +226,7 @@ class PredictionEvent extends StoryEvent {
   }
   format() {
     return formatTemplate(strings.story_predicted_victory,
-      { players: formatList(this.players.map(PlayerSpan), strings.story_predicted_victory_empty), team: TeamSpan(this.team) }
+      { players: formatList(this.players.map(PlayerSpan), strings.story_predicted_victory_empty), team: TeamSpan(this.team), }
     );
   }
 }
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -225,9 +225,10 @@ class PredictionEvent extends StoryEvent {
     }
   }
   format() {
-    return formatTemplate(strings.story_predicted_victory,
-      { players: formatList(this.players.map(PlayerSpan), strings.story_predicted_victory_empty), team: TeamSpan(this.team), }
-    );
+    return formatTemplate(strings.story_predicted_victory, {
+      players: formatList(this.players.map(PlayerSpan), strings.story_predicted_victory_empty),
+      team: TeamSpan(this.team),
+    });
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Match/MatchStory.jsx
```diff
@@ -540,6 +554,14 @@ const generateStory = (match) => {
   // Intro
   events.push(new IntroEvent(match));
 
+  // Prediction
+  match.players.forEach((player) => {
+    if (player.pred_vict === true) {
+      const order = player.player_slot < 5 ? player.player_slot - 89 : player.player_slot - 212;
```
howardchung: what are 89 and 212?

villith: StoryEvents use the time of an event as their unique key, but predictions don't have a time stamp

So to get a unique key per instance of PredictionEvent I normalized player_slot to the -89-80 range using 89 for radiant (0-4) slots and 212 for dire (127-132) slots

howardchung: Why not just do `player_slot % 128`?

howardchung: oh, all 10 have to be unique.  In that case why not just use player_slot directly?  They'll be unique, and I don't think there's a requirement for the range to be contiguous.

villith: if i'm not mistaken it is literally the time of the event -- for example IntroEvent is set to -90, when players load into the game

villith:
now displays as 

`player1, player2, player3, player4, and player5 predicted Radiant would win`

it will only display if at least one prediction was made

src/components/Match/MatchStory.jsx
```diff
@@ -540,6 +557,18 @@ const generateStory = (match) => {
...
+  // Prediction
+  let predExists = false;
+  match.players.forEach((player) => {
+    if (player.pred_vict === true) {
+      predExists = true;
+    }
+  });
+  if (predExists === true) {
+    events.push(new PredictionEvent(match, -89));
+    events.push(new PredictionEvent(match, -88));
```
mdiller: Looks like this will add a prediction event even if nobody on that team predicted, as long as the other team did. Example match: 3159888768.

villith: that is intentional

onepredictionexample

mdiller: Ohh, okay.

howardchung:
LGTM, looks like just a lint issue to fix

PR #949, opened 2017-05-08
Fix item buy time

### LINKED ISSUE #437: Multiple purchases of the same item show wrong timings on overview
https://www.opendota.com/matches/2762021523/overview

Both Daedalus are marked at 26:08. The Purchases tab correctly shows the second one being purchased at 53:02.

pinkiesky:
fixes #437

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -1,5 +1,6 @@
 /* global API_HOST */
 import React from 'react';
+import findLast from 'lodash/findLast';
 import heroes from 'dotaconstants/build/heroes.json';
 import items from 'dotaconstants/build/items.json';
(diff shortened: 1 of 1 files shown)


howardchung:
Do you have a sample match ID that you tested this against?

pinkiesky:
https://www.opendota.com/matches/2762021523/overview
[Before](https://ibb.co/mYBHJ5) and [with fix](https://ibb.co/dJjRBQ)

pinkiesky:
Fix also affects the buy time of all items in general. 
I think the use of the `purchase_log` is much more reasonable than the `first_purchase_time`.
Is there any reason for use `first_purchase_time`? Perhaps I do not see the whole picture as a whole?

howardchung:
It's just simpler since it doesn't require iterating over the purchase log.

Issue #759, opened 2017-02-27
Roaming heroes


masad-frost:
Use match/player/is_roaming implemented in odota/core#1414

Obviously should be used in the match performance tab, maybe add a tooltip on the `Roaming` field that displays the lane they were mostly on (match/player/lane_role).

Not sure where or how else can we use this.

howardchung:
This should be pretty easy

maschwenk:
Something I'd be interested in picking up. This project is awesome.

howardchung:
Sure!  Clone the UI and get it running.  The API response already contains this data, so this should just be as simple as adding `(Roaming)` to the column in the performances tab if the `is_roaming` property is true.

PR #954, opened 2017-05-12
Add roaming info to Lane col in Performances tab

hannaliebl:
fixes https://github.com/odota/ui/issues/759

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -364,12 +364,10 @@ export const performanceColumns = [
     tooltip: strings.tooltip_lane,
     field: 'lane_role',
     sortFn: true,
-    displayFn: (row, col, field) => (row.is_roaming ?
-      <div>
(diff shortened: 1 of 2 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -363,7 +364,12 @@ export const performanceColumns = [
     tooltip: strings.tooltip_lane,
     field: 'lane_role',
     sortFn: true,
-    displayFn: (row, col, field) => strings[`lane_role_${field}`],
+    displayFn: (row, col, field) => (row.is_roaming ?
+      <div>
+        <span>{strings[`lane_role_${field}`]}</span>
+        <span className={subtextStyle.subText}>Roaming</span>
```
howardchung: You could render just this line on `row.is_roaming` which removes the duplication of the lane_role line?

howardchung: Also please put the `Roaming` string in the en-US lang file.  The rest of the files will be automatically filled later.

PR #970, opened 2017-05-22
Consider moving location of 'Logout' button

### LINKED ISSUE #960: Consider moving location of 'Logout' button
This would be hugely appreciated from a UX standpoint. The user page and logout options are so close to one another that I find myself logging out unintentionally from time to time. Nesting it in the gear icon would be a nice, simple way to handle this. 

For as little as I imagine people use it the Logout button, it's too easy to hit by mistake.

Ricardo-Silva91:
fixes #960 

Like asked, I nested the logout link inside the settings drop down.
Used the 'power settings new' icon.
Let me know if any changes/corrections are needed.


howardchung:
LGTM

@blukai ?

Issue #849, opened 2017-03-28
Trends should be averaged per week/month/patch/moving mean


previsualconsent:
Instead of showing how the average changes with more games which just has winrate converge on 50% usually, it would be nice to see the actual trend. 

For example: if you want a trend of GPM, you would average gpm of all matches in a time window.
For ease of coding, you could even just average 10-20 consecutive games regardless of match time. 

This way one could tell how much their _data_ has fluctuated over time instead of just get a more more and more accurate average.

howardchung:
Could try it out and see how it looks.  I'm worried that the "buckets" of 10 matches will be affected by outliers.

previsualconsent:
Could put error bars on it.

rafd:
:+1:

I suggest using an [exponential moving mean](https://en.wikipedia.org/wiki/Exponential_smoothing). 

If you approve, I can submit a PR.

Below I describe a few alternatives to showing trends in graphs. I've graphed the same data, my LH/10, which tends to vary a lot between games.  _Pay particular attention to the streak of low scoring games in the middle followed by a few big outliers._

Notice how the cumulative mean graph is pretty much useless vs. any of the other approaches.

Any of the alternatives is a significant improvement over the current approach. In practice, the differences between the others are minor. 



--------


## A. Cumulative Mean

<img width="656" alt="screenshot 2017-04-24 14 15 44" src="https://cloud.githubusercontent.com/assets/89664/25352321/a6f975d4-28f9-11e7-8eb1-415156c9fb65.png">

The current implementation. Doesn't actually show 'trends' (just shows 'convergence to mean' given any data). Data is pretty much a straight line.

```graph(i) = sum(game(0)...game(i)) / i```
 
Cons:

  - data always looks the same (ups and downs then converge to average)
  - initial games affect the graph, but later games have little effect
  - only a significant and sustained change will alter the graph
  - can't detect recent trends

## B. Moving Mean

<img width="662" alt="screenshot 2017-04-24 14 15 26" src="https://cloud.githubusercontent.com/assets/89664/25352425/f7ef0ab2-28f9-11e7-95e2-cf407a1324d3.png">

Average of the last N games. 

```graph(i) = sum(game(i-range)...game(i)) / range```

Pros:

  - actually shows trends
 - weight parameter can be adjusted (could even do it dynamically in the UI)

Cons:

  - only takes into account recent games (within the 'range')
  - affected by short-term outliers
[...]

PR #972, opened 2017-05-25
Modify player trends to compute moving average over player's available match history.

vaibhav-y:
Preliminary PR for #849. Things that may need to be addressed:

* Choice of smoothing factor (how do we arrive at this value?)
* Potential addition of new controls to `/trends/*` to control smoothing factor (vary over the interval `(0,1]`)


howardchung:
Hmm, I thought we wanted to implement a moving mean approach?

vaibhav-y:
I can write up something for moving mean on a separate branch, so we get a comparison of the two in action over actual data. However I'm unsure how to do that idiomatically in node (since I don't use node). I'll try to look up the right way to do it here. Currently I'm thinking of keeping a buffer and counter, and updating the buffer to store the moving mean value.

howardchung:
You have access to the entire array, so you can just create your own array, then .slice() pieces of the original array, compute the mean and push that into your new array.  It's not as clean as .reduce() but it does the job.

vaibhav-y:
So I've done this in a roundabout way to save on speed. The initial reduce is still there, and it serves to compute the partial sums of data over the match history. The sum over the number of games can then be simply computed by taking the difference over a region of the required size. This is faster as the original method has a lot of repetitive addition and this saves by performing all of it at once while reading the response.

src/actions/player/playerTrendsActions.js
```diff
@@ -63,9 +68,22 @@ export const getPlayerTrends = (playerId, options = {}, fieldName) => (dispatch)
...
           win,
         });
-
         return dataList;
       }, []);
+
+      // Apply mean-smoothing by taking differences of the previously computed integral
+      for (let i = 0; i < trends.length; i += 1) {
+        if (i < trends.length - chunkSize) {
+          trends[i].value = (trends[i + chunkSize].value - trends[i].value) / chunkSize;
```
howardchung: Instead of assigning to the `i`th element I think it should be the `i + chunkSize` element since we want each data point to reflect the last 20 games.

howardchung: Similarly I think we should start at 20 and go to end

src/actions/player/playerTrendsActions.js
```diff
@@ -63,9 +68,22 @@ export const getPlayerTrends = (playerId, options = {}, fieldName) => (dispatch)
...
+        } else {
```
howardchung: Don't think we need this else block, ultimately we only want to do this this 480 times? (500 - chunkSize)  Moving mean isn't valid for the first 20 elements.

howardchung: Probably we want to slice them off the array once we're done (so trends only ends up being 480 elements

howardchung:
Can you also update the subheader text on the trends page to reflect the new calculation?

src/actions/player/playerTrendsActions.js
vaibhav-y: Gotcha. I've updated the loop to start from 20 and `data(i) = data(i) - data(i - chunkSize)` (using `i + chunSize` would invalidate the most recent 20. This way we only remove the first 20 from consideration as you asked)

src/actions/player/playerTrendsActions.js
vaibhav-y: Else block was due to last 20 having bad data, that I tried to fix, now that we've decided to drop the first 20 games, this is not necessary. Will be removed

vaibhav-y: Done. For the subheader text, what would be preferable? I see there are various internationalization files, do I update the statement in all of them. I think we can go with "Moving average by 20 games over last 480 games"

vaibhav-y: This is what it looks like now image

src/lang/bg-BG.json
```diff
@@ -810,7 +810,7 @@
   "histograms_name":"Histograms",
   "histograms_description":"Percentages indicate win rates for the labeled bin",
   "trends_name":"Trends",
-  "trends_description":"Cumulative average over last 500 games",
+  "trends_description":"Moving average of 20 games over last 480 games",
```
howardchung: I think just "Moving mean of last 20 games"

howardchung: You only need to update the en-US.json file.  The rest will be filled in automatically during the localization step.

vaibhav-y: Done. I reverted the rest. Changes are now only in en-US.json

src/lang/bg-BG.json
```diff
@@ -866,4 +866,4 @@
   "region_15":"Перу",
   "region_16":"Индия",
   "region_19":"Япония"
-}
\ No newline at end of file
+}
```
howardchung: Can you also revert these newline adds?  might conflict with the localization PR.

src/actions/player/playerTrendsActions.js
```diff
@@ -63,9 +68,21 @@ export const getPlayerTrends = (playerId, options = {}, fieldName) => (dispatch)
...
+      // Compute in reverse order so that first n can be discarded
+      for (let i = trends.length - 1; i > chunkSize - 1; i -= 1) {
+        trends[i].value = (trends[i].value - trends[i - chunkSize].value) / chunkSize;
+
+        // Update graph index so it starts at 1 (since we only display 480 at a time)
+        trends[i].x -= chunkSize;
+      }
+
+      // Discard first 20 elements
+      trends.splice(0, chunkSize);
```
howardchung: This seems to no-op if there are fewer than 20 elements in the array.

howardchung: nvm, I was looking at the return value of the function, not the mutated value of trends.

PR #1006, opened 2017-06-08
More info into table for small screen

vovastradamus:


howardchung:
Do you have a screenshot showing the difference?

vovastradamus:
image
image
expand table to width screen

howardchung:
@blukai @dartharaknus opinions on this?  I'm not sure how I like how it looks with no border.  The more correct fix is to tab the table in mobile view or something.

howardchung:
OK, let's give it a try

PR #1044, opened 2017-06-30
Fix for error in #1042

### LINKED ISSUE #1042: Error thrown when viewing Story
With game: https://www.opendota.com/matches/3280538799/story

I get exception: `Cannot read property 'key' of undefined | stack: TypeError: Cannot read property 'key' of undefined
    at re (https://www.opendota.com/build/bundle.js:56:341737)
    at t.value (https://www.opendota.com/build/bundle.js:56:343303)
    at t.value (https://www.opendota.com/build/bundle.js:56:343501)`

I think it's the truthiness check here.  https://github.com/odota/ui/blob/master/src/components/Match/MatchStory.jsx#L579

killerLog will always evaluate to true, even if it's an empty array.  Should check that killerLog is an instanceof an array, THEN check for the 0th index and whether or not it's an object.

I'd make the change myself, but I'm so unfamiliar with the codebase at this point that I'm worried about breaking other stuff.

gu3st:
fixes #1042 

Fix for the error I reported on Story in #1042 

That said, that error should be looked at in the parser as well.. as it's a weird case where a First Blood event doesn't have a target.  I'll attach the replay later as a separate bug.

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -576,7 +576,7 @@ const generateStory = (match) => {
     events.push(new FirstbloodEvent(match,
       match.objectives[fbIndex].time,
       match.objectives[fbIndex].player_slot,
-      (killerLog instanceof Array && killerLog[0] ? killerLog[0].key : null)));
+      (Array.isArray(killerLog) && killerLog[0] ? killerLog[0].key : null)));
   }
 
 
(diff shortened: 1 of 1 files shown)


src/components/Match/MatchStory.jsx
```diff
@@ -576,7 +576,7 @@ const generateStory = (match) => {
     events.push(new FirstbloodEvent(match,
       match.objectives[fbIndex].time,
       match.objectives[fbIndex].player_slot,
-      killerLog ? killerLog[0].key : null));
+      (killerLog instanceof Array && killerLog[0] ? killerLog[0].key : null)));
```
howardchung: According to this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

Arrays.isArray() is preferred.

Alternatively just `killerLog && killerLog[0]` is probably sufficient to handle this case

howardchung:
LGTM, will let @mdiller take a look anyway

PR #1066, opened 2017-07-15
Story updates

mdiller:
Switched to the new `building_kill` format in the objectives log, while retaining support for the old format


src/components/Match/MatchStory.jsx
```diff
@@ -126,6 +126,15 @@ const formatList = (items, noneValue = []) => {
   }
 };
 
+// is given a hero in the format of npc_dota_hero_{heroname}, and returns the player corresponding to that hero, else undefined
+const heroToPlayer = (match, heroname) => match.players.find((player) => {
```
howardchung: doesn't the event have player_slot that you can use?

src/components/Match/MatchStory.jsx
```diff
@@ -326,41 +335,70 @@ class LanesEvent extends StoryEvent {
   }
 }
 
+// returnes a formatted template when given a TowerEvent or BarracksEvent
+const formatBuilding = (event) => {
+  let template = strings.story_building_destroy;
+  if (event.player) {
+    template = event.player.isRadiant === event.team ? strings.story_building_deny_player : strings.story_building_destroy_player;
```
howardchung: team is a number, isRadiant is a boolean?  Do you need a cast here?

src/components/Match/MatchStory.jsx
mdiller: I was figuring that the new `key` field would be more reliable than player_slot. Should I change it back?

src/components/Match/MatchStory.jsx
mdiller: for events, `team` is a boolean where radiant is true, and dire is false. It's different from how they are passed in through the objectives array.

src/components/Match/MatchStory.jsx
howardchung: I think the player_slot is populated on the server side using similar logic, so probably easier not to reimplement it

mdiller: sounds good

PR #1073, opened 2017-07-19
ti7 predictions

howardchung:
Add page for TI7 predictions

Since this is a temporary page I opted not to externalize the strings.

Demo:
http://hchung.westus.cloudapp.azure.com:8080/predictions

Any design comments @blukai ?


blukai:
Time values should be formated, maybe add hero images for hero related columns and for player - add links to player profiles.
BTW maybe split tables for players, heroes, matches, teams?

howardchung:
Switched to TableLink and formatted the times.

I think it's easiest to read if it's all in one table (also makes it easier for people to copy/paste if they are compiling predictions.

Will consider images and links but I think the primary interest on this page will be for a link to the Explorer data, not the hero/player page.

howardchung:
There seems to be an issue where TableLink is re-encoding the percent characters in the link string.

dartharaknus:
Can we have a tile view like how it is in Dota client? This list view kinda feels non-interesting.

howardchung:
I actually think the table is easier and quicker to digest, although I agree tiles would probably be more aesthetically pleasing.

PR #1077, opened 2017-07-23
Add Networth Indicators to Story

### LINKED ISSUE #1065: Story: Net Worth difference at x minutes in
The story shows a heading every 10 minutes. It would be nice to show the net worth difference at each checkpoint, e.g:

screen shot 2017-07-12 at 2 07 31 pm

or something similar.

Or if not in the header, as a line item directly below it.

mdiller:
Adds a networth indicator to each time marker in the story tab.

Suggestions welcome. I feel like there is a better way to show the text here.

fixes #1065

Example:
image

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -29,7 +29,7 @@ const GoldSpan = amount => (
       height="17px"
       role="presentation"
       src={`${API_HOST}/apps/dota2/images/tooltips/gold.png`}
-      style={{ marginLeft: 3 }}
+      style={{ marginLeft: '3px' }}
     />
   </span>
 );
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -569,7 +569,10 @@ class TimeMarkerEvent extends StoryEvent {
           {GoldSpan(this.radiant_gold)}
         </div>
         <div style={{ color: this.radiant_gold > this.dire_gold ? styles.green : styles.red, left: `${this.radiant_percent}%` }}>
-          {Math.abs(this.radiant_percent - this.dire_percent)}% / {GoldSpan(Math.abs(this.radiant_gold - this.dire_gold))} Diff
+          {formatTemplate(strings.story_networth_diff, {
+            percent: Math.abs(this.radiant_percent - this.dire_percent),
+            gold: GoldSpan(Math.abs(this.radiant_gold - this.dire_gold)),
+          })}
         </div>
         <div style={{ width: `${this.dire_percent}%` }}>
           {GoldSpan(this.dire_gold)}
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -539,8 +539,18 @@ class ItemPurchaseEvent extends StoryEvent {
 }
 
 class TimeMarkerEvent extends StoryEvent {
-  constructor(minutes) {
+  constructor(match, minutes) {
     super(minutes * 60);
+    this.radiant_gold = match.players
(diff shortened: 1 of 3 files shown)


howardchung:
Color the diff based on which team is ahead?

mdiller:
image

src/components/Match/MatchStory.jsx
```diff
@@ -23,12 +23,13 @@ const TEAM = {
...
 const GoldSpan = amount => (
   <span key={`gold_${amount}`} className={styles.storySpan}>
-    <font color={styles.golden}>{amount} </font>
+    <font color={styles.golden}>{amount.toLocaleString()}</font>
     <img
       width="25px"
       height="17px"
       role="presentation"
       src={`${API_HOST}/apps/dota2/images/tooltips/gold.png`}
+      style={{ marginLeft: 3 }}
```
howardchung: "3px"?

src/components/Match/MatchStory.jsx
```diff
@@ -550,10 +561,24 @@ class TimeMarkerEvent extends StoryEvent {
...
+      <h3 key={`minute_${this.minutes}_subheading`} style={{ marginBottom: 0 }}>
         {formatTemplate(strings.story_time_marker, { minutes: this.minutes })}
       </h3>,
-      <hr key={`minute_${this.minutes}_hr`} />,
+      <div key={`minute_${this.minutes}_networth_text`} className={styles.storyNetWorthText}>
+        <div style={{ width: `${this.radiant_percent}%` }}>
+          {GoldSpan(this.radiant_gold)}
+        </div>
+        <div style={{ color: this.radiant_gold > this.dire_gold ? styles.green : styles.red, left: `${this.radiant_percent}%` }}>
+          {Math.abs(this.radiant_percent - this.dire_percent)}% / {GoldSpan(Math.abs(this.radiant_gold - this.dire_gold))} Diff
```
howardchung: externalize the "DIff" string

src/components/Match/MatchStory.jsx
```diff
@@ -562,10 +587,17 @@ class GameoverEvent extends StoryEvent {
...
+      .filter(player => player.isRadiant)
+      .map(player => player.kills)
+      .reduce((a, b) => a + b, 0);
+    this.dire_score = match.dire_score || match.players
+      .filter(player => !player.isRadiant)
+      .map(player => player.kills)
+      .reduce((a, b) => a + b, 0);
   }
   format() {
+    console.log(`score: ${this.radiant_score}`);
```
howardchung: remove logging when debugging complete

mdiller: whoops

src/components/Match/MatchStory.jsx
mdiller: Whoops.
Also, should this be "adv" or some other shortened version of "advantage" since we are specifying who has the advantage via color?

howardchung: Think either is fine

src/lang/en-US.json
```diff
@@ -561,6 +561,7 @@
   "story_item_purchase": "{player} purchased a {item} at {time}",
   "story_predicted_victory": "{players} predicted {team} would win",
   "story_predicted_victory_empty": "No one",
+  "story_networth_diff": "{percent}% / {gold} Diff",
```
howardchung: I don't think the percent part really needs to be localized, only the Diff part, but if this is easier then just keep it

mdiller: I did it this way because different languages might prefer something other than '/' and also the order/placement of the 'Diff' could be different in other languages as well.

PR #1081, opened 2017-07-27
Add lasthit delta to the farm table

dali99:
Pull request for adding delta last hits to the last section of the farm tab

screen shot 2017-07-27 at 08 48 28-fullpage


howardchung:
Maybe it should only show up on hover as a tooltip?

@blukai ?

dali99:
Changed commit message and pull request title to what is actually in the pull request.

Sorry, was pretty tired when making this, and tooltip hover was the original idea, but we found out it was kinda nice not having to hover to see the information.

PR #1092, opened 2017-07-30
1033 abandoned indicator player sumary

### LINKED ISSUE #405: Party size information on match row

### LINKED ISSUE #1033: Show abandoned indicator on player summary
Why does no stats recorded count as loss, please fix it. By the way i realy like opendota more than dotabuff.

aarmora:
fixes #1033 
fixes #405 

This should address [this issue](https://github.com/odota/ui/issues/1033) and [this issue](https://github.com/odota/ui/issues/405).


howardchung:
The issue with party_size should be fixed and show up in future matches.  Do you want to put that change in this PR as well?

aarmora:
Yes, I'll do that. Thanks!

src/utility/index.jsx
```diff
@@ -265,14 +272,24 @@ export const transformations = {
...
-  match_id_and_game_mode: (row, col, field) => (
-    <div>
-      <TableLink to={`/matches/${field}`}>{field}</TableLink>
-      <span className={subTextStyle.subText} style={{ display: 'block', marginTop: 1 }}>
-        {strings[`game_mode_${row.game_mode}`]} / {strings[`lobby_type_${row.lobby_type}`]}
-      </span>
-    </div>
-  ),
+  match_id_and_game_mode: (row, col, field) => {
+    const partySize = (partySize) => {
```
howardchung: maybe we want to indicate something for solo as well?

aarmora: Yeah, I wondered about that. I'll add a single person for that.

aarmora: An example 
image.

I couldn't find any matches that were in a party. Null would return nothing, solo would return like the above picture.

aarmora:
@blukai Not sure how to respond to this. Is this a request to shrink them? How much would seem not too big to you?

src/utility/index.jsx
```diff
@@ -264,14 +273,30 @@ export const transformations = {
...
+        return [
+          <SocialPerson color="rgb(179, 179, 179)" style={iconStyles} />,
+        ];
+      } else if (partySize === null) {
+        return null;
+      }
+
+      return [
+        <SocialPeople color="rgb(179, 179, 179)" style={iconStyles} />,
+        ` x${row.party_size}`,
```
blukai: instead of adding space here I recomend setting margin-left to 5px

src/utility/index.jsx
```diff
@@ -18,6 +18,14 @@ import subTextStyle from 'components/Visualizations/Table/subText.css';
 import findLast from 'lodash.findlast';
 import _ from 'lodash/fp';
 import util from 'util';
+// import SvgIcon from 'material-ui/SvgIcon';
+import SocialPeople from 'material-ui/svg-icons/social/people';
+import SocialPerson from 'material-ui/svg-icons/social/person';
+
+const iconStyles = {
+  marginBottom: -5,
```
blukai: if u'll set width & height to 18px, u can delete negative margin bottom (I'm not a big fan of negative margins 😛) and use `vertical-align`

src/utility/index.jsx
```diff
@@ -18,6 +18,14 @@ import subTextStyle from 'components/Visualizations/Table/subText.css';
...
+  marginLeft: 10,
```
blukai: 5, like right 🦄

aarmora:
@blukai Great feedback! I've updated my code to fit better with your suggestions.

PR #1095, opened 2017-07-31
Make the chat looks like actually chat looks like

blukai:
Mobile friendly ✔️, with spam detection & filters

image


dartharaknus:
Looks great. But I think the filters need to be above the table. Side doesn't work as good.

src/lang/en-US.json
```diff
@@ -919,6 +919,14 @@
   "vision_placed_sentry": "placed Sentry at",
   "vision_ward_log": "Ward Log",
 
+  "chat_radiant": "Radiant",
```
howardchung: I think we can reuse general_dire and general_radiant

src/components/Match/Chat/Chat.jsx
```diff
@@ -0,0 +1,225 @@
...
+                {message}
+              </article>
+            </li>
+          );
+        }))}
+      </ul>
+    );
+
+    const Filters = () => (
+      <ul className={styles.Filters}>
```
howardchung: I think these should be labeled as filters

blukai:
@howardchung I deleted duplicated by me strings from lang and added heading for filters.
@dartharaknus I'll move the filters above the messages, and then we'll decide how best.

blukai:
I replaced switchers by checks since switchers are awful in horizontal placement. Also I was thinking to categorize it, like 'side', 'type', 'other' and replacing `chatwheel` filter by 2 filters: phrases and audio. Btw I like more how it looks on top instead of right. Let me know what u think.

image

image

I didn't push commit yet, will wait for reply and then act 🦄

howardchung:
It's not immediately clear that the number is a count of that type of message.

blukai:
image

src/components/Match/Chat/Chat.jsx
```diff
@@ -0,0 +1,294 @@
...
+
+    this.state = {
+      radiant: true,
+      dire: true,
+      text: true,
+      phrases: true,
+      audio: true,
+      all: true,
+      allies: true,
+      spam: false,
```
howardchung: I think default behavior should be unfiltered.  We just want to show the chat as it was, and allow the user the option to filter if they want?

src/components/Match/Chat/Chat.jsx
```diff
@@ -0,0 +1,294 @@
...
+        f: (arr = this.raw) => arr.filter(msg => !isRadiant(msg.player_slot)),
+        type: 'faction',
+        disabled: () => !this.state.radiant,
+      },
+      text: {
+        f: (arr = this.raw) => arr.filter(msg => msg.type === 'chat'),
+        type: 'type',
+        disabled: () => this.state.phrases === false && this.state.audio === false,
+      },
+      phrases: {
```
howardchung: I've actually recently removed phrases from the data (leaving only audio chatwheel messages)

src/lang/en-US.json
```diff
@@ -296,7 +296,7 @@
   "heading_analysis": "Analysis",
   "heading_cosmetics": "Cosmetics",
   "heading_log": "Log",
-  "heading_chat": "Chat",
+  "heading_chat": "Chat messages",
```
howardchung: I think we should just leave this as "Chat"

src/lang/en-US.json
```diff
@@ -341,11 +341,12 @@
   "heading_median": "Median",
   "heading_distinct_heroes": "Distinct Heroes",
   "heading_team_elo_rankings": "Team Elo Rankings",
-  
+
   "subheading_avg_and_max": "in last %d matches",
+  "subheading_chat": "and categorized filters with counts beside",
```
howardchung: I don't think this subheading is needed.

src/components/Match/Chat/Chat.jsx
blukai: Hm, anyway I think better to keep this for matches that got this data

PR #1098, opened 2017-08-01
Made player wardmap responsive to screen size (#390)

### LINKED ISSUE #390: make player wardmap responsive
player page wardmaps should scale to the size of the screen, as given in redux-responsive.

jhonnold:
fixes #390 

I added an extra field for the redux-responsive and started the window listener in the actual component itself (not sure if you wanted it watching the whole time).

Padding is 25 on either side so width - 50 and maxed to 1440.

Screens:

fireshot capture 13 - zomby - wardmap - opendota -_ - http___localhost_8080_players_58052218_wardmap

fireshot capture 14 - zomby - wardmap - opendota -_ - http___localhost_8080_players_58052218_wardmap

fireshot capture 12 - zomby - wardmap - opendota -_ - http___localhost_8080_players_58052218_wardmap

### DIFF
--- a/src/components/Player/Pages/Wardmap/Wardmap.jsx
+++ b/src/components/Player/Pages/Wardmap/Wardmap.jsx
@@ -1,12 +1,12 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import { calculateResponsiveState } from 'redux-responsive';
 import { getPlayerWardmap } from 'actions';
 import Heatmap from 'components/Heatmap';
 import Container from 'components/Container';
 import strings from 'lang';
 import { unpackPositionData } from 'utility';
 import styles from './Wardmap.css';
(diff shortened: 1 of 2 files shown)


src/components/Player/Pages/Wardmap/Wardmap.jsx
```diff
@@ -14,6 +15,7 @@ const getData = (props) => {
 class RequestLayer extends React.Component {
   componentWillMount() {
     getData(this.props);
+    window.addEventListener('resize', this.props.updateWindowSize);
```
howardchung: I think you can re-render based on a prop change passed from redux-responsive?  So you don't need to add your own event listener

howardchung:
In that case, LGTM, although it would be nice not to have to bind our own event listeners for each responsive component (there are a few others)

PR #1111, opened 2017-08-14
Replaced flag-icon-css with local flags

flankstaek:
Removed npm module flag-icon-css and used svg flag icons in
assets/images/flags to replace them because it seemed like nothing was
being read from the flag-icon-css library.

This probably wasn't ideal but it was easy enough to just copy the files over and it's one less npm module to worry about.


blukai:
flags as npm module to have no worries about flag updates since some countries ceased to exist or split or new ones will be founded

howardchung:
I think flags update rarely enough that it shouldn't be much of an issue (especially because I think Valve doesn't update their country list anyway, seeing as nonexistent countries like Yugoslavia are still there)

PR #1119, opened 2017-08-17
Reduce max width of app

howardchung:
Got some feedback suggesting reducing the max-width to reduce the amount of whitespace on large screens.

I adjusted it to 1080px max width.  Seems like 1920x1080 is the most common screen size.

UPDATE: 1080 might be too aggressive, some of the tables are getting squashed.  Changed to 1200


howardchung:
Will merge this, we can check it out in stage and tweak the numbers if there are styling issues with the reduced width.

PR #1120, opened 2017-08-17
Static Order for Tooltips in XP/Gold Graph

### LINKED ISSUE #1078: Order of rows changes on Networth/XP table
This is the table you get on hover for the graphs of networth and XP. Seems like they should be consistent.
image
image

timestep:
fixes #1078 

https://github.com/odota/ui/issues/1078

C3 has an internal tooltip order function that gets run unless defined via the order key. I'm open to suggestions on how to identify if it's a gold/xp graph since the function is in a closure without access to the `drawGraphs` `props` argument. For now, I think it's an ideal solution for something like this.

### DIFF
--- a/src/components/Match/MatchGraph.jsx
+++ b/src/components/Match/MatchGraph.jsx
@@ -49,8 +49,11 @@ const drawGraphs = (props, id) => {
       },
       tooltip: {
         contents(d, defaultTitleFormat, defaultValueFormat, color) {
-          d.sort((a, b) => b.value - a.value);
           return this.getTooltipContent(d, defaultTitleFormat, valueFormat || defaultValueFormat, color);
+        },
+        order: (a, b) => {
+          if (a.id === 'Gold' || b.id === 'Gold') return a.id - b.id;
+          return b.value - a.value;
         },
       },
     });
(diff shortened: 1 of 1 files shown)


src/components/Match/MatchGraph.jsx
```diff
@@ -49,9 +49,12 @@ const drawGraphs = (props, id) => {
       },
       tooltip: {
         contents(d, defaultTitleFormat, defaultValueFormat, color) {
```
howardchung: I think this function can be removed completely, the only reason we implemented it was for the sort

src/components/Match/MatchGraph.jsx
```diff
@@ -49,9 +49,12 @@ const drawGraphs = (props, id) => {
...
-          d.sort((a, b) => b.value - a.value);
           return this.getTooltipContent(d, defaultTitleFormat, valueFormat || defaultValueFormat, color);
         },
+        order: (a, b) => {
+          if (a.id === 'Gold' || b.id === 'Gold') return a.id - b.id;
```
howardchung: Can you check if `props.type === 'difference'` here to determine whether to sort?

src/components/Match/MatchGraph.jsx
timestep: It provides the level of the experience via Line24 on this file. If we removed it we would lose that functionality.

src/components/Match/MatchGraph.jsx
timestep: Yep! Done!

timestep:
@howardchung Updated

Issue #1122, opened 2017-08-18
Unable to load profile


quekyuxuan:
Hi,

When I try to access my profile https://www.opendota.com/players/124964145 , I get stuck with a loading screen (screenshot below).
<img width="1321" alt="screen shot 2017-08-18 at 04 15 14" src="https://user-images.githubusercontent.com/12409804/29442810-e878e526-83cb-11e7-853a-540a88a47cfe.png">
I am able to load other profiles normally, and have tried to load the profile when both when logged in and out from opendota. I have checked the box (and saved the setting) to expose public match data in the dota 2 client. For what it's worth, I've tried accessing this on both Chrome (version 60.0.3112.90) and Safari (version 10.1.1) on macOS Sierra, as well as on Chrome for Android.

Thanks

mbyakow:
Console output ![image](https://user-images.githubusercontent.com/7271758/29443219-babad092-8400-11e7-95b8-b7dca9980b14.png)

howardchung:
Thanks, this looks like an issue with a change to the mmr estimate data format.  Will push a fix on the API

mbyakow:
#1123

howardchung:
I rolled out an API fix to prevent the crash, looks like the header is loading properly now.

@mbyakow also submitted a fix that handles this case and better defaults for players who no matches recorded.

I'll try to ship that to prod this weekend.

PR #1123, opened 2017-08-18
Displaying of default values for player's MMR and win rate.

mbyakow:


src/components/Player/Header/PlayerStats.jsx
```diff
@@ -46,7 +46,7 @@ export const PlayerStatsCards = ({
         />
         <CardTitle
           className={styles.playerStats}
-          subtitle={`${((wins / (wins + losses)) * 100).toFixed(2)}%`}
+          subtitle={wins + losses ? `${((wins / (wins + losses)) * 100).toFixed(2)}%` : 'N/A'}
```
howardchung: I think there's a N/A string in the strings file to use, otherwise can you put it in en-US.json?

mbyakow:
Fix for #1122

PR #1132, opened 2017-08-20
Linked overview page table headers to corresponding tabs

### LINKED ISSUE #1106: Make table titles in overview links to in depth section
If I'm looking at something like heroes played on overview and want to see the big table, I'm too lazy to scroll up and click the tab there. It would be cool to be able to just click the table title.

jhonnold:
fixes #1106 

Added a titleTo prop into the Container - which passes into Header - which then renders the link if need be.

(The header is blue as my mouse hovered over it)
image

### DIFF
--- a/src/components/Container/index.jsx
+++ b/src/components/Container/index.jsx
@@ -22,9 +22,9 @@ AsyncContainer.propTypes = {
   children: node,
 };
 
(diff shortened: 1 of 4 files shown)


jhonnold:
#1106

Issue #1131, opened 2017-08-20
Ability Build view


Xanvial:
Hi
There's a bug when viewing ability build on mobile, I can't view all levels after pressing triple dots button. Only last few level is visible.
OS: Android
Browser: Opera ver. 42.9
Resolution: 1080 x 1920 using portrait view

Suggestion, maybe move the ability build on different table on Overview tab. My experience is I need a while to found where to view it because the placement isn't noticeable

howardchung:
@blukai do you think we should split it off into a table?  I agree it's probably better to have it visible without requiring a user interaction.

blukai:
I think we can do it

jhonnold:
Working on putting this in its own table - did we still want it on the overview page or on its own tab now?

howardchung:
fixed in #1135

PR #1135, opened 2017-08-24
Moved Ability Build to its own table

jhonnold:
I have moved the ability build to its own table per #1131. 

I made a new table called AbilityBuildTable - following standard layout for that.
I then proceeded to map the array from the ability build into keys in order for it to be used by the prop in the Table.

I had to standardize a couple of values due to how they come from the backend.

I'm not sure if this will cause a performance hit or not - if it does I will have to figure something else out.

Picture with dire scrolled to end:
image

### DIFF
--- a/src/components/Match/AbilityBuildTable.jsx
+++ b/src/components/Match/AbilityBuildTable.jsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { isRadiant, getTeamName } from 'utility';
+import { IconRadiant, IconDire } from 'components/Icons';
+import Heading from 'components/Heading';
+import Table from 'components/Table';
+import styles from './Match.css';
+
+const filterMatchPlayers = (players, team = '') =>
+  players
+    .filter(player => (team === 'radiant' && isRadiant(player.player_slot)) || (team === 'dire' && !isRadiant(player.player_slot)) || team === '')
+    .sort((a, b) => a.player_slot - b.player_slot);
+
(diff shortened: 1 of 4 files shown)


howardchung:
Could we reduce the padding in order to try to fit all 25 columns without scrolling?

Also I think we can remove the pvgna icon to save some more space, since they're still on the main table.

blukai:
maybe better move **ability builds** to **purchases** tab and call it **builds** instead?

jhonnold:
Okay I'll move it to the purchases and rename that tab to builds and ill reduce padding to get all 25 on the screen for default size

coreymaher:
Ability builds are available without parsing a replay, correct? Doesn't that mean it shouldn't be in a tab that requires a parse?

howardchung:
Yeah, I think we should leave it on the overview tab as it's basic information.

jhonnold:
Alright won't be moving it then

jhonnold:
Here it is after i set a negative margin to shrink padding and lowered the size of the image and removed the pvgna icon

image

jhonnold:
Wait don't pass this in just yet - i just noticed the MK in the screen isn't showing the levels correctly. I'll look into that

jhonnold:
Or not - dotabuff is showing it the same way https://www.dotabuff.com/matches/3360245838

howardchung:
I'll merge this now, @blukai if you have any further comments check the stage version and we can submit any additional fixes before deploying to production?

PR #1138, opened 2017-08-27
Story improvements

drusepth:
Hey guys, just wanted to include a few small updates to `MatchStory`. I'm tempted to make more, but I also recognize a lot of changes are preferential, so I'll keep PRs small to enable picking and choosing which changes we want. 

This PR:

1. Switches between 'a' and 'an' before game types in the `IntroEvent`, so we'll see **an** All Draft and **a** Captain's Draft.
2. Includes the final length of the match in the game intro, so readers know about how long to expect the match to last. 
  *- This also approximates the game time to either "about X minutes", "almost an hour", and "over X hours" to read nicely.
3. Adds a new `ChatMessageEvent` to include dialogue from players in the story, which makes for some fun interactions like seeing chat after a first blood, or seeing everyone's "gg"s after the match ends.

IMO, adding dialogue to the match story makes it read more like a story with characters rather than a narrative of the game's events. 

Some screenshots:

2017-08-27-210311_648x109_scrot

2017-08-27-211746_512x149_scrot

2017-08-27-211211_918x173_scrot

2017-08-27-210624_568x207_scrot

cc @mdiller @howardchung

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -114,11 +114,11 @@ const articleFor = (followingWord) => {
   // Whether we use a or an depends on the sound of the following word, but that's much hardder to detect programmatically,
   // so we're looking solely at vowel usage for now.
   if (['A', 'E', 'I', 'O', 'U'].includes(followingWord.charAt(0))) {
-    return strings['article_before_vowel_sound'];
-  } else {
-    return strings['article_before_consonant_sound'];
+    return strings.article_before_vowel_sound;
   }
-}
+
+  return strings.article_before_consonant_sound;
+};
 
 const formatApproximateTime = (timeSeconds) => {
   const timeMinutes = parseInt(timeSeconds / 60, 10);
@@ -133,10 +133,9 @@ const formatApproximateTime = (timeSeconds) => {
   } else if (timeMinutes >= 50) {
     // If the time is between 50 and 60 minutes, describe it as "almost an hour"
     return `${strings.advb_almost} ${strings.time_h}`;
-  } else {
-    // Otherwise, describe the time in minutes
-    return `${strings.advb_about} ${util.format(strings.time_mm, timeMinutes)}`;
   }
+  // Otherwise, describe the time in minutes
+  return `${strings.advb_about} ${util.format(strings.time_mm, timeMinutes)}`;
 };
 
 const renderSentence = (template, dict) => toSentence(formatTemplate(template, dict));
@@ -220,7 +219,7 @@ class ChatMessageEvent extends StoryEvent {
     return formatTemplate(strings.story_chatmessage, {
       player: PlayerSpan(this.player),
       message: this.message,
-      said_verb: (this.message.charAt(this.message.length - 1) === '?') ? strings['story_chat_asked'] : strings['story_chat_said'],
+      said_verb: (this.message.charAt(this.message.length - 1) === '?') ? strings.story_chat_asked : strings.story_chat_said,
     });
   }
 }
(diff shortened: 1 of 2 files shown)


howardchung:
haha, I like this.  @mdiller for review

src/components/Match/MatchStory.jsx
```diff
@@ -109,6 +109,33 @@ const toSentence = (content) => {
   return result;
 };
 
+const pluralize = (count, object, pluralization_override=null) => {
```
howardchung: lint will probably want this in camelCase

src/components/Match/MatchStory.jsx
```diff
@@ -146,14 +173,17 @@ class IntroEvent extends StoryEvent {
     this.game_mode = match.game_mode;
     this.region = match.region;
     this.date = new Date(match.start_time * 1000);
+    this.match_duration_seconds = match.duration;
   }
   format() {
     return formatTemplate(strings.story_intro, {
+      game_mode_article: ['A', 'E', 'I', 'O', 'U'].includes(strings[`game_mode_${this.game_mode}`].charAt(0)) ? 'an' : 'a',
```
howardchung: More generally I think the rule is to use "an" if the first syllable is a vowel sound.  For example, "hour" uses "an" even though the first letter is not a vowel.  I don't know if that applies to any of the game modes right now, but worth considering.

howardchung: 'an' and 'a' should probably go into the string file

src/components/Match/MatchStory.jsx
```diff
@@ -176,6 +206,22 @@ class FirstbloodEvent extends StoryEvent {
...
+    super(obj.time + 70);
+    this.type = obj.type;
+    this.player = match.players.find(player => player.player_slot == obj.player_slot);
+    this.message = obj.key.trim();
+  }
+  format() {
+    return formatTemplate(strings.story_chatmessage, {
+      player: PlayerSpan(this.player),
+      message: this.message,
+      said_verb: (this.message.charAt(this.message.length - 1) == '?') ? 'asked' : 'said'
```
howardchung: move words to string file

src/components/Match/MatchStory.jsx
drusepth: Oh, good call. Should I go ahead and camelCasify all of these snake_case variables? (e.g. pluralized_object, time_seconds, etc?)

src/lang/en-US.json
```diff
@@ -751,8 +752,10 @@
   "time_s": "a second",
   "time_ss": "%d seconds",
   "time_m": "a minute",
+  "time_ms": "minute",
```
howardchung: just curious, what does the "s" mean here?

src/components/Match/MatchStory.jsx
howardchung: yes please.  `npm run lint` will tell you if you need to make any other lint fixes

drusepth: Awesome, thanks for the tip. Will make those changes.

src/lang/en-US.json
drusepth: I'm very open to changing the key here. I noticed we already had values for "a second", "a minute", etc until time_x, so I was sticking the singular word until time_xs, for "singular". 

Now that I look at it, that's not very clear. Any suggestions for a better key here? I'd like to avoid dropping the 'a' from the existing time_x keys to avoid going through and reworking wherever they're used, if possible.

howardchung: you should be able to use time_h or time_hh depending on the number of hours (no need to add a new key?)

src/components/Match/MatchStory.jsx
drusepth: Good call. Doesn't affect any of the current game modes, but I abstracted out an `articleFor(word)` method we can improve down the line if we ever get a game mode that breaks things (or want to use this elsewhere).

drusepth:
@howardchung Thanks for the comments. I made some changes for eslint and some abstractions around using the existing time keys we have, so it should be good to review again whenever is convenient for you. :+1:

mdiller:
Looks good to me!

Issue #1121, opened 2017-08-17
Show link to instructions for enabling MMR updates


Rabid5:
![graph for bug report]
![graph for bug report](https://user-images.githubusercontent.com/31115124/29438192-0c20c3c0-8383-11e7-992b-3abbcaf256bb.PNG)

howardchung:
Did you follow the directions here?

https://blog.opendota.com/2016/01/13/yasp-mmr-and-you/

blukai:
@howardchung I regularly see problems with updating the MMR, perhaps we should add text that explains how it works on the MMR page?

howardchung:
Yeah, that's probably a good idea

CauchyAbel:
@blukai Waiting on your idea of how to implement it. The original post has been fixed to show OpenDota, mmr, and you

blukai:
@CauchyAbel in `src/components/Alerts` u can fing `Warning`, u can create`Info` component by copying warning and replacing icon, color. I guess add link to the blogpost that explains how it works whould be enough. 🤔

CauchyAbel:
@blukai Which one is better?

![](https://cdn.discordapp.com/attachments/197560147529170944/351576027702624267/unknown.png)

![](https://cdn.discordapp.com/attachments/197560147529170944/351576327884636160/unknown.png)

blukai:
@CauchyAbel Maybe it's better to put it under the heading above graph? 🦄

PR #1144, opened 2017-08-30
Cauchy|Added Info component as well as mmr instruction on MMR page.

CauchyAbel:
image

UI as shown, new to react xD

### DIFF
--- a/src/lang/en-US.json
+++ b/src/lang/en-US.json
@@ -443,7 +443,7 @@
   "matches_highest_mmr": "High MMR",
   "matches_lowest_mmr": "Low MMR",
 
-  "mmr_not_up_to_date": "Why is my mmr not up to date?",
+  "mmr_not_up_to_date": "Why is my MMR not up to date?",
 
   "npc_dota_beastmaster_boar_#": "Boar",
   "npc_dota_lesser_eidolon": "Lesser Eidolon",
--- a/src/lang/en-US.json
+++ b/src/lang/en-US.json
@@ -443,6 +443,8 @@
   "matches_highest_mmr": "High MMR",
   "matches_lowest_mmr": "Low MMR",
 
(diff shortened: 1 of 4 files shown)


CauchyAbel:
@blukai @howardchung Fix for #1121 , please review, thanks!

src/lang/en-US.json
```diff
@@ -443,6 +443,8 @@
   "matches_highest_mmr": "High MMR",
   "matches_lowest_mmr": "Low MMR",
 
+  "mmr_not_up_to_date": "Why is my mmr not up to date?",
```
howardchung: capitalize MMR

CauchyAbel:
image

Change pushed, please review @blukai @howardchung

blukai:
Is it possible to make it show if last mmr update was more than ~1 day ago for example?

CauchyAbel:
@blukai I should be able to implement that, will give it a try :) 👍

CauchyAbel:
@blukai Just started up the front end and found that if you don't have a game a day ago, the MMR graph won't show MMR for that day, seems it is only catching the MMR changes. For example I haven't played ranked game for 5 days and it only shows my MMR until 5 days ago. Don't know how we could detect the MMR update though.

blukai:
```js
{
  account_id: 210377205,
  match_id: 3141276771,
  solo_competitive_rank: 5481,
  competitive_rank: null,
  time: "2017-04-26T15:09:11.489Z"
}
```
I guess u can compare the last time when it was updated to current time

CauchyAbel:
@blukai But I mean it is still possible that MMR was updated 5 (more than 1 day) days ago even though it is up to date. Just like in my case I didn't play for five days so the updated time should have 5 days difference to current date.

howardchung:
I think it's fine to just always show the message.

Issue #1040, opened 2017-06-27
Siltbreaker matches are being included in the player profile overview Averages/Maximums in last 20 matches


awailes:
When I go to my Player Profile, and am on the Overview tab, the calculations for the Averages/Maximums in last 20 matches is including  Event / Co-Op Bots matches from siltbreaker (the custom game).  These matches are labeled as Event / Co-Op Bots matches.  These are not real Dota matches and should not be included in the calculations for the Averages/Maximums in last 20 matches.  I have attached a picture that shows the issue.

![bug](https://user-images.githubusercontent.com/29739702/27608784-b3cd8a4a-5b56-11e7-81cf-f40125388bde.png)

blukai:
https://github.com/odota/core/issues/1474

howardchung:
Yeah, although I think it makes more sense to filter these matches from the calculation on the client side, so I'll close the core issue in favor of this one.

PR #1146, opened 2017-09-01
1040 - Filter Siltbreaker matches from averages

zya6yu:
Update Summary.jsx to only exclude matches with game mode type 19
(Event matches) from the recent game averages.

The getRecentPlayerMatches endpoint only returns 20 matches, so
this could end up using fewer than 20 matches, or even 0 matches
(which probably isn't desirable).

Tried using the getPlayerMatches endpoint instead, but that does
not have data on other fields that we care about, e.g. gpm, xpm.

### DIFF
--- a/src/components/Match/AbilityBuildTable.jsx
+++ b/src/components/Match/AbilityBuildTable.jsx
@@ -26,7 +26,7 @@ const standardizeLevel = (skilledAt, obj) => {
 
 const convertArrayToKeys = (obj, fieldName = '') => ({
   ...obj,
-  ...obj[fieldName].reduce(
+  ...(obj[fieldName] || []).reduce(
     (acc, cur, index) => ({
       ...acc,
       [`ability_upgrades_arr_${standardizeLevel(index, obj) + 1}`]: cur,


zya6yu:
Completely new to open source (also to react) so apologies in advance for being unfamiliar with the workflow -- would love to get involved with the project though!

zya6yu:
Also, not sure how to link it properly, but this is the issue I was trying to address: https://github.com/odota/ui/issues/1040

Was developing against my friend's profile id: 69592941

src/components/Player/Pages/Overview/Summary.jsx
```diff
@@ -34,19 +35,28 @@ const SummOfRecMatches = ({ matchesData }) => {
...
 
-  const numRows = Math.min(MAX_MATCHES_ROWS, matchesData.length);
+  const desiredNumRows = Math.min(MAX_MATCHES_ROWS, matchesData.length);
 
-  for (let i = 0; i < numRows; i += 1) {
+  // XXX - getPlayerRecentMatches seems to only return a maximum of 20 matches
+  // meaning that this set of "validMatches" can be fewer than 20. Tried
+  // switching to the getPlayerMatches endpoint but that does not provide
+  // data on several fields (e.g. gold_per_min, xp_per_min, last_hits, etc.)
+  const validMatches = _.take(_.filter(matchesData, (match) => {
```
blukai: I think we are trying to avoid lodash and others, so u can do it by
```js
const validMatches = matchesData
    .filter(match => match.game_mode !== 19)
    .slice(0, desiredNumRows);
```

src/components/Player/Pages/Overview/Summary.jsx
```diff
@@ -34,19 +35,28 @@ const SummOfRecMatches = ({ matchesData }) => {
...
+    // filter out event matches from averages
+    return match.game_mode !== 19;
+  }), desiredNumRows);
+
+  const numRows = validMatches.length;
+
+  _.each(validMatches, (match) => {
```
blukai: here u can use `forEach`

src/components/Player/Pages/Overview/Summary.jsx
```diff
@@ -34,19 +35,28 @@ const SummOfRecMatches = ({ matchesData }) => {
...
     dataKeys.map((key) => {
```
blukai: and here replace `map` with `forEach`

src/components/Player/Pages/Overview/Summary.jsx
```diff
@@ -34,19 +35,28 @@ const SummOfRecMatches = ({ matchesData }) => {
 
   const dataKeys = Object.keys(data);
 
-  const numRows = Math.min(MAX_MATCHES_ROWS, matchesData.length);
+  const desiredNumRows = Math.min(MAX_MATCHES_ROWS, matchesData.length);
 
-  for (let i = 0; i < numRows; i += 1) {
+  // XXX - getPlayerRecentMatches seems to only return a maximum of 20 matches
```
howardchung: getPlayerRecentMatches is more performant and therefore we use it here.

getPlayerMatches can access all columns, but you need to declare the ones you want with the `project` query parameter, otherwise it just returns a small default set.

howardchung:
You might just want to change the header to avoid a specific number so we don't have that issue

PR #1151, opened 2017-09-04
1040 - Filter Siltbreaker matches from averages

### LINKED ISSUE #1040: Siltbreaker matches are being included in the player profile overview Averages/Maximums in last 20 matches

zya6yu:
fixes #1040 

Only pass non-siltbreaker matches from Overview to Summary

This could be updated to use the getPlayerMatches endpoint instead
of getPlayerRecentMatches if we specify the fields we want to
display averages on (and probably some upper bound of matches we
want to retrieve.)

### DIFF
--- a/src/components/Player/Pages/Overview/Overview.jsx
+++ b/src/components/Player/Pages/Overview/Overview.jsx
@@ -22,6 +22,8 @@ const MAX_HEROES_ROWS = 10;
 const MAX_PEERS_ROWS = 5;
 
 const Overview = ({
+  validRecentMatches,
+  numValidRecentMatches,
   matchesData,
   matchesLoading,
   matchesError,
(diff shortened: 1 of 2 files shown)


zya6yu:
accidentally closed the other PR when mucking around trying to get my branches in sync... old PR was https://github.com/odota/ui/pull/1146

moved the filtering of matches to overview so we can display the correct number of games being used for the average -- the filtered matches are then passed through to summary and no additional filtering is done there

Wasn't sure what the process was for updating the strings themselves since there are translations into so many different languages

howardchung:
If you need to do a string update just update en-US.json, the rest will be filled automatically later and translated when volunteers get to them via crowdin.

PR #1152, opened 2017-09-05
Update MatchStory (a la Reddit)

stevenchung:
Context: https://www.reddit.com/r/DotA2/comments/6ya3k6/detail/

Quick inline fixed for corner case

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -219,7 +219,9 @@ class ChatMessageEvent extends StoryEvent {
     return formatTemplate(strings.story_chatmessage, {
       player: PlayerSpan(this.player),
       message: this.message,
-      said_verb: (this.message.charAt(this.message.length - 1) === '?') ? strings.story_chat_asked : strings.story_chat_said,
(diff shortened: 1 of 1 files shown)


gu3st:
TBH for meme status I prefer "asked"... but that's just me.

I'm also a fan of the codegolfing I applied to it :D 

`said_verb: this.message.substr(-1,1) === '?' && strings.story_chat_asked || strings.story_chat_said`

howardchung:
I personally think "?" is a valid question (if a minimal one) so I think "asked" is correct.

The Reddit thread does mention a few other things though, like:
* Handling "?!", or some other strings with question marks not as the last character
* Adding additional terms such as "exclaimed" for an all caps message

@drusepth @mdiller for input

drusepth:
Honestly, I actually prefer "asked" on just "?", but not with a super-strong preference. I read it as a question in my head.

I have no idea whether "!?" or "?!" should result in "asked" or "said", though. "Exclaimed" would probably be good here, as well as on all-caps messages and/or messages ending in "!".

howardchung:
I think this is superseded by #1156

PR #1153, opened 2017-09-07
update rune icons

blukai:
before:
image

after:
image


howardchung:
Source for these icons?

blukai:
It's frames from emoticons
image

PR #1156, opened 2017-09-09
Add further sentiment analysis to ChatMessageEvent

NiQDeveloper:
* Added some simple analysis on the message, to make it a little more varied than "said", "asked".
* Moved the strings into the the relevant json - If anyone can translate to other languages that would be good.

example chat log

image

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -159,26 +159,27 @@ const formatList = (items, noneValue = []) => {
 // evaluate the sentiment behind the message - rage, question, statement etc
 const evaluateSentiment = (event, lastMessage) => {
   const { message, player, time } = event;
-  let stringAppendature = 'normal';
+  const sentiment = message.indexOf('?') !== -1 ? ['question'] : ['statement'];
 
-  if (message.split(' ').length > 10) {
-    stringAppendature = 'long';
-  } else if (message.toUpperCase() === message && /\w/.test(message)) {
-    stringAppendature = 'shouted';
-  } else if (/(\?|!|@|~|#|\$){2,}/.test(message)) {
-    stringAppendature = 'excited';
-  }
-
-  let type = message.indexOf('?') !== -1 ? 'question' : 'statement';
-  if (lastMessage && lastMessage.time + 60 > time) {
-    if (lastMessage.player.player_slot === player.player_slot) {
-      type += '_continued';
-    } else if (lastMessage.message.indexOf('?') !== -1) {
-      type += '_response';
+  if (lastMessage && lastMessage.time + 130 > time) {
+    if (lastMessage.player_slot === player.player_slot) {
+      sentiment.push('continued');
+    } else if (lastMessage.key.indexOf('?') !== -1) {
+      sentiment.push('response');
     }
   }
 
-  return strings[`${type}_${stringAppendature}`];
+  if (message.split(' ').length > 10) {
+    sentiment.push('long');
+  } else if (message.toUpperCase() === message && /\w/.test(message)) {
+    sentiment.push('shouted');
+  } else if (/(\?|!|@|~|#|\$){2,}/.test(message)) {
+    sentiment.push('excited');
+  } else {
+    sentiment.push('normal');
+  }
+
+  return strings[sentiment.join('_')];
 };
 
 // Abstract class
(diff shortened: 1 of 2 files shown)


howardchung:
@drusepth @mdiller for review

NiQDeveloper:
@howardchung

I'm used to bitbucket sorry, so unsure how to add reviewers after the PR was created.

I have added a few screenshots with example conversations from recent games. Overall I feel this makes the story really fun to read.

src/components/Match/MatchStory.jsx
```diff
@@ -156,6 +156,31 @@ const formatList = (items, noneValue = []) => {
...
+  } else if (message.toUpperCase() === message && /\w/.test(message)) {
+    stringAppendature = 'shouted';
+  } else if (/(\?|!|@|~|#|\$){2,}/.test(message)) {
+    stringAppendature = 'excited';
+  }
+
+  let type = message.indexOf('?') !== -1 ? 'question' : 'statement';
+  if (lastMessage && lastMessage.time + 60 > time) {
+    if (lastMessage.player.player_slot === player.player_slot) {
+      type += '_continued';
```
howardchung: consider making an array and using .join() on it to avoid having to remember where you need underscores

src/components/Match/MatchStory.jsx
```diff
@@ -691,9 +717,19 @@ const generateStory = (match) => {
...
     .filter(obj => obj.type === 'chat')
-    .map(obj => new ChatMessageEvent(match, obj));
+    .map((obj) => {
+      let event;
+      if (lastMessage !== undefined) {
+        event = new ChatMessageEvent(match, obj, lastMessage);
+      } else {
+        event = new ChatMessageEvent(match, obj);
+      }
+      lastMessage = event;
```
howardchung: maybe do `.map(obj, i)` and then use `i` and `i - 1` to find the last message?  That removes the dependence on state outside the function

NiQDeveloper: Yeah good call, much better scope handling.

src/components/Match/MatchStory.jsx
NiQDeveloper: Easy done.

src/components/Match/MatchStory.jsx
```diff
@@ -156,6 +156,32 @@ const formatList = (items, noneValue = []) => {
   }
 };
 
+// evaluate the sentiment behind the message - rage, question, statement etc
+const evaluateSentiment = (event, lastMessage) => {
+  const { message, player, time } = event;
+  const sentiment = message.indexOf('?') !== -1 ? ['question'] : ['statement'];
```
drusepth: Small note: this'll render any message with a `?` anywhere as a question, no? I think this sounds reasonable (even though we can get messages like 
`"que? you suck m8", Ember Spirit asked`
tagged with "asked"), but worthwhile to point out explicitly. Maybe a future PR could break up sentence fragments so we can split text like the above into something like
`"que?" Ember Spirit asked, "you suck m8"`.  :+1:

src/components/Match/MatchStory.jsx
```diff
@@ -156,6 +156,32 @@ const formatList = (items, noneValue = []) => {
...
+
+  if (lastMessage && lastMessage.time + 130 > time) {
+    if (lastMessage.player_slot === player.player_slot) {
+      sentiment.push('continued');
+    } else if (lastMessage.key.indexOf('?') !== -1) {
```
drusepth: Also making explicit: looks like someone can't "respond" to another message unless that other message is a question, right? Again, I think this is fine; just making sure it's intended. 👍

src/lang/en-US.json
```diff
@@ -1065,5 +1065,35 @@
...
+  "statement_continued_normal": "added",
+
+  "question_response_long": "advised",
+  "question_response_shouted": "asked back, out of frustration",
+  "question_response_excited": "disputed",
+  "question_response_normal": "countered",
+
+  "question_continued_long": "propositioned",
+  "question_continued_shouted": "asked furiously",
+  "question_continued_excited": "lovingly asked",
```
drusepth: 💓

src/components/Match/MatchStory.jsx
NiQDeveloper: Breaking up sentence fragments sounds fantastic! Would require a bit of a refactor around ChatMessageEvent. I'll have a look into it, but definitely not in this PR. Don't want to bloat it too much.

The check was swapped explicitly mainly to handle the "?!" at the end of a sentence, or someone saying something along the lines of `is your faceless void going battlefury? lul`. Happy to change it back, maybe even change it to a regex check if there is a ? in the last set of punctuation. `\w(?:\W*(\?)\W*)$`

src/components/Match/MatchStory.jsx
NiQDeveloper: Yes, was intended. Thinking can split it out into 2 separate locale's - responses to questions, and responses in general. At first it was getting too many responses without the question check, but then I added in the time check which seems to cover most scenario's.

Issue #1164, opened 2017-09-12
incorrect time in story


howardchung:
reported via reddit:
https://www.reddit.com/r/DotA2/comments/6zj5g2/i_really_like_opendotas_attention_to_detail_but/?st=j7h4ne1g&sh=45bb6810

attn @drusepth

dchasepdx:
Hey,

I took a look at this one. It seems that for any match between 75 and 120 minutes it will return 'almost an hour'. I've got a quick fix that I'll make a pr for. It just changes 75 to 120 in the code below. 
```
else if (timeMinutes > 60 && timeMinutes <= 75) {
    // If the time is an hour to a quarter after, describe it as "over an hour"
    return `${strings.advb_over} ${strings.time_h}`;
  }
```

PR #1166, opened 2017-09-12
Fixes incorrect time in story

dchasepdx:
fixes [#1164](https://github.com/odota/ui/issues/1164)
Extends the range of timeMinutes so matches between 75 and 120 minutes are described as over an hour.


howardchung:
Thanks!  tagging @drusepth for informational purposes

drusepth:
Ah, whoops. Thanks for this. Looking at the code, it looks like there's one other small bug I included that'll make games longer than 2 hours also show up as "almost an hour". I'll open a quick PR for that as well. Thanks for fixing this though, @dchasepdx :)

PR #1170, opened 2017-09-14
Add sentiment analysis for xD, :D, lol

spdionis:
I thought it weird to see "[user] shouted." in the story mode for messages like "XD" so i made a small change to handle these cases.

If anyone has better ideas for these messages i will update them.

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -174,6 +174,8 @@ const evaluateSentiment = (event, lastMessage) => {
 
   if (message.split(' ').length > 10) {
     sentiment.push('long');
+  } else if (['XD', ':D', 'LOL'].indexOf(message.replace('?', '').toUpperCase()) !== -1) {
+    sentiment.push('laughed');
   } else if (message.toUpperCase() === message && /\w/.test(message)) {
     sentiment.push('shouted');
   } else if (/(\?|!|@|~|#|\$){2,}/.test(message)) {
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -174,7 +174,7 @@ const evaluateSentiment = (event, lastMessage) => {
 
   if (message.split(' ').length > 10) {
     sentiment.push('long');
-  } else if (['XD', ':D', 'LOL'].indexOf(message.replace('?', '').toUpperCase()) !== -1) {
+  } else if (['XD', ':D', 'LOL'].includes(message.replace('?', '').toUpperCase())) {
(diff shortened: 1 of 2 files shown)


src/components/Match/MatchStory.jsx
```diff
@@ -174,6 +174,8 @@ const evaluateSentiment = (event, lastMessage) => {
 
   if (message.split(' ').length > 10) {
     sentiment.push('long');
+  } else if (['XD', ':D', 'LOL'].indexOf(message.replace('?', '').toUpperCase()) !== -1) {
```
howardchung: es6 has `.includes()`

src/lang/en-US.json
```diff
@@ -1071,29 +1071,35 @@
...
   "statement_continued_shouted": "continued furiously",
   "statement_continued_excited": "continued",
   "statement_continued_normal": "added",
+  "statement_continued_laughed": "continued",
 
   "question_response_long": "advised",
   "question_response_shouted": "asked back, out of frustration",
   "question_response_excited": "disputed",
   "question_response_normal": "countered",
+  "question_response_laughed": "laughed wondering",
```
howardchung: wonderingly?

howardchung:
@mdiller @drusepth for review

src/lang/en-US.json
drusepth: how about "laughed heartily" here?

src/components/Match/MatchStory.jsx
spdionis: Doesn't work on Internet Explorer, will the toolchain handle it properly?

src/lang/en-US.json
spdionis: "laughed heartily" is not a response imo.
I thought about "wonderingly" but it sounds weird.

I could change it but i'm still not satisfied. We need a word to express a feeling of amazement I think?

drusepth: We could always just fall back on "laughed", even though it overlaps with another case.

```
"I'm playing with only one hand!", Invoker said.
"lol??", Ember Spirit laughed.
```

src/components/Match/MatchStory.jsx
howardchung: yes, babel should polyfill it for older browsers

mdiller:
Looks good!

src/lang/en-US.json
howardchung: I think just "laughed" is fine

spdionis:
Updated according to suggestions

PR #1193, opened 2017-09-26
ogp image for better previews

blukai:
Open Graph protocol image for better previews of posted links.
https://developers.facebook.com/docs/sharing/best-practices?locale=en_US#images


404.html
```diff
@@ -6,6 +6,7 @@
     <title>OpenDota</title>
     <meta name="explanation" content="Microsoft browsers require custom 404 pages to be at least 512 bytes in length, so this text is here as a placeholder in order to fill the necessary space to get to that number.">
     <link rel="icon" type="image/png" href="/assets/images/logo.png">
+    <meta property="og:image" content="/assets/images/logo-preview.jpg" />
```
howardchung: Any reason for jpg instead of png?

blukai: the size of jpg turned out to be less than png for this image

Issue #1171, opened 2017-09-14
Remove Image Alt Text in Story Mode


HellGate94:
even if the image fails to load the text is there anyway and it would show duplicated text. this duplicated text also shows when you try to copy paste. example:

"Radiant won a teamfight by trading NecrophosNecrophos and Death ProphetDeath Prophet for RubickRubick, Shadow ShamanShadow Shaman, and UnderlordUnderlord, resulting in a net worth increase of 314314 gold."

where it should be:

"Radiant won a teamfight by trading Necrophos and Death Prophet for Rubick, Shadow Shaman, and Underlord, resulting in a net worth increase of 314 gold."

howardchung:
I think lint rules require image elements to have alt text.

The images shouldn't be failing to load either, is this happening consistently?

HellGate94:
html wants an alt text but it doesn't care if its empty so alt=""

also they don't fail to load but if they would then you would see the alt text + the manual text so its a duplicate like when you try to copy paste

Steffo99:
This should be closed.

PR #1199, opened 2017-09-29
Fix <img> alt text in Story Mode

Steffo99:
This should fix the problem described in Issue #1171 by removing the alt text on the hero icons and by changing the alt text on the gold icon from `9999 gold` to ` gold`.


mdiller:
Looks good to me! Only suggestion I have is that maybe this should be added to the strings/lang file.

Steffo99:
I added the string to the en-US lang file, as you said.

PR #1200, opened 2017-09-29
eslint warnings fix

### LINKED ISSUE #1020: fix lint warnings
We're still getting some lint warnings that we might want to clean up at some point, like adding alt attributes to image elements.

farmasek:
#1020 
```(jsx-a11y/alt-text)```
```(no-shadow)``` 
```(no-console)```


howardchung:
More proptypes were added in #1204 so we have new lint warnings.  What's the best practice instead of using `object` and `array`?

PR #1203, opened 2017-10-01
Migration to styled-components

### LINKED ISSUE #1189: improve css maintainability
Some thoughts on improving css maintainability:

* We should go one of two routes with CSS
  * Simplicity: Make it barebones, nothing but css-loader and style-loader in the webpack config
  * Power: If we decide we want more power, variables etc, then use javascript to add that functionality (like JSS)
* This means we avoid having to use semi-official solutions like postcss, which also isn't compatible with data we might want to pass from javascript
* e.g. we have two ways of making things responsive right now, either via detecting the screen size in our components and passing the appropriate styling/className, or using css media queries
* If we use JSS then our existing logic to detect screen size can be reused
* Likewise it makes sense to have common variable definitions in JS.  Having those in JSON/JS would make it straightforward to read rather than having to run it through a css processor of some kind
  * e.g. https://github.com/odota/ui/blob/master/src/components/palette.css

Material-UI v1 will use JSS for styling:
https://material-ui-1dab0.firebaseapp.com/customization/css-in-js/

Feedback?
@mike-robertson 
@blukai 
@dartharaknus 
@gu3st

howardchung:
fixes #1189 

(Still WIP, but making early PR for feedback so if I'm doing something wrong I can fix it before continuing)

Slowly moving all components over to use styled-components.

* When we're done we should be able to remove the postcss loaders from webpack config
* We should also no longer have *.css files in the project
* Moving things defined in palette.css to constants.js
* styled-components lets us use JS to drop in strings into the css, either from constants or based on props based to the component


gu3st:
A pattern that we found in our project, was instead of having inline const's that defined the styled components, moving them into a "Styled.js" inside each folder (as well as the opportunity for shared styles?) that just exported each rule independently. 

There's also the opportunity to have some named files that export single (or a shared set of components) for commonly used components.

howardchung:
I'll keep that in mind if I find that we use the same set of styles repeatedly.  So far I'm mainly just finding reuses of particular CSS properties, e.g. `colorBlue`

PR #1204, opened 2017-10-01
Add prop types to some components

### LINKED ISSUE #145: Add propTypes for all component
Early on I stupidly decided we could skip this, but I think that (at least going forward), we should add propTypes to components we create and we should circle back at some point (at the same time as unit tests probably) to add propTypes into existing components.

murilodag:
fixes #145

### DIFF
--- a/src/components/Announce/Announce.jsx
+++ b/src/components/Announce/Announce.jsx
@@ -1,5 +1,6 @@
 /* global localStorage */
 import React from 'react';
+import PropTypes from 'prop-types';
 import RaisedButton from 'material-ui/RaisedButton';
 import strings from 'lang';
 import { connect } from 'react-redux';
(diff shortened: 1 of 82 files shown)


src/components/Announce/Announce.jsx
```diff
@@ -85,6 +94,16 @@ class RequestLayer extends React.Component {
   }
 }
 
+RequestLayer.propTypes = {
+  getPulls: PropTypes.func,
+  error: PropTypes.string,
+  loading: PropTypes.bool,
+  data: PropTypes.oneOfType([
```
howardchung: I think this is always an array, did you find a case where it isn't?

murilodag: Yes, in the render function there is a case where it gets data.items.

murilodag:
Hello @howardchung,

I fixed all the issues I manage to find using the console, but I think I didn't cover a lot of scenarios.

PR #1210, opened 2017-10-03
 add story_gold to pt-BR translation

mblodorn:
Add the story_gold according to #1199


howardchung:
Translations should go through the translation site: https://translate.opendota.com/

I don't think this new string is there yet, but it should show up when the translations PR is merged

PR #1218, opened 2017-10-04
Migrate xp networth graph to Recharts

farmasek:
Regarding #540 

Reworked it to look it dota-like

Still could use some polishing, but basic PoC is done, decide if its good enough to rework the rest.

Basically everything is possible, but scroll to zoom might be problem. But at least like [this](http://recharts.org/#/en-US/examples/BrushBarChart) it should be possible

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1214,8 +1214,7 @@
     "balanced-match": {
       "version": "0.4.2",
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-0.4.2.tgz",
-      "integrity": "sha1-yz8+PHMtwPAe5wtAPzAuYddwmDg=",
-      "dev": true
+      "integrity": "sha1-yz8+PHMtwPAe5wtAPzAuYddwmDg="
     },
     "base64-js": {
       "version": "1.2.0",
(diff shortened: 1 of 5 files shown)


howardchung:
Looks good to me!  I think it's a definite improvement on c3.  Should we migrate over the rest of the graphs?

There'll be:
* 4 graphs on Graphs match tab
* Trends graphs on player page
* MMR graph on player page
* Histograms on player page
* MMR distribution graph
* Explorer graphs (optional, can just disable for now)

howardchung:
The dependency is pretty big, but still smaller than c3 + d3 combined:
```
dotaconstants: 885.11 KB (12.3%)
material-ui: 788.42 KB (10.9%)
react-dom: 699.35 KB (9.71%)
recharts: 502.88 KB (6.98%)
c3: 350.1 KB (4.86%)
d3: 328.93 KB (4.57%)
core-js: 239.53 KB (3.33%)
```

howardchung:
I pushed these commits + some c3 removal to https://github.com/odota/ui/tree/recharts

howardchung:
Replace with https://github.com/odota/ui/pull/1221 which is on a local branch (easier for everyone to push to)

PR #1221, opened 2017-10-05
Recharts graphs (switch from c3)

### LINKED ISSUE #540: Better graph UI
I was wondering if you could use similar graphics libraries to the ones found on http://london.my-netdata.io/default.html

The Graphs page could benefit from these graphics, maybe a green line above median for radiant, red line below for dire (like the read/write graphs they have). Gauges could also be used for benchmark metrics, etc.

howardchung:
fixes #540


howardchung:
All done!  There are still a few tooltips that could be prettified (they're just white text right now)

howardchung:
Also better mobile support would be nice, right now the graphs just run off the screen.

PR #1226, opened 2017-10-07
update to airbnb16

howardchung:
Still WIP, a lot of indentation issues to manually fix


howardchung:
Latest jsx-a11y config also seems to require `href` attribute on Link elements, probably we should just disable this rule

PR #1230, opened 2017-10-07
Fix translation for zh-TW

siliconion:
Current zh-TW translation is merely copy of zh-CN. I've fixed all translation to traditional Chinese.


howardchung:
Translations should go through CrowdIn (https://translation.opendota.com) but I'll merge this since it's a bulk translation.  Might run into merge conflicts later though

PR #1231, opened 2017-10-07
Change port from 8080 to 3000 in README.md

Luchanso:
When I use `npm start`, server run on 3000 port


howardchung:
Yeah, I think this is what create-react-app does by default, although it will use the PORT env var if present.

PR #1235, opened 2017-10-07
Add story book

Luchanso:
I added [storybook](https://storybook.js.org/) for faster development

And add BurgerButton & ButtonGarden stories

image


package.json
```diff
@@ -50,9 +52,13 @@
     "seamless-immutable": "^7.1.2",
     "styled-components": "^2.2.0",
     "uuid": "^3.1.0",
-    "wordcloud": "^1.0.6"
+    "wordcloud": "^1.0.6",
+    "@storybook/react": "^3.2.12",
```
howardchung: should be in devdependencies

package.json
```diff
@@ -50,9 +52,13 @@
...
+    "@storybook/addon-actions": "^3.2.12",
+    "@storybook/addon-links": "^3.2.12"
   },
   "devDependencies": {
+    "babel-preset-es2015": "^6.24.1",
```
howardchung: don't think this is needed?

Luchanso: Yeah, because without it not work storybook

package.json
Luchanso: image

howardchung: OK, although it's a bit confusing since these dependencies don't end up in the webpack bundle

howardchung:
OK to merge or do you want to add more stories?

Luchanso:
I can add more stories later, in other PR, okey?

package.json
Luchanso: I want fix it, maybe in first time add /* eslint-disable import/no-extraneous-dependencies */ in stories files?

PR #1234, opened 2017-10-07
Moved "heading="Ability Build" to json file

tonp:

PR #1237, opened 2017-10-07
Moved string from overview.jsx to en-US.json

dhelic98:


howardchung:
Sorry, I guess everyone really wanted to do this, see #1236  and #1234

Issue #338, opened 2016-10-25
Load ability, item tooltips from api instead of dotaconstants


blukai:
https://github.com/odota/ui/blob/master/src/components/Visualizations/inflictorWithValue.jsx

howardchung:
If we don't have a performance/load time issue then we don't necessarily have to do this.  However, if we want to make the tooltips available in multiple languages, we will probably want to do this because loading multiple languages of abilities will get big really quickly.

howardchung:
Did a quick check, abilities data is currently about 720KB uncompressed and 120KB compressed.  As of this writing the bundle size is 923KB, so this is a pretty significant chunk of the bundle.

howardchung:
Doing this will also enable translated data, since the client can just request data in the desired language.

howardchung:
I also recently did this for items. Heroes is left (but is also the smallest of the three), plus being heavily used means we may not benefit as much from the code splitting.

PR #1241, opened 2017-10-08
implement abilities in redux store

howardchung:
We only use the abilities data for rendering ability icons and tooltips.

The idea is that it's a really big bundle (740KB uncompressed, 101KB compressed), so by loading it separately we might get better performance.

I implemented this by making the ability fetch a separate API call to the blob on GitHub and storing the response in the Redux store.  Then I used `connect` to enable the InflictorWithValue component to read from it.

Any feedback?

Improvements:
* Avoid the flash of default auto attack icon that happens before the ability data is loaded
* Point to a custom API data source that supports requesting data in the correct language #338 
* Not optimal: This causes the data version to be tied to what's on master rather than the version shipped with dotaconstants


src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -217,28 +218,44 @@ export default (inflictor, value, type, ptooltip) => {
...
+InflictorWithValueComp.propTypes = {
+  inflictor: PropTypes.string,
+  value: PropTypes.string,
+  type: PropTypes.string,
+  ptooltip: PropTypes.shape({}),
+  abilities: PropTypes.shape({}),
+};
+
+const mapStateToProps = state => ({
+  abilities: state.app.abilities.data || {},
```
howardchung: don't assign a default object here, this causes react to re-render the component all the time

howardchung:
Note: dynamic import may be a better solution to this problem.

PR #1246, opened 2017-10-09
Graph range fix, MMR brush implemented

farmasek:
#### Implemented graph range
- MMR graph set to minimal MMR - 150 and maximal +150
- Fixed Date displayed on tooltip to localdatestring
- Fixed IMG of hero occasionally crashing (could not reproduce so i put it in try catch and return empty div when fail, it probably happen when you move fast so img fetch is interrupted with another)
- Trends set to auto - auto (will start and end how recharts thinks its the best )
##### Experimental Brush on MMR
- selector to select range for date in mmr graph (kinda zoom functionality) on local machine its performance heavy when selecting so if it will be issue it can be removed.

If i forgot some graphs let me know.

### DIFF
--- a/src/components/Visualizations/Graph/TrendGraph.jsx
+++ b/src/components/Visualizations/Graph/TrendGraph.jsx
@@ -72,7 +72,18 @@ const TooltipStylesDiv = styled.div`
     font-size: ${constants.fontSizeCommon};
   }
 `;
-
+const getHeroImg = (data) => {
+  try {
+    return (<img
+      className="heroImg"
+      src={`${process.env.REACT_APP_API_HOST}${heroes[data.hero_id]
+    .img}`}
+      alt=""
+    />);
+  } catch (e) {
+    return <div />;
+  }
+};
 const TrendTooltipContent = ({ payload, name }) => {
   const data = (payload[0] || {}).payload;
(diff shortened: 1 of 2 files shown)


src/components/Visualizations/Graph/TrendGraph.jsx
```diff
@@ -72,7 +72,18 @@ const TooltipStylesDiv = styled.div`
     font-size: ${constants.fontSizeCommon};
   }
 `;
-
+const getHeroImg = (data) => {
+  try {
+    return (<img
+      className="heroImg"
+      src={`${process.env.REACT_APP_API_HOST}${heroes[data.hero_id]
```
howardchung: Probably the cause of this is just not having data.hero_id defined, so we should just set a default hero object so we don't try to read .img of undefined

howardchung:
Some additional fixes I want to make for this, but I'll just merge this and make my fixes on top

src/components/Visualizations/Graph/TrendGraph.jsx
howardchung: found the issue is actually due to payload[0], where payload is not defined, so I just added an additional check for that

PR #1260, opened 2017-10-13
Romanian translation

grasozauru:
Pretty much everything is translated, but chatwheel sentences (which i'd say they're better off in english) and things that i did not find a way to make them sound good in romanian


howardchung:
Thanks!  Normally translations should go through CrowdIn (https://translate.opendota.com) to avoid merge conflicts.

PR #1265, opened 2017-10-17
Improve manifest.json by adding several resolution of icons

superoo7:
To achieve better experience for mobile user with several resolution of icons (Progressive Web App Approach)


superoo7:
For icon generation I would suggest using this link [https://app-manifest.firebaseapp.com]❨https://app-manifest.firebaseapp.com❩

PR #1268, opened 2017-10-18
Cauchy|Sort players before rendering crosstable.

### LINKED ISSUE #1196: Bug: combat grid has heroes on wrong side
some of the dire heroes are with radiant and vice versa

https://www.opendota.com/matches/3465425218/combat

image

CauchyAbel:
Fixes #1196. Please review @blukai @howardchung


coreymaher:
Wouldn't it be better to fix whatever is sorting in the first place?
It should probably clone the array before sorting so unintended things like this don't happen.

howardchung:
yeah, I agree that's probably the better fix.  @CauchyAbel do you want to take a look at where the Table component is doing the sorting and make sure it creates a new copy of the array?

CauchyAbel:
@howardchung Sounds good, I will take a look.

CauchyAbel:
Basically I just deep cloned the players array before rendering the Laning tab. Meanwhile the original match for everywhere else.

coreymaher:
I wonder if the `defaultSort` function should be changed to return a copy of the passed in array instead of sorting in place?
 https://github.com/odota/ui/blob/beb2b02b99604e31c439a90ebe3de67c168cec93/src/utility/index.jsx#L372

Or maybe the array should be cloned before calling `defaultSort`? https://github.com/odota/ui/blob/beb2b02b99604e31c439a90ebe3de67c168cec93/src/components/Table/Table.jsx#L111

src/components/Match/Laning/index.jsx
```diff
@@ -20,17 +20,17 @@ class Laning extends React.Component {
     this.setState({ ...this.state, selectedPlayer: playerSlot });
   }
   render() {
-    const { match } = this.props;
+    const players = [...this.props.match.players];
```
howardchung: this won't be sufficient by itself since this is the not the only table that sorts.  Probably this needs to occur within the Table component's sorting logic.

howardchung:
yeah I think those approaches would work @coreymaher 

I don't know if JS has a function like Python's `sorted()` which returns a new array.  We might have to shallow clone it before sorting.

CauchyAbel:
OK it seems this change could be as simple as this....

src/components/Match/Laning/index.jsx
```diff
@@ -20,17 +20,17 @@ class Laning extends React.Component {
     this.setState({ ...this.state, selectedPlayer: playerSlot });
   }
   render() {
-    const players = [...this.props.match.players];
+    const { match } = this.props;
```
howardchung: are these changes needed?

CauchyAbel: Oh this change is to revert the code back to original state. If you check out all the change, it was only a copy of array xD.

dartharaknus:
Squash all the commits into one to avoid this kind of confusion 👍

Issue #1269, opened 2017-10-18
Sum tables in the "Farm" tab


coreymaher:
None of the tables in the Farm tab have totals. I think that would be a useful addition.

PR #1270, opened 2017-10-18
Sum all tables in the Farm tab. Added sumValueFn to allow summing on …

coreymaher:
…columns which do not map to a simple field.

Addresses #1269

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -511,7 +511,7 @@ export const lastHitsTimesColumns = (match) => {
       sortFn: row => row.lh_t && row.lh_t[minutes],
       displayFn: row => `${row.lh_t[minutes]} (+${row.lh_t[minutes] - row.lh_t[minutes - (bucket / 60)]})`,
       relativeBars: true,
-      sumValueFn: row => ((row.lh_t) ? row.lh_t[minutes] : 0),
+      sumFn: row => ((row.lh_t) ? row.lh_t[minutes] : 0),
     });
   }
   return cols;
@@ -848,7 +848,7 @@ export const goldReasonsColumns = [heroTdColumn].concat(Object.keys(strings).fil
   sortFn: row => (row.gold_reasons ? row.gold_reasons[gr.substring('gold_reasons_'.length)] : 0),
   displayFn: (row, column, value) => value || '-',
   relativeBars: true,
-  sumValueFn: row => (row.gold_reasons ? (row.gold_reasons[gr.substring('gold_reasons_'.length)] || 0) : 0),
+  sumFn: row => (row.gold_reasons ? (row.gold_reasons[gr.substring('gold_reasons_'.length)] || 0) : 0),
 })));
 
 export const xpReasonsColumns = [heroTdColumn].concat(Object.keys(strings).filter(str => str.indexOf('xp_reasons_') === 0).map(xpr => ({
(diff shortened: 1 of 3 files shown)


coreymaher:
On a few of the columns I was unable to use the default `sumFn` property since that assumes each row contains a value that matches the `field` of the column. This isn't the case for some of the computed columns, for example, the last hits at 5 minute increments.

Unfortunately the `sumValueFn` I created is essentially a copy of the `sortFn` since both these features need the column to determine the value. I'm not sure if we can refactor things so both these features can use the same function, or if sorting and summing will eventually need different logic.

howardchung:
To avoid adding a new property, perhaps you could have the Table check to see if sumFn is a function, and if so use it over the sortFn?

coreymaher:
@howardchung I just pushed up a new commit to reuse the sumFn property instead of creating a new sumValueFn property.

src/components/Match/matchColumns.jsx
```diff
@@ -838,6 +848,7 @@ export const goldReasonsColumns = [heroTdColumn].concat(Object.keys(strings).fil
   sortFn: row => (row.gold_reasons ? row.gold_reasons[gr.substring('gold_reasons_'.length)] : 0),
   displayFn: (row, column, value) => value || '-',
   relativeBars: true,
+  sumFn: row => (row.gold_reasons ? (row.gold_reasons[gr.substring('gold_reasons_'.length)] || 0) : 0),
```
howardchung: isn't this the same as sortFn?

The || 0 shouldn't be necessary I think

src/components/Table/Table.jsx
```diff
@@ -219,7 +219,8 @@ class Table extends React.Component {
                 <MaterialTableRow>
                   {columns.map((column, colIndex) => (
                     <MaterialTableRowColumn key={`${colIndex}_sum`} style={{ color: column.color }}>
-                      {column.sumFn && (column.field !== 'life_state_dead' ?
+                      {column.sumFn && typeof column.sumFn === 'function' && abbreviateNumber(data.map(column.sumFn).reduce(sum, 0))}
```
howardchung: the only thing that differs is the part inside `data.map`.  Optimally the if condition would branch there so that we minimize duplicated code.

That might require moving this code out into a function outside the JSX so it's easier to write

src/components/Match/matchColumns.jsx
coreymaher: Pretty much!
The`|| 0` is needed since the value could be undefined. If it is, then the `.reduce(sum, ...)` doesn't work and always returns undefined.
`sortFn` could probably use the same logic, or something similar, but it's probably not as noticeable since undefined probably sorts similarly to `0`.

howardchung: In that case maybe add the || 0 to the sortFn and then you can replace this line with `true` to reduce duplication?

coreymaher:
@howardchung I took another crack at it and made the follow changes:
- Moved rendering the totals row to a new function `renderSumRow` so I could have a scope to define variables
- Made the `sumFn` column property a function which is passed into reduce so columns have more control over how the sums are calculated (this allowed me to actually sum the "Unit Kills - Other" column)
- Created a `displaySumFn` column property function which allows a column to override how it's sum is displayed. This allowed me to remove the custom logic for death time from `Table.jsx` and allowed me to display the totals for the "Unit Kills - Other" column.

Let me know what you think.

howardchung:
I actually preferred the old signature for consistency with sortFn and displayFn.

* You should still be able to achieve the same effect by resolving a multi-valued cell into a single value that's a sum of the individual values in sumFn.
* Instead of displaySumFn perhaps just have a new property or something that gets passed to the Table as a string.  Displaying as time is probably something we might need to reuse elsewhere.

coreymaher:
@howardchung I saw you merged this. Was there any other changes you wanted me to make? I know it wasn't implemented exactly how you were hoping.

PR #1272, opened 2017-10-20
Fixed incorrect KDA name

builder-247:


howardchung:
This is intentional, although it does seem to be confusing people

PR #1278, opened 2017-10-25
Put BrowserStack logo back so it shows up in README.md

asmaloney:
This was (inadvertently?) removed in c863153177cba616af352b1d0435ef4f43bd75c0

Also resized and optimized the PNG.

I am not affiliated with BrowserStack - I just noticed the image wasn't showing up properly. So this is a drive-by PR.


howardchung:
Thanks!

PR #1281, opened 2017-10-25
Add relative negative bar

### LINKED ISSUE #1263: Invert color bar logic for time dead
I would like to suggest an enhancement.

Considering the match details -> Performance -> color of the time dead bars.
Assuming a good time dead is short, and assuming a green color indicates a good performance, I would suggest to enhance the experience by turning the scale around.
Thereby short death would be shown green and long death red.

Luchanso:
I added relative negative bar, close #1263

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -170,7 +170,16 @@ class Table extends React.Component {
                         const altValue = typeof bars === 'function' && percentBarsWithValue ? bars(row) : null;
                         let valEl = null;
                         let barPercentValue = 0;
-                        if (relativeBars || relativeNegativeBars) {
+                        if (relativeNegativeBars) {
+                          const min = getColumnMin(data, field, getValue);
+                          const max = getColumnMax(data, field, getValue);
+
+                          barPercentValue = (1 - ((value - min) / (max - min))) * 100;
+
+                          valEl = displayFn
+                            ? displayFn(row, column, value, index, barPercentValue)
+                            : <span>{value}</span>;
+                        } else if (relativeBars) {
                           // Relative bars calculates the max for the column
                           // and gets the percentage of value/max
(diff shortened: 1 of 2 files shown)


Luchanso:
Ahhh... it is not correct

Luchanso:
I fix problems

howardchung:
What does this new bar type do?

#1263 seems to be referring to changing the color scheme, not adding a new bar type.

Luchanso:
Okay, I'll add other color style

чт, 26 окт. 2017 г. в 6:48, Howard Chung <notifications@github.com>:

src/components/Table/Table.jsx
```diff
@@ -164,12 +165,21 @@ class Table extends React.Component {
...
 
                       let fieldEl = null;
-                      const bars = relativeBars || percentBars || percentBarsWithValue;
+                      const bars = relativeBars || relativeNegativeBars || percentBars || percentBarsWithValue;
                       if (bars) {
                         const altValue = typeof bars === 'function' && percentBarsWithValue ? bars(row) : null;
                         let valEl = null;
                         let barPercentValue = 0;
-                        if (relativeBars) {
+                        if (relativeNegativeBars) {
```
howardchung: Instead of redoing all this logic, I think we should just pass a single property down called "invertedScaling" and flip the barPercentValue

Luchanso:
I was try just flipping, but it's not correct

чт, 26 окт. 2017 г. в 12:24, Howard Chung <notifications@github.com>:

howardchung:
You might have to pass the invert property down to the TablePercent component since it seems that this code does not deal with colors at all.

Luchanso:
Yes, you right.

чт, 26 окт. 2017 г. в 12:32, Howard Chung <notifications@github.com>:

Luchanso:
I will do it in other PR

PR #1284, opened 2017-10-30
Add inverse color in Percent.jsx

### LINKED ISSUE #1263: Invert color bar logic for time dead

Luchanso:
Fix color bars for Death, close #1263

image

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -561,6 +561,7 @@ export const performanceColumns = [
     sortFn: true,
     displayFn: (row, col, field) => formatSeconds(field) || '-',
     relativeBars: true,
+    inverseRelativeBars: true,
     sumFn: true,
   },
   {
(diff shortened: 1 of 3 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -561,6 +561,7 @@ export const performanceColumns = [
     sortFn: true,
     displayFn: (row, col, field) => formatSeconds(field) || '-',
     relativeBars: true,
+    inverseRelativeBars: true,
```
howardchung: can you rename this to "invertBarColor" which is more descriptive?

Luchanso: Done

PR #1286, opened 2017-10-31
Ignore unused items that are whitelisted or purchased at the end of a game

### LINKED ISSUE #974: Match analysis shouldn't count some items as unused
The following cases probably shouldn't be accounted as "unused items":
* Any item that was in inventory less than a minute, especially those purchased right before the game end
  * Example: https://www.opendota.com/matches/3208159551/analysis - I as Phoenix purchased Glimmer, disassembled it and sold immediately, but it was counted as 2 unused items
* Quelling blade
* RoB/Aquila (might be purchased for pushing purposes specifically)
* Linken's on cores? (a passive effect is usually enough)

NickelOz:
fixes #974 

First off, thanks for all your work in making such an awesome site, I've been using it for the past few years and it's made accessing and viewing all my dota stats much easier and enjoyable.

I've done a little bit of work to address #974 .

The only thing I'm uncertain about is the format of times for events in the game log, should I assume that they are in seconds?

I'm quite happy to make any changes needed. I'm fairly new to contributing so any feedback/improvements are more than welcome!

### DIFF
--- a/src/actions/analyzeMatch.js
+++ b/src/actions/analyzeMatch.js
@@ -156,7 +156,7 @@ export default function analyzeMatch(match, _pm) {
     },
     // unused item actives (multiple results?)
     unused_item(m, pm) {
-      whitelistUnusedActiveItems = [
+      const whitelistUnusedActiveItems = [
         'quelling_blade',
         'faerie_fire',
         'enchanted_mango',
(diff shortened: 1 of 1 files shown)


howardchung:
LGTM

Issue #1299, opened 2017-11-05
vision tab still uses 7.06 map


tobyfloyd:
the vision tab still uses the  7.06 mini map instead of the new 7.07 map.

howardchung:
do we have a new map image to use?

tobyfloyd:
yeah there is a new one.


you can find for example at the dota 2 wiki:


https://dota2.gamepedia.com/Minimap
https://dota2.gamepedia.com/File:Minimap_7.07.png





-----Ursprüngliche Mitteilung----- 
Von: Howard Chung <notifications@github.com>
An: odota/ui <ui@noreply.github.com>
Cc: tobyfloyd <rrealtobi@aol.com>; Author <author@noreply.github.com>
Verschickt: So, 5. Nov 2017 11:46
Betreff: Re: [odota/ui] vision tab still uses 7.06 map (#1299)



do we have a new map image to use?

PR #1300, opened 2017-11-06
updated map from 7.06 to 7.07 map

maxlefebvre:
Fixes issue #1299 

I hope I did this right, couldn't find a contributing.md


blukai:
But it will just replace map for previous patch. Match object has `patch` field and [dotaconstants](https://github.com/odota/dotaconstants/blob/master/build/patch.json) contains patch info, so I think we should use this.

maxlefebvre:
Looks I overlooked that issue. I could try to get the patch field if you think that's better. Another solution I think would work is that I add some consts for the previous maps and add those minimaps into the DotaMap folder like this. What do you think?

`// 10/31/2017 @ 5:00pm (UTC) - patch 7.07
const is707 = unixDate => unixDate > 1509426000;
// 12/13/2016 @ 5:00pm (UTC) - patch 7.00
const is700 = unixDate => unixDate > 1481648400 && unixDate < 1509426000;
// 12/16/2015 @ 5:00pm (UTC) - patch 6.86
const is686 = unixDate => unixDate > 1450242000 && unixDate < 1481648400;
// 09/25/2014 @ 5:00pm (UTC) - patch 6.82
const is682 = unixDate => unixDate > 1411621200 && unixDate < 1450242000;
// Before 09/25/2014 @ 5:00pm (UTC) - pre patch 6.82
const isPre682 = unixDate => unixDate < 1411621200;

// TODO make this a generic function that can determine which of multiple maps to use
const getUrl = (startTime) => {
  if (startTime === null || is707(startTime)) {
    return '/assets/images/dota2/map/detailed_707.png';
  }
  else if (startTime === null || is700(startTime)) {
    return '/assets/images/dota2/map/detailed_700.png';
  }
  else if (startTime === null || is686(startTime)) {
    return '/assets/images/dota2/map/detailed_686.png';
  }
  else if (startTime === null || is682(startTime)) {
    return '/assets/images/dota2/map/detailed_682.png';
  }
  else if (startTime === null || isPre682(startTime)) {
    return '/assets/images/dota2/map/detailed_pre682.png';
  }
  return '/assets/images/dota2/map/detailed.png';
};`

howardchung:
Yes, I believe that's the better way to go.  You can probably reduce the amount of code with a while loop with breaks, and also the patch date data is available in the dotaconstants files so you don't need to manually type in the times for each.

maxlefebvre:
Unfortunately, with school I'm not finding time to do this anymore, so ill push my changes which is the alternate/less efficient solution. You guys can either close the PR and let someone else implement it or merge it as a temp fix, it's up to you guys. Thanks

PR #1310, opened 2017-11-11
Determine which dota map to display from the patch the match was played in

### LINKED ISSUE #1299: vision tab still uses 7.06 map

jayantak:
fixes #1299


src/components/DotaMap/index.jsx
```diff
@@ -9,15 +10,26 @@ const setMapSizeStyle = (width, maxWidth) => ({
...
-const isPost700 = unixDate => unixDate > 1481648400;
+const dotaMaps = [
+  { patchName: '7.07', mapImage: '/assets/images/dota2/map/detailed_707.png' },
+  { patchName: '7.00', mapImage: '/assets/images/dota2/map/detailed_700.png' },
+  { patchName: '6.86', mapImage: '/assets/images/dota2/map/detailed_686.png' },
+  { patchName: '6.82', mapImage: '/assets/images/dota2/map/detailed_682.png' },
+  { patchName: '6.70', mapImage: '/assets/images/dota2/map/detailed_pre682.png' },
+];
+
+const defaultMap = '/assets/images/dota2/map/detailed.png';
```
howardchung: I think this is the pre_700 map from before.  Now we have more accurate maps.

I think the default should be the latest map `dotaMaps[0].mapImage`, and we should delete detailed.png

jayantak: Okay, sounds good

jayantak: Done

src/components/DotaMap/index.jsx
```diff
@@ -9,15 +10,25 @@ const setMapSizeStyle = (width, maxWidth) => ({
...
-// TODO make this a generic function that can determine which of multiple maps to use
 const getUrl = (startTime) => {
-  if (startTime === null || isPost700(startTime)) {
-    return '/assets/images/dota2/map/detailed_700.png';
+  if (startTime == null) return dotaMaps[0];
+  for (let i = 0; i < dotaMaps.length; i += 1) {
+    if (startTime >= patchDate[dotaMaps[i].patchName]) return dotaMaps[i].mapImage;
   }
-  return '/assets/images/dota2/map/detailed.png';
+  return dotaMaps[0];
```
howardchung: .mapImage

jayantak: Good catch, fixed

Issue #1294, opened 2017-11-01
Handle new heroes in Combat and Log Tabs 


ABigg:
Clicking the Combat or the Log tab in matches where the new heroes have been played results in this:

![opendotesbug](https://user-images.githubusercontent.com/33290444/32283778-f260e30e-beea-11e7-8a84-1dd3ee1a19b6.png)

The issue can also be replicated on Chrome:

![opendotabug2](https://user-images.githubusercontent.com/33290444/32283880-39ae9d50-beeb-11e7-8cdd-9ed8a3ec3851.png)

It doesn't seem to happen in matches where the new heroes were not played.

howardchung:
Issue is fixed by adding new heroes to hero list, but we should still handle this without crashing

howardchung:
This should not be too hard to do, although now that the data for the new heroes has been added, you might need to repro it by faking some data or modifying the heroes object imported from dotaconstants to simulate not having a hero object for a new ID.

PR #1316, opened 2017-11-17
fix to #1294

Argn0:
I'm a total noob. I don't expect these change to get merged but I had fun anyway :D

### DIFF
--- a/src/components/Match/CrossTable.jsx
+++ b/src/components/Match/CrossTable.jsx
@@ -16,6 +16,8 @@ import constants from '../constants';
 
 const teamIconStyle = { width: '30px', height: '30px' };
 
+const default_hero = { name: 'Unknown Hero' };
+
 const CrossTable = ({
   match,
   field1,
--- a/src/components/Match/CrossTable.jsx
+++ b/src/components/Match/CrossTable.jsx
@@ -67,7 +67,7 @@ const CrossTable = ({
           let ptotal2 = 0;
 
           match.players.slice(match.players.length / 2, match.players.length).forEach((player2) => {
-            const hero2 = heroes[player2.hero_id];
+            const hero2 = heroes[player2.hero_id] || default_hero;
             ptotal1 += (player[field1] && hero2.name in player[field1]) ? player[field1][hero2.name] : 0;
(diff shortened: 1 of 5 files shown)


src/components/Match/MatchStory.jsx
```diff
@@ -46,7 +46,7 @@ const TeamSpan = isRadiant => (
 
 // Modified version of PlayerThumb
 const PlayerSpan = (player) => {
-  if (!player || !heroes[player.hero_id]) {
+  if (!player || (!player && !heroes[player.hero_id])) {
```
howardchung: I don't think this change is needed

src/components/Match/CrossTable.jsx
```diff
@@ -16,6 +16,8 @@ import constants from '../constants';
 
 const teamIconStyle = { width: '30px', height: '30px' };
 
+const default_hero = { name: 'Unknown Hero' };
```
howardchung: Move the english string to langs/en-US.json (don't need to do the other languages manually)

howardchung:
do npm run lint to check for lint issues

src/components/Match/MatchStory.jsx
Argn0: I don't exactly understand what's happening but if I remove this change then the Story tab crashes.

example: matches/3564865624/story

src/components/Visualizations/Graph/MatchGraph.jsx
```diff
@@ -188,7 +188,7 @@ class PlayersGraph extends React.Component {
               {match.players.map((player) => {
                 const hero = heroes[player.hero_id] || {};
                 const playerColor = playerColors[player.player_slot];
-                const isSelected = (hoverHero === heroes[player.hero_id].localized_name);
+                const isSelected = (hoverHero === heroes[player.hero_id] && heroes[player.hero_id].localized_name);
```
rodrigofepy: This line is a string comparison, I think you should move the object validation in front of the comparison.
Should look like this: `heroes[player.hero_id] && (hoverHero === heroes[player.hero_id].localized_name);`

Argn0: thanks. I changed it as you suggested.

howardchung:
combining with https://github.com/odota/ui/pull/1318

PR #1317, opened 2017-11-17
Hide home page login button if user is logged in

amalik12:
If the user is already logged in, the "Login for automatic replay parsing" button disappears.
Before:

After:

### DIFF
--- a/src/components/Home/Buttons.jsx
+++ b/src/components/Home/Buttons.jsx
@@ -7,13 +7,6 @@ import strings from 'lang';
 import { IconSteam } from 'components/Icons';
 import { ButtonsDiv } from './Styled';
 
-const mapStateToProps = (state) => {
-  const { data } = state.app.metadata;
-  return {
-    user: data.user,
-  };
-};
-
 const Buttons = ({ user }) => (
   <ButtonsDiv>
     {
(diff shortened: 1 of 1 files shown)


src/components/Home/Buttons.jsx
```diff
@@ -1,19 +1,31 @@
...
 import { Link } from 'react-router-dom';
+import PropTypes from 'prop-types';
 import FlatButton from 'material-ui/FlatButton';
+import { connect } from 'react-redux';
 import strings from 'lang';
 import { IconSteam } from 'components/Icons';
 import { ButtonsDiv } from './Styled';
 
-export default () => (
+const mapStateToProps = (state) => {
```
howardchung: I like to put mapstatetoprops at the bottom (next to the connect) but that's just style

src/components/Home/Buttons.jsx
```diff
@@ -37,3 +49,18 @@ export default () => (
     </div>
   </ButtonsDiv>
 );
+
+Buttons.propTypes = {
+  user: PropTypes.shape({}),
+};
+
+class RequestLayer extends React.Component {
```
howardchung: is there any reason to have this requestlayer here?  you should be able to connect the buttons component directly.

amalik12:
made the changes!

PR #1324, opened 2017-11-19
Added new feature:  filter heroes by "at least __ games played"

### LINKED ISSUE #1157: As a user, I should be able to filter out heroes with low matches played while reviewing hero performance.
When filtering my heroes played for ranked matches only, for example, I want to sort by winrate but there are too many heroes for which I have fewer than 5 games played.  I should be able to check a box or enable a filter to remove heroes for which I have fewer than some number of games played (e.g. 5, 10, 15).

wanderrful:
Fixes #1157

I created issue #1157 because I wanted to review my hero win rates and determine my consistently best heroes... but the data was always distorted by heroes for which I have relatively few games played.  

I've been working on and off for the past couple of months on trying to figure out how to implement this feature.  Recently came up with a quick-and-dirty solution that seems to work perfectly fine on localhost.  I hope that you guys will accept my pull request here so that I can finally say that I'm good enough with this web dev stuff to be able to contribute to Open Source projects!

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -108,9 +108,6 @@ class Table extends React.Component {
       currentPage: this.state.currentPage - 1,
     });
   }
-  getURLParameter(name) {
-    return decodeURIComponent((new RegExp('[?|&]' + name + '=' + '([^&;]+?)(&|#|;|$)').exec(window.location.search) || [null, ''])[1].replace(/\+/g, '%20')) || null;
-  }
   render() {
     const {
       columns,
@@ -122,16 +119,10 @@ class Table extends React.Component {
       pageLength = 20,
     } = this.props;
     const {
-      sortState, sortField, sortFn, currentPage
+      sortState, sortField, sortFn, currentPage,
     } = this.state;
     const dataLength = this.props.data.length;
     let { data } = this.props;
-    
-    let filter_gamesplayed_amount = this.getURLParameter("games_played");
-    if (filter_gamesplayed_amount) {
-      data = data.filter( hero => hero.games > filter_gamesplayed_amount );
-    }
-
     if (maxRows && maxRows <= dataLength) {
       data = data.slice(0, maxRows);
     }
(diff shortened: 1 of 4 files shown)


src/lang/en-US.json
```diff
@@ -184,6 +184,12 @@
   "filter_last_month": "Last month",
   "filter_last_3_months": "Last 3 months",
   "filter_last_6_months": "Last 6 months",
+  "filter_with_games_played": "Games played",
```
howardchung: Possibly change this to "At least this many games played" and then each item in the list doesn't need its own string (you can just use the number 5/10/15...)

src/components/Table/Table.jsx
```diff
@@ -108,6 +108,9 @@ class Table extends React.Component {
       currentPage: this.state.currentPage - 1,
     });
   }
+  getURLParameter(name) {
```
howardchung: I suggest adding a filter option to odota/core called "having" or something (the SQL equivalent to this filter), which ties in better with the existing filter framework

wanderrful:
I have reverted the implementation of the filter and only left the front-end of it.  After attempting a recommended change over at odota/core, I've made a partner Pull Request here:  https://github.com/odota/core/pull/1563

src/lang/en-US.json
```diff
@@ -184,6 +184,12 @@
   "filter_last_month": "Last month",
   "filter_last_3_months": "Last 3 months",
   "filter_last_6_months": "Last 6 months",
+  "filter_with_games_played": "At least this many games played",
```
howardchung: suggestion from earlier about not having an individual string for each option still stands

howardchung: actually you can reuse the string from the Explorer:
"explorer_having"

wanderrful:
All of the other filters in TableFilter.config.js still have "text" members, so I just replaced the localized string text with string literals as you asked.  Wasn't sure if it was okay to just have only the "value" member by itself.  Also deleted the relevant localized text entries.

wanderrful:
Removed the localized string and replaced the filter name with the one you mentioned in your comment.

PR #1335, opened 2017-11-24
toggle to show/hide small purchases

### LINKED ISSUE #1332: consider adding toggle to show/hide small purchases
The items tab has always been pretty crowded.

Consider adding a toggle that hides small items like TP scrolls and wards, and default it to on.

Example:
https://www.opendota.com/matches/3577473492/purchases

Argn0:
fixes #1332 

consumables in starting items are of interest I think


src/components/Match/matchColumns.jsx
```diff
@@ -462,15 +462,17 @@ export const fantasyColumns = [
   },
 ].concat(fantasyComponents);
 
-export const purchaseTimesColumns = (match) => {
+export const purchaseTimesColumns = (match, showCommItems) => {
+  // IDs of common consumable items
+  const commCon = showCommItems ? [] : [46, 216, 44, 42, 43, 38, 257, 265, 188, 16];
```
howardchung: I suggest using the names for greater readability

src/components/Match/matchColumns.jsx
```diff
@@ -462,15 +462,17 @@ export const fantasyColumns = [
...
   const cols = [heroTdColumn];
   const bucket = 300;
   for (let i = 0; i < match.duration + bucket; i += bucket) {
     const curTime = i;
     cols.push({
       displayName: `${curTime / 60}'`,
       field: 'purchase_log',
-      displayFn: (row, column, field) =>
+      displayFn: (row, column, field) => [
```
howardchung: what's the purpose of adding [ here?

src/components/Match/matchColumns.jsx
Argn0: shoul I put the array with item names outside the function?

howardchung: sure you can put it in utility file if you think it might be reused

Argn0: ok done

howardchung:
check lint `npm run lint`

Argn0:
ok I think I've fixed everything

src/utility/index.jsx
```diff
@@ -592,3 +592,18 @@ export const wilsonScore = (up, down) => {
     1 + (z * z / n)
   );
 };
+
+export const commonConsumableItems = [
+  'Town Portal Scroll',
```
howardchung: actually I mean the internal item name, like `tpscroll`, since those are unlikely to change.

Argn0: items already had a qual property. I'm so blind :)

howardchung:
Opinions on hiding/showing the starting items?  It might be confusing to the user that they don't count as consumables for display purposes

Argn0:
Maybe we could rename time 0  to 'Starting Purchases'

Argn0:
I remove consumables from starting items. I agree it was kinda confusing.

Argn0:
I removed consumables from starting items, because I agree it was kinda confusing.

Issue #1140, opened 2017-08-27
Make log similar to chat


blukai:

Luchanso:
Can you write more information about it?

blukai:
@Luchanso 
Create the "Message" component and display log events similar to messages on the chat tab, like this:
![image](https://user-images.githubusercontent.com/20866892/32013716-9cf0ba5a-b9c4-11e7-8c5e-039805910a7f.png)
Also maybe make filters (kills, objectives, runes) like on the chat tab

DomitriusAnthony:
Is this still open? Would like to try!

Luchanso:
@DomitriusAnthony Yeah, u can! 👍Maybe just modify this chat?
![image](https://user-images.githubusercontent.com/2098777/38411605-1c6dc3b8-3990-11e8-953e-83e1797cef4b.png)

What you think @howardchung?

howardchung:
Sounds good!  You can probably reuse the same component that chat uses.

ChrisMissal:
I believe this one is complete, the log looks great!

![image](https://user-images.githubusercontent.com/67798/42465021-6eff6b84-8370-11e8-91c4-5aa264d38032.png)

PR #1339, opened 2017-11-25
improvements to match log

Argn0:
the blood splatter image is from here and is free for commercial use
https://pixabay.com/en/ink-red-splatter-abstract-paint-303244/


src/components/Match/MatchLog.jsx
```diff
@@ -118,6 +119,69 @@ const logColumns = [heroTdColumn, {
           />
         );
       }
+      case 'objectives': {
+        if (row.detail === ' First Blood') {
```
howardchung: I suggest matching on a key value as display strings may change

src/utility/index.jsx
```diff
@@ -592,3 +592,47 @@ export const wilsonScore = (up, down) => {
...
+    npc_dota_goodguys_tower3_bot: `${strings.general_radiant} ${strings.bot_tower} ${strings.tier3}`,
+    npc_dota_goodguys_tower4: `${strings.general_radiant} ${strings.heading_tower} ${strings.tier4}`,
+    npc_dota_goodguys_melee_rax_top: `${strings.barracks_value_1024} ${strings.heading_barracks}`,
+    npc_dota_goodguys_melee_rax_mid: `${strings.barracks_value_256} ${strings.heading_barracks}`,
+    npc_dota_goodguys_melee_rax_bot: `${strings.barracks_value_64} ${strings.heading_barracks}`,
+    npc_dota_goodguys_range_rax_top: `${strings.barracks_value_2048} ${strings.heading_barracks}`,
+    npc_dota_goodguys_range_rax_mid: `${strings.barracks_value_512} ${strings.heading_barracks}`,
+    npc_dota_goodguys_range_rax_bot: `${strings.barracks_value_128} ${strings.heading_barracks}`,
+  },
+  dire: {
```
howardchung: since the only thing different about the dire strings is that the first part is different, you can probably cut the code in half by making this a function that takes a param for radiant/dire

src/components/Match/MatchLog.jsx
```diff
@@ -118,6 +119,69 @@ const logColumns = [heroTdColumn, {
...
+              />
+              <p>{strings.th_firstblood_claimed}</p>
+            </span>
+          );
+        }
+        if (tb.radiant[row.key]) {
+          return (
+            <span>
+              <img
+                src="/assets/images/dota2/dotastructuresmallradiant.png"
```
howardchung: The structure icon looks kind of strange--maybe just use the IconRadiant and IconDire that we already have?

src/components/Match/MatchLog.jsx
Argn0: Unfortunately I can't find any key value I could use.

src/utility/index.jsx
Argn0: There is no radiant/dire value I could pass to a function. I use the translateBuildings object itself to figure out if it's radiant or dire.

I thought about it for 20 minutes but I can't figure out how to shorten the code :(

src/components/Match/MatchLog.jsx
howardchung: If you look at the raw API data there is a type field:
https://api.opendota.com/api/matches/3584342151

Maybe it's not being passed down to this component it should be possible to do so

src/lang/en-US.json
```diff
@@ -68,6 +68,8 @@
   "barracks_value_1024": "Radiant Top Melee",
   "barracks_value_2048": "Radiant Top Ranged",
 
+ "building_ancient": "Ancient",
```
howardchung: can you double-check to see if there's a string for this already?

src/utility/index.jsx
howardchung: Don't you have row.isRadiant on the row?

src/lang/en-US.json
Argn0: there is a  string for the rank "Ancient" but not the building. They are translated differently for example in German.

howardchung:
@blukai any comments?

@Argn0 the longer term goal is to make this log resemble the updated chat log more closely.  See #1140 

Is that something you wanted to do in this PR, or should we merge this and then do more changes later?

Argn0:
I will try to make further changes.

Argn0:
Done.

src/components/Match/MatchLog.jsx
```diff
@@ -4,16 +4,53 @@ import strings from 'lang';
...
+  );
+};
+
+const radiantConditions = (field, row) => {
+  if (row.alt_key === 'CHAT_MESSAGE_COURIER_LOST') {
+    return row.team !== 2;
+  }
+  return field === true || (row.unit && row.unit.indexOf('goodguys') !== -1) || row.team === 2;
+};
+const direConditions = (field, row) => {
```
howardchung: isRadiant/isDire might be a better name

src/utility/index.jsx
```diff
@@ -592,3 +592,30 @@ export const wilsonScore = (up, down) => {
...
+    tower2_top: ` ${strings.top_tower} ${strings.tier2}`,
+    tower3_top: ` ${strings.top_tower} ${strings.tier3}`,
+    tower1_mid: ` ${strings.mid_tower} ${strings.tier1}`,
+    tower2_mid: ` ${strings.mid_tower} ${strings.tier2}`,
+    tower3_mid: ` ${strings.mid_tower} ${strings.tier3}`,
+    tower1_bot: ` ${strings.bot_tower} ${strings.tier1}`,
+    tower2_bot: ` ${strings.bot_tower} ${strings.tier2}`,
+    tower3_bot: ` ${strings.bot_tower} ${strings.tier3}`,
+    tower4: ` ${strings.heading_tower} ${strings.tier4}`,
+    melee_rax_top: `${' Top'} ${strings.building_melee_rax}`,
```
howardchung: for consistency maybe put the space in the template string like you did above

PR #1344, opened 2017-11-29
Enabled service worker in public folder

superoo7:
I tried to make the enabler serviceworker js code in external file, it simulate the previous error 
` "unexpected token <" error`, so I tried to make a `script` tag in `index.html`, service worker is working again

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,11 +1,11 @@
-// Service
-
+/* eslint-disable */
+// Service Worker
 self.addEventListener('install', function(event) {
-    console.log("Install Service Worker", event);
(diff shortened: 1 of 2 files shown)


superoo7:
There is an issue for build `   3:1   error    Unexpected use of 'self'  no-restricted-globals` in `sw.js`, need to disable checking on `sw.js` since service worker does not affect main js file

public/sw.js
```diff
@@ -0,0 +1,15 @@
+// Service
```
howardchung: you can disable lint with `/* eslint-disable */`

PR #1353, opened 2017-12-01
disable party indicators when sort is applied to table issue #1280

### LINKED ISSUE #1280: Player groups in match statistics
When sorting is applied, player group bindings acting strange:

Maybe need to disable sorting like on Dotabuff?

Argn0:
fixes #1280 

https://deploy-preview-1353--opendota-stage.netlify.com/matches/3582800334

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -132,6 +132,10 @@ class Table extends React.Component {
     if (paginated) {
       data = data.slice(currentPage * pageLength, (currentPage + 1) * pageLength);
     }
+    data.forEach((row) => {
+      const r = row;
+      r.sortApplied = !!sortField && (sortField !== 'player_slot' || sortState !== 'asc');
+    });
     return (
       <StyledBody>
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.jsx
```diff
@@ -132,6 +132,10 @@ class Table extends React.Component {
     if (paginated) {
       data = data.slice(currentPage * pageLength, (currentPage + 1) * pageLength);
     }
+    data.forEach((row) => {
+      const r = row;
+      r.sortApplied = !!sortField && (sortField !== 'player_slot' || sortState !== 'asc');
```
howardchung: why the double negate?

Argn0: I use it instead of a type cast to boolean

howardchung: I prefer the type cast, it's more readable

Argn0: No idea what I was trying to do there, but it works just fine like this now. :)

Issue #1352, opened 2017-12-01
remove avg mmr, player mmr from match pages


howardchung:
These values will slowly get more and more inaccurate

PR #1354, opened 2017-12-01
Remove Avg. MMR and players' MMR from match page #1352

warchildmd:
#1352 Removed the average match MMR and individual players' MMR from match page.

Removing the MMR leaves the subtitle empty, or ugly (when there are the `randomed` or `predicted` flags), so I replaced it with the competitive rank name. But as the rank is not returned in the /matches API for now, it will show as Unknown.

### DIFF
--- a/src/components/Visualizations/Table/HeroImage.jsx
+++ b/src/components/Visualizations/Table/HeroImage.jsx
@@ -402,4 +402,27 @@ Mmr.propTypes = {
   number: PropTypes.number,
 };
 
+export const CompetitiveRank = ({ rankTier }) => (
+  <span>
+    <section
+      data-hint="Competitive rank"
+      data-hint-position="bottom"
+    >
+      <SocialPerson />
+    </section>
+    {rankTier && rankTier === parseInt(rankTier, 10) ?
+      <span>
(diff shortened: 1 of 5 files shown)


src/components/Visualizations/Table/HeroImage.jsx
```diff
@@ -402,4 +402,27 @@ Mmr.propTypes = {
...
+  <span>
+    <section
+      data-hint="Competitive rank"
+      data-hint-position="bottom"
+    >
+      <SocialPerson />
+    </section>
+    {rankTier && rankTier === parseInt(rankTier, 10) ?
+      <span>
+        {strings[`rank_tier_${parseInt(rankTier / 10, 10)}`]} { (parseInt(rankTier, 10) > 9) && <span>[{parseInt(rankTier % 10, 10)}]</span> }
```
howardchung: We may want to move this logic to a utility function (we also use it for rank distribution, should use the shared function there)

src/components/Visualizations/Table/HeroImage.jsx
```diff
@@ -402,4 +402,27 @@ Mmr.propTypes = {
   number: PropTypes.number,
 };
 
+export const CompetitiveRank = ({ rankTier }) => (
+  <span>
+    <section
+      data-hint="Competitive rank"
```
howardchung: remove this string or move it to lang file?  Not sure what data-hint is used for

warchildmd:
I'm a React newbie, please take a look again

PR #1356, opened 2017-12-02
removed all jist.tv related stuff

Argn0:


howardchung:
@albertcui to validate that we can do this now

albertcui:
Yep we are good to remove. Thanks!

PR #1359, opened 2017-12-03
add ranked medals to player pages

### LINKED ISSUE #1351: add ranked medals to player pages
medal images added in #1337 

Data is present in the players API call, just need to render the correct image

GravityMsc:
fixes #1351 

Emmmm, I add ranked medals in front of the playerName.
Could you give me some suggestions about its size or position?
---
add a tooltip to the medal and add some Chinese transitions(sorry, I forgot to use https://translate.opendota.com/)

Q: How to solve 76 rank tier? It will cause a display problem.


howardchung:
Size looks good.  Could you add a tooltip to the medal?

GravityMsc:
@howardchung   OK, but I have found that the rank_tier, such as Sumail. It is 76 now. how to solve this problem?
And I 'll add a tooltip, like 
image
is it OK?

howardchung:
I think 76 should be treated as 75 for display purposes (convert all values larger than 5 to display 5)

GravityMsc:
@howardchung  I will do that immediately, and change the function of utility

howardchung:
we might want to keep the text as 6, just use 5 for the stars since we don't have a 6 star image.

GravityMsc:
@howardchung I‘ve finished.

PR #1365, opened 2017-12-04
moved player slot indicator from left to right

Argn0:
I think it will look better like this with the new party indicators.


howardchung:
Do you still want to combine this with #1366?

Argn0:
Yes please.  I think it looks much better on the right.

PR #1369, opened 2017-12-04
party indicators are bugged in current implemantation

Argn0:
when the array with party IDs looks like this for example it does not work properly. 

https://stage.opendota.com/matches/3575778268/
[0, 1, 2, **3, 3**, 4, 5, 6, 7, 8]  (party ID array)

player 0, 1 and 2 get each treated like an individual party. then the two players with ID 3. 
players  5,6,7,8,9   get treated like a **single** party.

another example:
https://stage.opendota.com/matches/3595762027
party_id array [0, 0, 1, 1, 2, 3, 4, 5, 5, 5]

should be:
party_0: player 0 & 1
party_1: player 2 & 3
party_2: player 7 & 8 & 9

more bugged examples: 
https://stage.opendota.com/matches/3593083952/

also team matches with arrays like  this [0,0,0,0,0,0,0,0,0,0]:
https://stage.opendota.com/matches/3597234172

I tried to patch this issue without changing the original implementation too much.

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -63,7 +63,6 @@ function findParties(partyArray) {
 
 const partyStyles = (row, match) => {
   if (match.players && match.players.map(player => player.party_id).reduce(sum) > 0 && row.party_size > 1) {
-    console.log(match.players.map(player => player.party_id))
     const parties = findParties(match.players.map(player => player.party_id));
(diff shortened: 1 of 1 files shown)


sepehr500:
I made a PR that fixes these bugs and enhances the styling. Also take a look at my implementation of the grouping. I tried to keep it as functional as I could. #1370

howardchung:
will take #1370

PR #1370, opened 2017-12-05
Fixed bugs and enhanced stying

sepehr500:
I fixed the grouping algo to work with solo/party mix matches and enhanced the styling.

image

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -9,7 +9,7 @@ import abilityKeys from 'dotaconstants/build/ability_keys.json';
 import buffs from 'dotaconstants/build/permanent_buffs.json';
 import util from 'util';
 import strings from 'lang';
-import { formatSeconds, abbreviateNumber, transformations, percentile, sum, subTextStyle, getHeroesById, rankTierToString } from 'utility';
+import { formatSeconds, abbreviateNumber, transformations, percentile, sum, subTextStyle, getHeroesById, rankTierToString, groupBy } from 'utility';
 import { TableHeroImage, inflictorWithValue } from 'components/Visualizations';
 import ReactTooltip from 'react-tooltip';
 import { RadioButton } from 'material-ui/RadioButton';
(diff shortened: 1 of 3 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -50,8 +50,14 @@ export const heroTdColumn = {
 };
 
 const partyStyles = (row, match) => {
-  if (match.players && match.players.filter(x => x.party_size > 1).length) {
-    return <div className={`group${row.party_id}`} />;
+  if (match.players && row.party_size > 1) {
+    const partyIndex = lodash.compose(
```
howardchung: avoid using lodash functions if possible to reduce bundle size.

howardchung:
I would roll back the last commit to avoid adding bundle size

src/components/Match/matchColumns.jsx
sepehr500: It was already being imported in the file. I can try importing the functions individually.

howardchung: I think because of tree shaking the bundle size is smaller if the functions aren't used.  I'd like to move the entire codebase away from lodash eventually, but haven't had the time to do so.

sepehr500:
@howardchung  I updated the PR to import the functions individually. Now bundle size addition should be negligible.

sepehr500:
@howardchung  I reverted the code.

src/components/Match/matchColumns.jsx
```diff
@@ -51,7 +51,26 @@ export const heroTdColumn = {
 
 const partyStyles = (row, match) => {
   if (match.players && match.players.filter(x => x.party_size > 1).length) {
-    return <div className={`group${row.party_id}`} />;
+    const groupedPlayers = lodash.compose(
```
howardchung: can we remove these lodash references?

sepehr500:
Would you prefer I Implement groupBy by hand? Or is there another lib you had in mind?

src/components/Match/matchColumns.jsx
sepehr500: Would you prefer I Implement groupBy by hand? Or is there another lib you had in mind?

howardchung: Yes, please, if possible.  You could just copy the lodash implementation (if it doesn't have any dependencies)

sepehr500: Okay. I will get to it tomorrow. Just out of curiosity, is there no plan to pull in any utility library? Things might get out of hand if people are copy-pasting lodash functions all the time.

howardchung: Generally we've found that with the latest ES6 features most things can be implemented nicely with native functions.  Any utility functions we need can be added to utility.js for reusability.

Argn0:
I like how it looks now. But sometimes there is a tiny gap between indicators. Is there a way to fix it?

sepehr500:
@howardchung I removed lodash from the function and fixed gap between indicators.

Argn0:
@sepehr500  Do you think he should move the player slot color indicators from left to right so they overlap less with the party indicators?

howardchung:
Thanks!  LGTM

Issue #1347, opened 2017-11-30
Stat request for the 'Vision' page.


j08:
On the detailed Vision page for a match, is it possible to add in an Average Lifespan of Observers and Sentries somewhere? You pretty much have all that data on the page already, but to see it in a column or box somewhere would make it faster and easier to digest.

I mainly play support and knowing if I was being de-warded often would help me.

A second feature for that page that would be nice if possible is in the Ward Log section to make the ward in the Type column clickable. When it is clicked it shows what it's position was on the map above. The reason this would help is if a ward dies quickly I could see where it was placed and why it was de-warded.

Support life is hard enough without having to spend gold I don't have on these things carries keep calling for :) Any help to keep them alive for longer helps (the wards that is, there is no hope for many of my carries).

howardchung:
1) Averages are certainly doable, just need to compute and render them into a box/table in client side.

2) @dartharaknus might know more about mapping the ward log table to the wards shown on the map.  I think it should be possible?

howardchung:
The second half of this feature request is implemented in #1373

j08:
Thank you. It's not quite what I meant, but the extra detail is very welcome!

PR #1373, opened 2017-12-06
 Show ward placements on hover in vision log

Argn0:
requested  in #1347 

https://deploy-preview-1373--opendota-stage.netlify.com/matches/3601207912/vision


src/lang/en-US.json
```diff
@@ -1175,6 +1175,8 @@
   
   "show_consumables_items": "Show consumables",
   "activated": "Activated",
-  "rune": "Rune"
+  "rune": "Rune",
+  "placement": "Placement on map"
```
howardchung: I would just call this "Placement"

PR #1374, opened 2017-12-06
updated ranked icons

moonrobin:
Also updated the wiki page for Valve's first party Resource Viewer tool.


howardchung:
Thanks!  Do we know what the criteria is for displaying the 7a, 7b, 7c icons?

@GravityMsc is planning to work on updating the code to display icons.

GravityMsc:
`api/players/xxxxxx` shows 3 kinds of rank number: competitive_rank, solo_competitive_rank, leaderboard_rank. Maybe I gues I only need leaderboard_rank, is it right?

And... I'm going to know the 7a icon. In my opinion, 7b means Top 100, 7c means Top 10.

howardchung:
yeah competitive/solo_competitive are party/solo MMR that aren't in the client anymore.  We still return the last known value.

GravityMsc:
OK, I still need an example to explain the 7a icon.
7c: No.10 -- 1
7b: NO.100 -- 11
7a: ? maybe its rank tier is 76 and have the leaderboard_rank number?

And Chinese rank tier transitions will be fixed in the future.

howardchung:
Yeah, I guess 76 outside the top 100 makes sense for 7a.

PR #1376, opened 2017-12-06
Show average lifespan of wards in vision tab

### LINKED ISSUE #1347: Stat request for the 'Vision' page.

Argn0:
fixes #1347

https://deploy-preview-1376--opendota-stage.netlify.com/matches/3597640305/vision


src/components/Match/matchColumns.jsx
```diff
@@ -1029,6 +1029,53 @@ export const teamfightColumns = [
...
+  displayName: (
+    <div style={{ display: 'inline-flex', verticalAlign: 'middle' }}>
+      <img height="15" src={`${process.env.REACT_APP_API_HOST}/apps/dota2/images/items/ward_observer_lg.png`} alt="" />
+      &nbsp;{strings.th_duration_shorthand}
+    </div>
+  ),
+  tooltip: strings.tooltip_duration_observer,
+  sortFn: true,
+  displayFn: row => formatSeconds(computeAverage(row, 'obs')) || '-',
+  relativeBars: true,
```
howardchung: Looks like the bars aren't working

Argn0: Ops I totally forgot about the bars. I'll try to do something about them.

howardchung: I think they use the displayFn value so if it's a string it behaves strangely.  We handled this for time dead though so there's a workaround

PR #1377, opened 2017-12-07
Hide stray hero icons on Buildings Map in Overview tab

Argn0:


howardchung:
isn't this the map that we want the icons on?

Argn0:
it's this one in Overview 
overviewmap

howardchung:
Yes, isn't that the map we want to show the icons on?  I don't think we actually show them anywhere else.

Argn0:
That's weird...What's the purpose of those icons? What do those positions represent?

Argn0:
Must be laning. I misunderstood the icons. I thought that map is only supposed to show the destroyed buildings .

howardchung:
Maybe we should retitle the map

PR #1380, opened 2017-12-08
Fix party indicators

Argn0:
they're still bugged in some cases

https://stage.opendota.com/matches/3600816414 ( those are 10 solo players, not 2 parties)

https://stage.opendota.com/matches/3597234172 (both teams have same color)

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -54,9 +54,8 @@ const partyStyles = (row, match) => {
     if (row.isRadiant) {
       return <div className={`group${0}`} />;
     }
-    else {
-      return <div className={`group${3}`} />;
-    }
+
+    return <div className={`group${3}`} />;
   }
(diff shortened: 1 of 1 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -50,7 +50,13 @@ export const heroTdColumn = {
 };
 
 const partyStyles = (row, match) => {
-  if (row.party_size === 1) {
+  if (row.lobby_type === 1) { // team matches
```
howardchung: This excludes ranked matches, you should check all players to see if there are any parties.

Argn0: do you mean ranked matches where both sides are a party of 5 players each?

howardchung: lobby_type 1 just means a non-ranked match, it doesn't say anything about whether it's a party match or not.

PR #1384, opened 2017-12-11
add 'abbr_number' and fix the position of the rank tier tooltip

### LINKED ISSUE #1383: replace "No. " string
https://github.com/odota/web/blob/2d252ebaaa9c54f1a6813f9bbc38bbf872c172c9/src/components/Player/Header/PlayerHeader.jsx#L166

@GravityMsc could you replace "No." with a `#` symbol or move it to the string file?

GravityMsc:
fix #1383

### DIFF
--- a/src/lang/en-US.json
+++ b/src/lang/en-US.json
@@ -10,6 +10,7 @@
   "abbr_not_available": "N/A",
   "abbr_pick": "P",
   "abbr_win": "W",
+  "abbr_number": "#",
 
   "analysis_eff": "Lane efficiency",
   "analysis_farm_drought": "Lowest GPM in 5 minute interval",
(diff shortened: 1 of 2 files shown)


src/lang/en-US.json
```diff
@@ -10,6 +10,7 @@
   "abbr_not_available": "N/A",
   "abbr_pick": "P",
   "abbr_win": "W",
+  "abbr_number": "#",
```
howardchung: if you're using a `#` symbol i don't think you need to use the string file since it doesn't need to be translated.

GravityMsc: I think this is necessary. I want to use '排名' in Chinese file.

GravityMsc: and to be honest, I don't know what is the proper word of the U.S.

howardchung: `No.` is ok for English.  If you use `#` in the string file the translators probably won't know what to do with it.

GravityMsc: I used 'No.' as a temporary word at the beginning, lol. I misunderstood it.

PR #1388, opened 2017-12-12
Breakdown of ability targets on hover in casts tab

Argn0:
fixes https://github.com/odota/web/issues/1172

core change here https://github.com/odota/core/pull/1577


Argn0:
https://deploy-preview-1388--opendota-stage.netlify.com/matches/3616524005/casts

howardchung:
I would suggest creating a new table, with each column being an opposing hero, and the number of times an ability was used in the cell.

That way you can keep the existing tooltip (I think this is needed because it's hard to know immediately what an ability is just from image) and I think it makes it easier to read.

src/lang/en-US.json
```diff
@@ -728,6 +728,7 @@
   "th_gold": "G",
   "th_xp": "XP",
   "th_abilities": "Abilities",
+  "th_target_abilities": "Target Abilities",
```
howardchung: Ability Targets sounds a little better?

src/components/Match/matchColumns.jsx
```diff
@@ -937,6 +937,27 @@ export const inflictorsColumns = [
...
 
+const sumValues = f => Object.values(f).reduce((a, b) => a + b);
+
+const targetTooltip = (t) => {
+  const targets = [];
+  Object.keys(t).forEach((target) => {
+    const heroicon = heroes[getHeroesById()[target].id] && process.env.REACT_APP_API_HOST + heroes[getHeroesById()[target].id].icon;
+    const j = (
+      <div style={{ display: 'inline-block', padding: '5px', paddingBottom: '20px' }}>
+        <span style={{ fontSize: '100%' }}>{`${t[target]}x`}</span>
```
howardchung: maybe overlay this onto the icon like how we do inflictorwthvalue?

src/lang/en-US.json
```diff
@@ -937,6 +938,7 @@
   "tooltip_multikill": "Longest multi-kill",
   "tooltip_killstreak": "Longest killstreak",
   "tooltip_casts": "Number of times this ability/item was cast",
+  "tooltip_target_abilities": "Breakdown of how many times this hero used his target abilities on each enemy/ally",
```
howardchung: Suggested: "How many times each hero was targeted by this hero's abilities"

Argn0:
pixel work :-)

deploy-preview-1388--opendota-stage.netlify.com/matches/3623193404/casts

PR #1393, opened 2017-12-14
Cap time to game duration when computing graphs and farm drought 

Argn0:
fixes https://github.com/odota/web/issues/1391

https://www.opendota.com/matches/3612852168/graphs

https://deploy-preview-1393--opendota-stage.netlify.com/matches/3612852168/graphs


howardchung:
Thanks, although we should probably fix this at the parsing level too.

PR #1395, opened 2017-12-14
Add toggle to exclude Turbo games from averages and make numbers clickable 

Argn0:
fixes https://github.com/odota/web/issues/1371

https://deploy-preview-1395--opendota-stage.netlify.com/players/226089075

There are people who almost exclusively play Turbo games. So I guess completely hiding the stats for them is not a good idea.


howardchung:
I'd actually prefer to exclude turbo without a switcher to keep the UI clean, but if you feel strongly that it should be configurable I'm ok with this

Argn0:
I'll test how it looks with a small checkbox

Argn0:
There are a lot of Turbo only players and the symbol doesn't show on profiles that have no Turbo games recently.

PR #1406, opened 2017-12-18
[WIP] Allow multiple selects/wheres on explorer UI

Argn0:
fixes https://github.com/odota/web/issues/984
fixes https://github.com/odota/web/issues/1644

projection of multiple columns (e.g. select kills, deaths, assists)
multiple where queries with OR (e.g. select * from two leagues)


howardchung:
Will comment with issues as I find them:

Error when multiple values are selected for group by

src/components/Explorer/queryTemplate.js
```diff
@@ -1,3 +1,25 @@
...
+  if (value instanceof Array) {
+    value.forEach((s, index) => {
+      r += str0 + s + str1;
+
+      if (index !== value.length - 1) {
+        r += ' OR ';
+      }
+    });
+  } else {
+    return `${str0 + value + str1})`;
```
howardchung: if you are just concatenating here you can just do `${str0}${value}${str1}`

src/components/Explorer/queryTemplate.js
```diff
@@ -1,3 +1,25 @@
+function templ(strings, value) {
+  if (value.length === 0) {
+    return '';
+  }
+  let r = 'AND (';
+  const str0 = strings[0];
+  const str1 = strings[1];
+
+  if (value instanceof Array) {
+    value.forEach((s, index) => {
```
howardchung: I suggesting building the interior using `.map` and then saving it to a variable.

At the end you can do `AND(${var})`

src/components/Explorer/ExplorerFormField.jsx
```diff
@@ -77,33 +77,19 @@ class ExplorerFormField extends React.Component {
...
-        />
+        {
+          <FormField
+            name={builderField}
+            label={label}
+            dataSource={dataSource}
+            formSelectionState={builder}
+            filter={AutoComplete.caseInsensitiveFilter}
+            strict
+            limit={5}
```
howardchung: may want to show more than 5. . . maybe 10?

src/components/Explorer/queryTemplate.js
```diff
@@ -1,3 +1,25 @@
...
+  if (!value.length) {
+    return '';
+  }
+  const str0 = strings[0];
+  const str1 = strings[1];
+  if (value instanceof Array) {
+    r = o.map((s, index) => {
+      const p = `${str0}${s}${str1}`;
+      if (index !== o.length - 1) {
+        return `${p} OR `;
```
howardchung: you can probably push p into an array and then join them with ' OR '

src/components/Explorer/queryTemplate.js
```diff
@@ -32,6 +54,13 @@ const queryTemplate = (props) => {
   // team
   // organization
   let query;
+  let grouparray = [];
+  if (!(group instanceof Array)) {
```
howardchung: Array.isArray might be better?

src/components/Explorer/queryTemplate.js
```diff
@@ -32,6 +54,13 @@ const queryTemplate = (props) => {
...
+    grouparray.push(group);
+  } else {
+    grouparray = group;
+  }
+  const groupexists = grouparray !== null && grouparray.length > 0 && grouparray[0] !== null;
```
howardchung: use camelCase (groupExists), same for groupArray

src/components/Explorer/queryTemplate.js
```diff
@@ -71,75 +100,90 @@ ${isTi7Team ? 'AND teams.team_id IN (5, 15, 39, 46, 2163, 350190, 1375614, 18383
...
-${select && select.distinct && !group ? `DISTINCT ON (${select.value})` : ''}
-${(group) ?
-    [`${group.groupKeySelect || groupVal} ${group.alias || ''}`,
-      (select && select.countValue) || '',
-      (select && select.avgPerMatch) ? `sum(${selectVal})::numeric/count(distinct matches.match_id) avg` : `${select && select.avg ? select.avg : `avg(${selectVal})`} avg`,
+    ${select ? select.map(x => (x.distinct && grouparray.length < 1 ? `DISTINCT ON (${x.value})` : '')).join('') : ''}
+    ${(groupexists) ?
+    grouparray.map(x =>
+      [`${x.groupKeySelect || groupVal[x.key]} ${x.alias || ''},`].filter(Boolean).join(',\n')).join('') : ''}
+${(groupexists) ? // eslint-disable-line no-nested-ternary
```
howardchung: I recommend not disabling no-nested-ternary, move the code out into another function if needed.

src/components/Explorer/queryTemplate.js
```diff
@@ -71,75 +100,90 @@ ${isTi7Team ? 'AND teams.team_id IN (5, 15, 39, 46, 2163, 350190, 1375614, 18383
...
+${organization ? templ`team_match.team_id = ${organization} AND (player_matches.player_slot < 128) = team_match.radiant` : ''}
+${laneRole ? templ`player_matches.lane_role = ${laneRole}` : ''}
+${region ? templ`matches.cluster IN (${region})` : ''}
+${minDate ? templ`matches.start_time >= extract(epoch from timestamp '${new Date(minDate).toISOString()}')` : ''}
+${maxDate ? templ`matches.start_time <= extract(epoch from timestamp '${new Date(maxDate).toISOString()}')` : ''}
+${tier ? templ`leagues.tier = '${tier}'` : ''}
 ${isTi7Team ? 'AND teams.team_id IN (5, 15, 39, 46, 2163, 350190, 1375614, 1838315, 1883502, 2108395, 2512249, 2581813, 2586976, 2640025, 2672298, 1333179, 3331948, 1846548)' : ''}
-${group ? `GROUP BY ${groupVal}` : ''}
-${group ? `HAVING count(distinct matches.match_id) >= ${having ? having.value : '1'}` : ''}
+${groupexists ? 'GROUP BY' : ''}${(groupexists && grouparray.map(x => ` ${groupVal[x.key]}`)) || ''}
```
howardchung: do you need to separate group by fields with commas?

src/components/Explorer/queryTemplate.js
```diff
@@ -71,75 +100,90 @@ ${isTi7Team ? 'AND teams.team_id IN (5, 15, 39, 46, 2163, 350190, 1375614, 18383
...
+${groupexists ? `HAVING count(distinct matches.match_id) >= ${having ? having.value : '1'}` : ''}
 ORDER BY ${
-  [`${group ? 'avg' : (select && select.value) || 'matches.match_id'} ${(order && order.value) || (select && select.order) || 'DESC'}`,
-    group ? 'count DESC' : '',
+  [`${groupexists ? typeof select === 'string' ? `"AVG ${select.text}"` : `"AVG ${select[0].text}"` : // eslint-disable-line no-nested-ternary
```
howardchung: suggest finding another way to write these without nested ternaries

src/components/Explorer/queryTemplate.js
```diff
@@ -71,75 +100,90 @@ ${isTi7Team ? 'AND teams.team_id IN (5, 15, 39, 46, 2163, 350190, 1375614, 18383
...
+    select ? select.map(x => `${x.value} DESC`).join(',')
+      : 'matches.match_id'} ${order ? order.map(x => x.value).join('') // eslint-disable-line no-nested-ternary
+    : select ? '' : 'DESC'}`,
+  groupexists ? 'count DESC' : '',
   ].filter(Boolean).join(',')} NULLS LAST
-LIMIT ${limit ? limit.value : 200}`;
+LIMIT ${limit ? limit.map(x => x.value).join('') : 200}`;
```
howardchung: limit cannot have multiple values, we should restrict this input

Argn0:
I fixed everything you mentioned in your last review, but I'm not sure what you meant by "do you need to separate group by fields with commas?"

Also I don't know why the deploy fails with lint errors in files I've never touched.

howardchung:
* The chips are displaying the internal string name, e.g. `ability_uses` rather than the display name "Ability Uses"
* Behavior has changed with the selects that do a join and group the results automatically (e.g. Ability Uses).  Possibly we should rethink how those selects work, it's currently kind of unintuitive that some select fields apply a grouping automatically.
* Selecting multiple fields that do a json_each join causes an exception:
https://deploy-preview-1406--opendota-stage.netlify.com/explorer?select=damage_instances&select=ability_used&minDate=2017-12-01T08%3A00%3A00.000Z&maxDate=&
(This is related to the above issue, possibly rearranging how these fields are selected will fix the issue)

howardchung:
Additional side thought:  It may be worth writing a test suite for the explorer that automatically tries out combinations of queries and checks the results with a tool such as:
https://github.com/forward/sql-parser

That might not capture all the nuances of Postgres syntax, but it might be a good start and help prevent regressions.

howardchung:
I don't think it ever makes sense to do a multi-json_each query.  Sure, you can do it but the resulting data is not that useful.  It also results in a lot of output rows which increases the chance the query will just time out.

Argn0:
How about we don't allow to combine multiple json_each fields and indicate it by marking those fields with a * and a hint somewhere that you cannot select multiple fields with a * symbol

howardchung:
I think we should just move those select fields somewhere else, either into the Group By options or something else altogether (maybe tabs on the explorer UI).

Argn0:
I moved all json_each fields to GROUP BY and set the chip limit to 1.

src/components/Explorer/ExplorerFormField.jsx
```diff
@@ -55,10 +56,22 @@ class ExplorerFormField extends React.Component {
...
+      fields, label, builderField, handleFieldUpdate, isDateField, builder, chipLimit,
     } = this.props;
     const dataSource = fields && fields[builderField];
     const fieldWidth = 280;
+    const translatedBuilder = {};
+    if (builder) {
+      Object.keys(builder).forEach((field) => {
+        if (Array.isArray(builder[field])) {
+          translatedBuilder[field] = builder[field].map((x) => {
+            let name = explorerFields()[field].find(f => f.key === x);
```
howardchung: hitting an exception when using the meta tab (it shares this component)

src/components/Explorer/queryTemplate.js
```diff
@@ -1,7 +1,31 @@
...
+    o.forEach((s) => {
+      r.push(`${str0}${s}${str1}`);
+    });
+    return `AND (${r.join(' OR ')})`;
+  }
+  return `AND ${str0}${value.value}${str1}`;
+}
+
+function validate(p) {
+  if (p !== null && p !== undefined && p.length > 0 && p[0] !== null && p[0] !== undefined) {
```
howardchung: shorthand: just return your if condition

src/components/Explorer/queryTemplate.js
```diff
@@ -71,71 +130,84 @@ ${isTi7Team ? 'AND teams.team_id IN (5, 15, 39, 46, 2163, 350190, 1375614, 18383
...
+${validate(playerPurchased) ? templ`(player_matches.purchase->>'${playerPurchased}')::int > 0` : ''}
+${validate(minDuration) ? templ`matches.duration >= ${minDuration}` : ''}
+${validate(maxDuration) ? templ`matches.duration <= ${maxDuration}` : ''}
+${validate(side) ? templ`(player_matches.player_slot < 128) = ${side}` : ''}
+${validate(result) ? templ`((player_matches.player_slot < 128) = matches.radiant_win) = ${result}` : ''}
+${validate(team) ? templ`notable_players.team_id = ${team}` : ''}
+${validate(organization) ? templ`team_match.team_id = ${organization} AND (player_matches.player_slot < 128) = team_match.radiant` : ''}
+${validate(laneRole) ? templ`player_matches.lane_role = ${laneRole}` : ''}
+${validate(region) ? templ`matches.cluster IN (${region})` : ''}
+${validate(minDate) ? templ`matches.start_time >= extract(epoch from timestamp '${new Date(minDate).toISOString()}')` : ''}
```
howardchung: date filters don't seem to be getting applied to the query, e.g.:
https://deploy-preview-1406--opendota-stage.netlify.com/explorer?select=assists&&minDate=2018-01-24T08%3A00%3A00.000Z

src/components/Match/PlayerThumb/index.jsx
```diff
@@ -19,7 +19,7 @@ const StyledImg = styled.img`
 `;
 
 const PlayerThumb = ({
-  player_slot, hero_id, name, personaname, hideText,
+  player_slot, hero_id, name, personaname, hideText, // eslint-disable-line camelcase
```
howardchung: you can remove these eslint disables, they have been fixed in master (and this is causing a merge conflict)

howardchung:
Also suggest setting min date to 30 days ago by default to improve query performance

Argn0:
-restored meta tab
-date filters are now getting applied to the query
-default min date is now 30 days ago

howardchung:
Default table doesn't seem to be rendering:
https://deploy-preview-1406--opendota-stage.netlify.com/explorer?

src/components/Explorer/ExplorerFormField.jsx
```diff
@@ -67,8 +80,8 @@ class ExplorerFormField extends React.Component {
             floatingLabelText={label}
             container="inline"
             autoOk
-            defaultDate={builder[builderField] ? new Date(builder[builderField]) : undefined}
-            onShow={this.resetField}
+            defaultDate={minDate ? new Date(new Date().setDate(new Date().getDate() - 30)) : undefined}
```
howardchung: There doesn't seem to be any way to remove the default date, perhaps we should look at setting this value a different way?

howardchung: e.g. if the user doesn't do anything, the value should be set to 30 days ago.

However, if they click on the field and don't select anything, it should clear the value.

Argn0:
-fixed the "At least this many matches" field
-don't display chips for fields where you can only make a single selection

howardchung:
Do we want to allow this some more bake/test time before we deploy it?

Argn0:
I'd say we wait until next weekend. I will thoroughly test it until then.

Argn0:
I couldn't find more issues. I think it can be deployed now.

howardchung:
certain combinations of select queries fail:
e.g. https://deploy-preview-1406--opendota-stage.netlify.com/explorer?minDate=&minPatch=&maxPatch=&=&group=&select=buybacks&select=distinct_heroes

I think we need to organize the select options a little better now that they can be selected in compound.

Argn0:
organized select fields into bundles

Argn0:
I made it impossible to select multiple "Timings" or "Kill Times"

howardchung:
looks like a few conflicts to resolve

Argn0:
Resolved merge conflicts.

src/components/Explorer/queryTemplate.js
```diff
@@ -1,7 +1,28 @@
...
+    }
+    o.forEach((s) => {
+      r.push(`${str0}${s}${str1}`);
+    });
+    return `AND (${r.join(' OR ')})`;
+  }
+  return `AND ${str0}${value.value !== undefined ? value.value : value}${str1}`;
+}
+
+function validate(p) {
```
howardchung: might want to name this better, does it validate an array?  That it has at least one non-null element?

src/components/Explorer/queryTemplate.js
```diff
@@ -1,7 +1,28 @@
+function templ(strings, value) {
```
howardchung: i would give this a more descriptive name.  is it building a where clause?

src/components/Explorer/queryTemplate.js
```diff
@@ -54,86 +110,99 @@ JOIN leagues using(leagueid)
...
-${result ? `AND ((player_matches.player_slot < 128) = matches.radiant_win) = ${result.value}` : ''}
-${team ? `AND notable_players.team_id = ${team.value}` : ''}
-${organization ? `AND team_match.team_id = ${organization.value} AND (player_matches.player_slot < 128) = team_match.radiant` : ''}
-${laneRole ? `AND player_matches.lane_role = ${laneRole.value}` : ''}
-${region ? `AND matches.cluster IN (${region.value.join(',')})` : ''}
-${minDate ? `AND matches.start_time >= extract(epoch from timestamp '${new Date(minDate.value).toISOString()}')` : ''}
-${maxDate ? `AND matches.start_time <= extract(epoch from timestamp '${new Date(maxDate.value).toISOString()}')` : ''}
-${tier ? `AND leagues.tier = '${tier.value}'` : ''}
+${validate(selectArray) ? selectArray.map(x => `AND ${x.value} IS NOT NULL `).join('') : ''}
+${minPatch ? templ`match_patch.patch >= '${minPatch}'` : ''}
```
howardchung: perhaps you can allow the templating function to return an empty string if e.g. minPatch is falsy?  Then you could avoid this outer ternary.

howardchung:
The bundle approach is OK, but at some point I think we should do something similar to Scenarios and split the UI into a tab for each section/bundle so it's more discoverable.

Argn0:
-renamed  templating function to "conjoin" and it now returns an empty string if value is falsy
-changed and renamed validate function

howardchung:
Sorry, I totally forgot about this PR.  I don't think there any more outstanding issues, we should be able to release it next week probably?

Argn0:
I added these changes https://github.com/odota/web/pull/1406/commits/2190d3830ded9b2a4454e0855e55297eb0175757

howardchung:
Looks like a merge conflict.  planning to deploy this with next release (finally!)

Argn0:
resolved

PR #1410, opened 2017-12-24
Feature/1378 fixed column table

### LINKED ISSUE #1378: [PROPOSAL] split tables on mobile devices/fix header and first column
A proposal to split wide tables for mobile devices:

Currently we deal with wide tables by making them scrollable to the right, which works but is suboptimal because the leftmost column (like the hero column) gets hidden and it becomes difficult to tell which row the data goes with.

This makes large tables almost unusable on mobile devices:
https://www.opendota.com/heroes/public

We could extend our column definitions to include a field called `group` and pass an object to each `Table` component that maps each group name to a display name, e.g.

A sample columndef:
```
const columns = [{ field: 'kills', group: 'kda'}, { field: 'deaths', group: 'kda' }, { field: 'assists' , group: 'kda'}, { field: 'gold_per_min': group: 'farm' }]
```

Then we could have an object mapping each key name to a display name:
```
const groups: { kda: strings.kda, farm: strings.farm };
```

The table would detect when it is being rendered on a small screen and when it does, automatically create a dropdown menu (rendered above the table) that has as options all of the group names in the table.  Only one group would be selected at a time, and the table only renders the columns with a `group` name matching the currently selected value in the dropdown.

In the above example, the group names would be 'kda' and 'farm'.

JaydenIvanovic:
fixes #1378

### DIFF
--- a/src/components/Heroes/index.jsx
+++ b/src/components/Heroes/index.jsx
@@ -100,7 +100,13 @@ class RequestLayer extends React.Component {
             icon=""
             twoLine
           />
-          <Table data={data} columns={_columns} loading={loading} />
+          <div style={{position:"relative"}}>
+            <Table data={data} columns={_columns} loading={loading} />
+            <Table data={data} columns={[_columns[0]]} loading={loading} fixedStyle={{
+              position: "absolute",
+              top: "0px"
+            }}/>
+          </div>
         </div>),
       route: '/heroes/public',
     }];
--- a/src/components/Heroes/index.jsx
+++ b/src/components/Heroes/index.jsx
@@ -100,12 +100,17 @@ class RequestLayer extends React.Component {
             icon=""
             twoLine
           />
-          <div style={{position:"relative"}}>
+          <div style={{ position: 'relative' }}>
             <Table data={data} columns={_columns} loading={loading} />
-            <Table data={data} columns={[_columns[0]]} loading={loading} fixedStyle={{
-              position: "absolute",
-              top: "0px"
-            }}/>
+            <Table
+              data={data}
+              columns={[_columns[0]]}
+              loading={loading}
+              fixedStyle={{
+              position: 'absolute',
+              top: '0px',
+            }}
+            />
           </div>
         </div>),
       route: '/heroes/public',
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -134,7 +134,7 @@ class Table extends React.Component {
       data = data.slice(currentPage * pageLength, (currentPage + 1) * pageLength);
     }
     return (
-      <StyledBody>
+      <StyledBody style={this.props.fixedStyle}>
         {paginated && <Pagination
           numPages={Math.ceil(dataLength / pageLength)}
           currentPage={currentPage}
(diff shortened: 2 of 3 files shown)


src/components/Heroes/index.jsx
```diff
@@ -100,7 +100,18 @@ class RequestLayer extends React.Component {
             icon=""
             twoLine
           />
-          <Table data={data} columns={_columns} loading={loading} />
+          <div style={{ position: 'relative' }}>
+            <Table data={data} columns={_columns} loading={loading} />
```
JaydenIvanovic: Planning to extract these two tables into a single component to hide the implementation details of the fixed column behaviour

src/components/Table/Table.jsx
```diff
@@ -134,7 +134,7 @@ class Table extends React.Component {
       data = data.slice(currentPage * pageLength, (currentPage + 1) * pageLength);
     }
     return (
-      <StyledBody>
+      <StyledBody style={this.props.fixedStyle}>
```
JaydenIvanovic: Planning to make this prop a class name instead which can be generalised for other use cases

howardchung:
preview link:
https://deploy-preview-1410--opendota-stage.netlify.com/heroes/public

howardchung:
Would a similar approach work for freezing the header row?  That one may be trickier to implemnet.

src/components/Heroes/index.jsx
```diff
@@ -100,7 +100,13 @@ class RequestLayer extends React.Component {
             icon=""
             twoLine
           />
-          <Table data={data} columns={_columns} loading={loading} />
+          <div style={{position:"relative"}}>
+            <Table data={data} columns={_columns} loading={loading} />
+            <Table data={data} columns={[_columns[0]]} loading={loading} fixedStyle={{
```
GravityMsc: Well, I have a suggestion about `column`which is more  flexible. 
maybe you can add a prop  `fixed: right(or left)` in the column items, then you can fix some of columns right or left.

JaydenIvanovic: I like this suggestion. Would the idea be that if multiple items had a prop of "fixed:right" that they would stack and be fixed to the right?

JaydenIvanovic:
I believe a similar approach will work for the header row (seems to be implemented as such in the originally linked table). I will tackle that.

src/components/Heroes/index.jsx
GravityMsc: I think you could create a file of column tools, then you can filter some of columns use `Array.filter` or something else.
and if the filter column exists, such as 
```
const rightFixedColumns = columnTools.getRight(columns);
const leftFixedColumns = columnTools.getLeft(columns);
```
after that, use your code to create two fixed table above the original table. located right and left.

Sorry, I don't know whether you understand what I said. My English is not good.

JaydenIvanovic: No worries, that explanation is perfect. I'll tackle this as soon as I work through the kinks of the fixed header.

howardchung:
Seems to be a lint issue blocking the preview

howardchung:
There seem to be some rendering issues on mobile (testing with chrome emulator)

JaydenIvanovic:
I've fixed some of the mobile rendering issues.

There still seems to be one when scrolling vertically (especially when you hit the bottom of the "list") and another regarding the fixed hero column at a certain breakpoint (can see the hero column of the base table under it). 

Another known issue is that clicking column headers to sort the data won't sort the fixed hero's column (which I'm in the middle of fixing now). 

The header itself is now fixed correctly and aligned with the appropriate columns.

JaydenIvanovic:
Synced with master and finished off the table. I still need to do more testing to ensure that I haven't missed anything but I believe it's functioning as expected.

PR #1416, opened 2018-01-02
team pages

### LINKED ISSUE #1142: team pages
Create pages for each team (will require adding new endpoints to odota/core but we have the data)

Things we might want to display:
* Players registered to the team
* Recent games played by the team
* Heroes played by the team

spudde123:
fixes #1142 

Working on #1142. Initial pages for teams are working with recent games, current members and most played heroes. I put also some links to point to the team pages in the team rankings and player search page but at least match pages still need them.

Thinking if I should try to put all the data available in tabs similarly as in player pages?


howardchung:
http://cloud-3.steamusercontent.com/ugc/543025270456493033/7F92FF439CEE7E0745519538E2EE3E21EBA6BFEC/

(there's an error when viewing on https, we might need to add a proxying layer in the odota/core code (like how we do the item/hero images) to make it work on https

We should add logos to the opposing team column (will need the same image proxying)

Round the displayed rating to the nearest integer (or Math.floor)

We should add player images from the client (need to extract from the client using gcfscape).

We'll want tabs in order to be able to list more than the 10 most played heroes/20 most played matches.  That shouldn't be very hard to add, you can use the existing tabbed pages as an example.

Otherwise, looks great!

spudde123:
The action is named 'teamPlayers' atm, for this overview page I was just filtering the array to display only the current members. I thought I'd put all the previous members available in a tab because some teams have a lot of them.

I noticed that the player data isn't that great in some cases. For example https://api.opendota.com/api/teams/5039050/players gives the players for Team Kinguin, but it isn't of much help. I haven't looked at the core side yet to see how the data gets updated

spudde123:
I'm testing the proxy layer in my fork of the core and got it working for now. How would you prefer the endpoint to work?

The urls are of the form http://cloud-3.steamusercontent.com/ugc/543025270456493033/7F92FF439CEE7E0745519538E2EE3E21EBA6BFEC/ so I thought I would make it use
'api.opendota.com/teamimages/543025270456493033/7F92FF439CEE7E0745519538E2EE3E21EBA6BFEC/' for example. It seems all the logo urls start with 'http://cloud-3.steamusercontent.com/ugc/'. Can we trust that to be the case in the future? I thought of using the entire url but then I could presumably make a request anywhere if I wanted to type something in.

I'm a bit of a novice in this regard so thanks for assistance.

howardchung:
I would suggest keeping the path the same as the original (/ugc/*).

Yes. seems like all of the logos have that hostname.

spudde123:
The team logos should be working now.

I also added the player images. I had some issues with displaying the empty portrait in case the player doesn't have an image available. Not sure if it's possible to nicely check if the asset exists in the assets folder. My current solution is to have a list of the account_ids that have a portrait to know what image to display.

There's quite a lot of players (or smurfs) that don't have a name on the notable players list. It might be good to give out also the current account name and avatar from /teams/players so that those accounts can be shown nicely.

howardchung:
I think we should add the image/png header to the odota/core proxy as well.  Also we should scale the images to make sure they maintain aspect ratio and are a consistent size:

(different sizes)
https://deploy-preview-1416--opendota-stage.netlify.com/teams/1838315
https://deploy-preview-1416--opendota-stage.netlify.com/teams/39

(aspect ratio)
https://deploy-preview-1416--opendota-stage.netlify.com/teams/1838315
Team Liquid logos in the opposing team column are stretched

src/components/Team/proPlayerImages.js
```diff
@@ -0,0 +1,189 @@
+export default ([
```
howardchung: suggest moving this to a manifest.json file in the images directory so that the manifest and images are in one place,

You can then require() the file from there.

src/components/Team/teamPages.jsx
```diff
@@ -0,0 +1,73 @@
...
+    >
+      <Table
+        columns={heroColumns}
+        data={heroData.data}
+        paginated
+      />
+    </Container>
+  ),
+}, {
+  name: 'Members',
```
howardchung: move to string file

howardchung: it only needs to go in the en-US file

src/components/Team/teamPages.jsx
```diff
@@ -0,0 +1,73 @@
...
+  key: 'members',
+  content: (generalData, matchData, heroData, playerData) => (
+    <div>
+      <Container
+        title="Current Members"
```
howardchung: move to string file, ditto for rest of the strings in this file

howardchung:
If you used a script to extract the png files from the vtex_c files could you check that in the `dev` folder as well?

howardchung:
As for providing the current player selected public name, we can do that in odota/core, just need to add a LEFT JOIN to the query.

spudde123:
I used the Decompiler from https://github.com/SteamDatabase/ValveResourceFormat

Usage was simply './Decompiler -i input_folder -o output_folder'.

howardchung:
Interesting, didn't know about that tool.  Perhaps you could add a page to the github wiki with a quick note?

spudde123:
There seems to be a [page](https://github.com/odota/web/wiki/Extracting-image-assets-from-game-client) that at least guides people to the right place

howardchung:
Oh yeah, I forgot the images are outside the src folder.  That's fine, we can keep your existing solution and just manually update it when there are new images added.

src/components/Team/teamDataColumns.jsx
```diff
@@ -0,0 +1,99 @@
...
+  const won = row.radiant_win === row.radiant;
+  const string = won ? strings.td_win : strings.td_loss;
+  const textColor = won ? constants.colorGreen : constants.colorRed;
+  return (
+    <span style={{ color: textColor }}>{string}</span>
+  );
+};
+
+const getPlayerImageUrl = (accountId) => {
+  if (proPlayerImages.indexOf(accountId) !== -1) {
```
howardchung: for large N using an object is faster than indexOf on an array, but with a static N here it may not be an issue.

howardchung:
LGTM, we should add the content-type header and optionally the extra player data to odota/core.

It would also be nice if we could fix the aspect ratio issue here, but it looks like the source images are stretched, so that may be difficult.

spudde123:
I'm a bit of a novice here again but concerning the content-type header, what should be done? The response is to be of the type application/octet-stream

howardchung:
It should be image/png (or whatever format the images are in).

howardchung:
I made the change in this commit:
https://github.com/odota/core/commit/f1c6f596ec35383c62b0c1f6fae542d5c072f458

PR #1426, opened 2018-01-13
Return back service worker which generated by react-scripts

Luchanso:
Now service worker make unnecessary optimisation for API fetches - it's not good, because we make duplicate requests.
image

Service worker from react-scripts very good configure.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -148,17 +148,5 @@
     <div class="sk-cube sk-cube9"></div>
   </div>
   </div>
-  <script>
-    if ('serviceWorker' in navigator) {
-        navigator.serviceWorker
-          .register('/sw.js')
-          .then(function() {
-            console.log('Service worker registered!');
-          })
-          .catch(function(err) {
-            console.log(err);
-          });
-      }  
-  </script>   
 </body>
 </html>
--- a/src/index.js
+++ b/src/index.js
@@ -9,8 +9,7 @@ import { getMetadata } from 'actions';
 import App from 'components/App';
 import constants from 'components/constants';
 import { injectGlobal } from 'styled-components';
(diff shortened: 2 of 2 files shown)


howardchung:
@superoo7 implemented the custom service worker in https://github.com/odota/web/pull/1344

We had an issue with the default CRA service worker not working after some updates. 

Would we have to unregister the existing service worker before deploying this to avoid conflicts?

Luchanso:
Yes, of course, we need disable old service worker. I will disable old service worker in this PR.

superoo7:
I am quite busy recently, but will try to find some time to check it out.

Some general service worker specification.
- Service Worker's file name should not never be changed.
- If name does not change, when service worker change any slightest of code, it will update automatically.

What I did in my previous PR:
- I write the service worker code manually without any tools like workbox
- I created register event, and activate event.

superoo7:
some side notes about service worker generated by react-script, it used `SWPrecacheWebpackPlugin` to generate service worker file, which is the stable and older tools to generate service worker.

You can read through what is written in previous [suggestion](https://github.com/odota/web/issues/1266)

howardchung:
I would still like to use the CRA service worker if we can figure out what the issue is (convention over configuration).

The issue seems to stem from the JS bundle name changing between updates, but this is something that create-react-app already does (putting the hash into the file name to prevent stale fetches), so I'm surprised that it's not working.  Maybe something has changed in the last few months to make it more consistent?

Luchanso:
I think this PR is not so good idea, because Dan says service worker is have some pitfalls https://twitter.com/dan_abramov/status/954146978564395008 and have some issue https://github.com/facebookincubator/create-react-app/issues/2554

PR #1427, opened 2018-01-20
fixed firstblood bug

### LINKED ISSUE #1424: Story showing wrong first blooded hero
https://www.opendota.com/matches/3677057975/story

In that match Earthshaker died first mid lane to Shadow Fiend. Story tells that Phantom Lancer died which is wrong. You can check replay or here is the screenshot of it happening:
first_blood

mdiller:
Fixed firstblood bug if the creeps get a last hit. fixes  #1424

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -227,13 +227,18 @@ class IntroEvent extends StoryEvent {
 }
 
 class FirstbloodEvent extends StoryEvent {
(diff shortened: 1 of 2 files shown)


mdiller:
I fixed most of the lint errors, but there is one that isn't from this pull request that looks like its built into a function, so I'm a bit hesitant to touch it. How did it get submitted in the first place?

howardchung:
lint looks ok on travis, netlify preview might just be outdated for some reason

PR #1429, opened 2018-01-22
Fix google analytics w/ page view on route change

albertcui:


src/index.js
```diff
@@ -156,12 +158,19 @@ li {
...
+const history = createHistory();
+history.listen((location) => {
+  ReactGA.pageview(location.pathname);
+});
+
 const rootElement = document.getElementById('root');
 const app = (
   <Provider store={store}>
-    <BrowserRouter>
+    <Router history={history}>
```
howardchung: does BrowserRouter not work here?

albertcui: No.  BrowserRouter does not take history.

PR #1432, opened 2018-01-27
MatchGraph: Layout Improvements to small screen ux

oliverdon:
* Inset Y Axis key to allow more space for graph
* Tighten left and right margins to allow more space for graph
* Remove X Axis key interval so it is auto selected for available space
* Remove X Axis ‘Time’ label, key is already rendered in time format and Label was unstyled and overlapped by Key

Examples:

mobile-old
mobile-new


oliverdon:
desktop-old
desktop-new

PR #1444, opened 2018-02-09
Uncalibrated MMR shows as 0 on MMR graph

Argn0:
fixes https://github.com/odota/web/issues/1443

example: https://www.opendota.com/players/129854531/mmr

### DIFF
--- a/src/components/Visualizations/Graph/MMRGraph.jsx
+++ b/src/components/Visualizations/Graph/MMRGraph.jsx
@@ -19,11 +19,11 @@ user-select: none;
 `;
 
 const filterNullValues = function (columns) {
-  const c = columns
-  c.solo_competitive_rank = c.solo_competitive_rank ? c.solo_competitive_rank : undefined
-  c.competitive_rank = c.competitive_rank ? c.competitive_rank : undefined
-  return c
-}
+  const c = columns;
+  c.solo_competitive_rank = c.solo_competitive_rank ? c.solo_competitive_rank : undefined;
+  c.competitive_rank = c.competitive_rank ? c.competitive_rank : undefined;
+  return c;
+};
 
 const formatXTick = (time) => {
   const date = new Date(time);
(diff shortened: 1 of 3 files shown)


src/components/Player/Pages/MMR/MMR.jsx
```diff
@@ -7,6 +7,15 @@ import strings from 'lang';
 import Container from 'components/Container';
 import Info from 'components/Alerts/Info';
 
+const filterNullValues = function (columns) {
```
howardchung: suggest doing this via .map instead

src/components/Player/Pages/MMR/MMR.jsx
```diff
@@ -17,7 +26,7 @@ const MMRInfo = (
 const MMR = ({ columns, error, loading }) => (
   <div>
     <Container title={strings.heading_mmr} subtitle={MMRInfo} error={error} loading={loading}>
-      <MMRGraph columns={columns} />
+      <MMRGraph columns={filterNullValues(columns)} />
```
howardchung: actually investigate inside the mmrgraph method, it may make more sense to filter in there.

Argn0:
We could also connect the lines instead of having a gap.

howardchung:
0's not really a valid value, so we shouldn't display it at all (really, we shouldn't even ingest the value in core)

src/components/Visualizations/Graph/MMRGraph.jsx
```diff
@@ -18,8 +18,8 @@ const StyledGraphArea = styled.div`
 user-select: none;
 `;
 
-const filterNullValues = function (columns) {
-  const c = columns;
+const filterZeroValues = function (column) {
```
howardchung: use an arrow function

src/components/Visualizations/Graph/MMRGraph.jsx
```diff
@@ -18,8 +18,8 @@ const StyledGraphArea = styled.div`
...
+  const c = column;
```
howardchung: suggest `return {...column, solo_competitive_rank: c.solo_competitive_rank || null...`

howardchung: don't need the const c = column

howardchung:
lint issue

PR #1448, opened 2018-02-17
Drafts page

### LINKED ISSUE #1434: add draft_timings data to UI
@bobsaget4lyfe added draft timings data in https://github.com/odota/core/pull/1560

We can display that data on pages where:
* the draft_timings field exists
* the picks_bans field does not exist

njgingrich:
fixes #1434 

Adds a draft page to Captain's Mode games which shows time taken for each draft along with time left over.

Example draft screenshot: 
localhost_3000_matches_3734512925_drafts

It's still pretty rough and there's some cleanup I'm still doing (linting, etc), but I figured it would be good to get an initial picture out for some feedback, especially on design.


src/components/Match/Draft/Draft.jsx
```diff
@@ -0,0 +1,243 @@
...
+  extraTime: PropTypes.number,
+};
+
+// one-based indexing (since draft[i].order starts at 1)
+const orderOne = [1, 3, 5, 7, 10, 12, 14, 16, 18, 20, 21];
+const orderTwo = [2, 4, 6, 8, 9, 11, 13, 15, 17, 19, 22];
+const picks = [7, 8, 9, 10, 15, 16, 17, 18, 21, 22];
+
+const Draft = ({ radiantTeam, direTeam, draft }) => {
+  const firstIsTeamTwo = draft[0].active_team === 2;
```
howardchung: check if draft[0] is undefined?

src/components/Match/matchPages.jsx
```diff
@@ -35,6 +36,19 @@ import Chat from './Chat';
 import { StyledFlexContainer, StyledFlexElement } from './StyledMatch';
 
 const matchPages = [Overview, {
+  name: strings.tab_drafts,
+  key: 'draft',
+  content: match => (
+    <div>
+      {match.game_mode === 2 &&
```
howardchung: what happens if the match is not game mode 2 (captains mode?)  Does it just show a blank page?

src/components/Match/matchPages.jsx
```diff
@@ -35,6 +36,19 @@ import Chat from './Chat';
...
+        <Draft
+          radiantTeam={match.radiant_team}
+          direTeam={match.dire_team}
+          draft={match.draft_timings}
```
howardchung: This won't exist for all matches, so ensure the component handles the cases where the draft_timings are null/undefined (unparsed matches)

howardchung:
@blukai for design feedback

src/components/Match/matchPages.jsx
njgingrich: Yea, I'm not sure what to display in that case, since atm draft_timings only returns data for all picks. I'd rather just not show the page at all (unless browsing a CM game), but wasn't sure how to do that.

src/components/Match/Draft/Draft.jsx
```diff
@@ -144,8 +144,13 @@ const orderOne = [1, 3, 5, 7, 10, 12, 14, 16, 18, 20, 21];
...
 
-const Draft = ({ radiantTeam, direTeam, draft }) => {
-  const firstIsTeamTwo = draft[0].active_team === 2;
+const Draft = ({
+  radiantTeam = {},
+  direTeam = {},
+  draft = [],
+}) => {
+  // if there is no draft data there is no meaning to firstIsTeamTwo
+  const firstIsTeamTwo = draft[0] && draft[0].active_team === 2;
```
howardchung: check draft as well just in case it's null (I think default [] does not replace null, only undefined)

src/lang/bg-BG.json
```diff
@@ -593,6 +593,7 @@
   "tab_wordcloud": "Wordcloud",
   "tab_mmr": "MMR",
   "tab_rankings": "Rankings",
+  "tab_drafts": "Draft",
```
howardchung: You only need to add to en-US file, the rest are automatically filled later

howardchung:
We should probably display some kind of message if the draft_timings data does not exist.

src/lang/bg-BG.json
njgingrich: Removed.

njgingrich:
OK, right now it will display just the ban nominations for non-CM games. (I can restrict that to just all-pick, I'm not sure what else to do. I'd rather just not show the draft page in the tabs if it's not CM/AP, if that's possible). If the draft_timings data from core gets updated we can show which ban nominations were actually banned, but now there's no way to tell AFAIK.

src/lang/bg-BG.json
```diff
@@ -1114,4 +1114,4 @@
   "rune": "Rune",
```
howardchung: looks like there's still newline changes on the localization files

howardchung:
it's possible to customize the pages based on the match, but you'll need to change the matchPages array to a function, and everything that uses matchPages needs to call the function instead of just directly using the array.

howardchung:
Looks like there's an issue when trying to view a match without draft_timings:
https://deploy-preview-1448--opendota-stage.netlify.com/matches/3192225582/draft

njgingrich:
Fixed that issue, and I've changed `disabled` on matchPages to be a function (the TabBar component checks that tab.disabled exists). I wasn't sure how to change the matchPages itself into a function, maybe I'm just not understanding exactly how you meant. Right now with the changes I've made the Draft tab is visible, but disabled unless the match is a Captain's Mode match.

howardchung:
hmm, still seeing a javascript error on the match linked above

njgingrich:
You're right, for whatever reason I thought I'd fixed it but it should *actually* be fixed now. (In that it will just show nothing), but the draft page shouldn't be accessible unless directly accessing the URL unless the draft does exist.

PR #1450, opened 2018-02-19
Retain search query in search bar after refresh

jayantak:
fixes https://github.com/odota/web/issues/1447

Now value of search bar will be filled before mount with the value of the search query in the parameter.

### DIFF
--- a/src/components/Search/SearchForm.jsx
+++ b/src/components/Search/SearchForm.jsx
@@ -23,6 +23,9 @@ class SearchForm extends React.Component {
     const { pathname } = window.location;
     if (params.q && pathname === '/search') {
       this.props.dispatchSearch(params.q);
+      this.setState({
+        query: params.q,
+      });
     }
   }
 


src/components/Search/SearchForm.jsx
```diff
@@ -23,6 +23,9 @@ class SearchForm extends React.Component {
     const { pathname } = window.location;
     if (params.q && pathname === '/search') {
       this.props.dispatchSearch(params.q);
+      this.setState({
```
howardchung: i think you can just remove the `     if (pathname === '/search') {` from the this.debouncedSetQuery() below

jayantak:
Interestingly, when I tried that, I got an error saying debounce expects a function. 
image

I think the debounce that is done may be in an earlier context, where the function does not exist yet (from the props), and this check was added to only do it after the search context was received. 

https://stackoverflow.com/a/28046731 (Don't like to use stackoverflow as source, but this guy gives a decent explanantion)

I'm sorry if I've misunderstood something, I don't have a lot of experience with debouncing. Maybe this is not the issue at all.

src/components/Search/SearchForm.jsx
jayantak: Interestingly, when I tried that, I got an error saying debounce expects a function. 
image

I think the debounce that is done may be in an earlier context, where the function does not exist yet (from the props), and this check was added to only do it after the search context was received. 

https://stackoverflow.com/a/28046731 (Don't like to use stackoverflow as source, but this guy gives a decent explanantion)

I'm sorry if I've misunderstood something, I don't have a lot of experience with debouncing. Maybe this is not the issue at all.

howardchung: Try wrapping it in `if (this.props.dispatchSetQuery)` to ensure it's defined

jayantak: Oh, yeah, sure. Could do that. Is that fine from a design perspective?

jayantak: So that doesn't seem to be working either. It's not recognising the dispatch function as a function. 

I thought this issue was not related to debouncing, only involved setting state at component mount, since the props will have the query.

howardchung: yeah, looks like this will be fine.  Thanks!

PR #1451, opened 2018-02-21
party size filter

Argn0:
closes https://github.com/odota/web/issues/1403

https://github.com/odota/core/pull/1613

### DIFF
--- a/src/components/Player/TableFilterForm/index.jsx
+++ b/src/components/Player/TableFilterForm/index.jsx
@@ -28,7 +28,7 @@ const Styled = styled.div`
 .showForm {
   overflow: hidden;
   transition: max-height 0.2s;
-  max-height: 1000px;
+  max-height: 1200px;
 }
 `;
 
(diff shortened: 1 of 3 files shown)


src/components/Player/TableFilterForm/index.jsx
```diff
@@ -204,6 +204,15 @@ class TableFilterForm extends React.Component {
...
               limit={1}
             />
+            <FormField
+              name="party_size"
+              label={strings.filter_party_size}
+              dataSource={data.partySize}
+              formSelectionState={formSelectionState}
+              history={history}
+              strict
+              limit={5}
```
howardchung: I think this should be 1

Argn0: But what if someone wants all games with party size 2 and party size 1 for example?

Argn0: Nevermind, that doesn't work probably.

src/components/Player/TableFilterForm/index.jsx
```diff
@@ -28,7 +28,7 @@ const Styled = styled.div`
 .showForm {
   overflow: hidden;
   transition: max-height 0.2s;
-  max-height: 1000px;
+  max-height: 1200px;
```
howardchung: will this just fail if we add more forms?  What if we just remove it?

Argn0: OK, removing it works. I tested it.

Issue #1446, opened 2018-02-15
Pagination should be added to URL as query string.


jojo-tutor:
In this link: https://www.opendota.com/teams/2163/matches, on the **"Matches"** and **"Heroes"** Tab, there are pagination, if you should go to other pages, you must add the current page to the URL, so that whenever the user will refresh the browser/share the URL, the current page will retain accordingly. This is a good practice. And also, in pagination, it's good to have an option like: **results per page**.

![opendota](https://user-images.githubusercontent.com/18729213/36236252-155ebbae-1230-11e8-9bde-ec67501dc0f7.jpg)

howardchung:
Yes, I think we can definitely add a page parameter that's bound to the PaginatedTable component and saves the state in the URL

gar-stauffer:
Hey, I'll pick this up. I still need to set up the OD environment; will work on this issue over course of week

coreymaher:
The problem with using the page number in the url when sorting descending is it gets out of date very quickly. You may be trying to share a link for a match on page 2, but tomorrow the match has been pushed down to page 3.

Is this a common use case that you want to support? If so, it's probably better to store something that can generate the same page list each time, for example the topmost match id.

PR #1459, opened 2018-02-27
Issue/1446 add pagination info to url

gar-stauffer:
This is in response to issue #1446 where we want to be able to track the current page and size in the URL. This makes it possible to share the current page with others. Not totally familiar with how React works, but this was P much out of react land. 

Something worth noting is that this PR now allows users to be able to change the Pagination size, if they add `pageSize` to the URL. I can comment out that code for the meantime, since there's currently no comparable feature in the UI.

Bug Found: In doing this, I found a bug that state of the Table is saved between Matches and Heroes tab, meaning if you select page 155 on Matches and go to Heroes, it will result in no results being shown. Will file an issue and solve on a separate ticket. (Ticket is #1460 )

Would like some feedback as to where to put the query keys. They're in the constants.js file, but that looks like its more for CSS styling... can create a new file or something if there isn't one already.

### DIFF
--- a/src/components/TabBar/TabBar.jsx
+++ b/src/components/TabBar/TabBar.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Link } from 'react-router-dom';
 import constants from '../constants';
+import querystring from 'querystring';
 
 const StyledMain = styled.main`
   position: relative;
@@ -47,6 +48,13 @@ const StyledSection = styled.section`
   }
 `;
 
+const filteredURLSearchParams = () => {
+  const PARAMS_TO_IGNORE = [constants.PAGINATION_PAGE_NUMBER, constants.PAGINATION_PAGE_SIZE];
+  const params = querystring.parse(window.location.search.substring(1));
+  PARAMS_TO_IGNORE.forEach(key => delete params[key]);
+  return querystring.encode(params);
+};
+
 const TabBar = ({ tabs, info }) => (
   <StyledMain>
     <StyledSection>
@@ -54,7 +62,7 @@ const TabBar = ({ tabs, info }) => (
         <Link
           key={`${tab.name}_${tab.route}_${tab.key}`}
           className={tab.key === info ? 'chosen' : ''}
-          to={tab.route + window.location.search}
+          to={tab.route + filteredURLSearchParams()}
           disabled={tab.disabled}
         >
           {tab.name}
--- a/src/components/TabBar/TabBar.jsx
+++ b/src/components/TabBar/TabBar.jsx
@@ -2,8 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Link } from 'react-router-dom';
-import constants from '../constants';
 import querystring from 'querystring';
+import constants from '../constants';
 
 const StyledMain = styled.main`
   position: relative;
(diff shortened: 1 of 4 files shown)


src/components/TabBar/TabBar.jsx
```diff
@@ -47,14 +48,21 @@ const StyledSection = styled.section`
   }
 `;
 
+const filteredURLSearchParams = () => {
+  const PARAMS_TO_IGNORE = [constants.PAGINATION_PAGE_NUMBER, constants.PAGINATION_PAGE_SIZE];
+  const params = querystring.parse(window.location.search.substring(1));
+  PARAMS_TO_IGNORE.forEach(key => delete params[key]);
+  return '?' + querystring.encode(params);
```
dkwl93: String interpolation?

dkwl93:
Builds are failing too?

src/components/TabBar/TabBar.jsx
gar-stauffer: Yeah, it looks like this was causing the build to fail, _should_ be fixed now

src/components/TabBar/TabBar.jsx
```diff
@@ -47,14 +49,21 @@ const StyledSection = styled.section`
   }
 `;
 
+const filteredURLSearchParams = () => {
+  const PARAMS_TO_IGNORE = [PAGINATION_PAGE_NUMBER, PAGINATION_PAGE_SIZE];
+  const params = querystring.parse(window.location.search.substring(1));
```
howardchung: I think you can just do `delete params[PAGINATION_PAGE_NUMBER]` and `delete params[PAGINATION_PAGE_NUMBER]` here and skip the overhead of the array creation and .forEach function

src/components/Table/Table.jsx
```diff
@@ -53,8 +55,31 @@ const toUnderline = (data, row, field, underline) => {
   return false;
 };
 
+const getParamFromURL = (key) => {
+  const params = querystring.decode(window.location.search.substring(1));
```
howardchung: isn't the function `parse`?  Is it aliased?

https://nodejs.org/api/querystring.html

src/components/TabBar/TabBar.jsx
```diff
@@ -2,7 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Link } from 'react-router-dom';
+import querystring from 'querystring';
 import constants from '../constants';
+import { PAGINATION_PAGE_NUMBER, PAGINATION_PAGE_SIZE } from '../keys';
```
howardchung: I think you can just use constants file instead of making a new file

src/components/Table/Table.jsx
```diff
@@ -8,12 +8,14 @@ import {
   TableRowColumn as MaterialTableRowColumn,
```
howardchung: If possible, see if you can get this to work with React Router.

See https://github.com/odota/web/blob/master/src/components/Form/FormField.jsx for an example of a component that updates the URL via React Router.

You'll need the `history` object passed to this component, so you may need to add that as a prop where `Table` is used and you need pages

Once you do that you can just call history.push() to set the URL and I think you can also read out params from the URL from that object without writing your own url parsing code.

gabimoncha: if you go in matches/match.id/log and if you enter in the `FormField name='type'` or `name='players'` anything, it won't save the request in the URL

howardchung: yeah, I don't think the log UI is connected to the URL right now.  That's a separate issue, though

PR #1463, opened 2018-02-28
Add visual depth to Turbo games filter button

Argn0:


howardchung:
IMO this doesn't really make the icon look more clickable. . . maybe we should look for a different approach?

Argn0:
You're right. Maybe I can come up with something better.

PR #1469, opened 2018-03-05
mobile app links in footer

Argn0:
https://github.com/odota/web/issues/1467


src/components/Footer/Footer.jsx
```diff
@@ -168,6 +168,16 @@ export default () => (
           </a>
         </small>
         <StyledHr />
+        <a href="https://play.google.com/store/apps/details?id=com.opendota.mobile&pcampaignid=MKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1">
+          <img alt="Get it on Google Play" src="https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png" height="46px" />
```
howardchung: is there a reason one image is hotlinked and one is not?

Argn0: It's because google officially provides this link for badges and apple does not. Apple only provides a download for the badge.

howardchung: I feel we should be consistent, if we need to use a local image for one then let's do it for both.  (also this prevents breakage if google ever moves their image)

howardchung:
Also could we move the icons to the right of the existing row rather than making the footer taller?

Argn0:
Ok I will move them to the right

Argn0:
OK done. It also wraps correctly on small screens.

PR #1472, opened 2018-03-07
Highlight logged in user's row in team tables with a blue tinge

Argn0:


howardchung:
Doesn't seem very noticeable to me, but maybe that was the intention?

Argn0:
I intentionally made it very subtle. Just barely enough so it catches your eye.

It also depends on your monitor and monitor calibration I guess. It's a lot less noticeable on my second monitor.

src/components/Match/TeamTable.jsx
```diff
@@ -3,9 +3,27 @@ import PropTypes from 'prop-types';
 import { isRadiant, getTeamName } from 'utility';
 import Heading from 'components/Heading';
 import { IconRadiant, IconDire } from 'components/Icons';
+import { connect } from 'react-redux';
+import styled from 'styled-components';
 import Table from 'components/Table';
 import PicksBans from './Overview/PicksBans'; // Displayed only on `Overview` page
 
+const StyledDiv = styled.div`
+  ${props => (props.user !== undefined ? `
```
howardchung: check for null too

src/components/Match/TeamTable.jsx
```diff
@@ -19,22 +37,31 @@ const TeamTable = ({
...
+  const user = players.find(player => player.account_id === loggedInId);
+
+  return (
+    <StyledDiv user={user}>
+      <Heading
+        title={`${getTeamName(radiantTeam, true)} - ${heading}`}
+        icon={<IconRadiant />}
+      />
+      <span id="radiant">
+        <Table data={filterMatchPlayers(players, 'radiant')} columns={columns} summable={summable} hoverRowColumn={hoverRowColumn} />
```
howardchung: since this only affects the table, i'd prefer the table handle the styling/logic rather than wrapping this whole thing in the styled div.  You can pass down the ID to the table, or a function that lets it determine which row should be highlighted.

Argn0:
I also made the effect very slightly more noticeable.

src/components/Match/TeamTable.jsx
```diff
@@ -3,9 +3,16 @@ import PropTypes from 'prop-types';
...
 import Heading from 'components/Heading';
 import { IconRadiant, IconDire } from 'components/Icons';
+import { connect } from 'react-redux';
 import Table from 'components/Table';
 import PicksBans from './Overview/PicksBans'; // Displayed only on `Overview` page
 
+function highLightUser(loggedInId, players, side) {
+  const user = players.find(player => player.account_id === loggedInId);
+  const radiant = user && isRadiant(user.player_slot);
+  return user && (side === radiant) && (user.player_slot + 1 - (!radiant * 128));
```
howardchung: what is the third part?  not sure I understand it.

src/components/Table/Styled.jsx
```diff
@@ -92,6 +92,17 @@ export const StyledBody = styled.div`
       }
     }
   }` : '')};
+
+  ${props => (props.highlightUser !== undefined && props.highlightUser !== null ? `
```
howardchung: to make this more readable maybe the style should just be applied directly to the row?  nesting css selectors makes the code more difficult to understand.

src/components/Match/TeamTable.jsx
Argn0: it's to determine the row with the player slot.

128 + 1 - (1 * 128) = 1

Argn0: or if its radiant then

0 + 1 - (0 * 128) = 1

Argn0:
Style is now directly applied on the row and I made the effect even more noticable.

src/components/Match/TeamTable.jsx
```diff
@@ -19,19 +26,20 @@ const TeamTable = ({
...
   hoverRowColumn = false,
+  loggedInId,
 }) => (
   <div>
     <Heading
       title={`${getTeamName(radiantTeam, true)} - ${heading}`}
       icon={<IconRadiant />}
     />
-    <Table data={filterMatchPlayers(players, 'radiant')} columns={columns} summable={summable} hoverRowColumn={hoverRowColumn} />
+    <Table data={filterMatchPlayers(players, 'radiant')} columns={columns} summable={summable} hoverRowColumn={hoverRowColumn} highlightUser={highLightUser(loggedInId, players, true)} />
```
howardchung: I think highlightUser should be highlightFn and passed as a function that the table runs each row through to determimne its' styling.

e.g. `const getHighlightFn = (loggedInId) => (row) => return row.account_id === loggedInId`

src/components/Table/Table.jsx
```diff
@@ -131,6 +133,7 @@ class Table extends React.Component {
       placeholderMessage,
       pageLength = 20,
       hoverRowColumn,
+      highlightFn = function () {},
```
howardchung: I don't think you should construct a new function by default, just let it be null and check for null before you try to run the function

PR #1478, opened 2018-03-11
Another attempt at trying to make the Turbo filter button appear clickable

Argn0:
just a proposal


howardchung:
hmm, I think the styling could probably be improved.  I think the current logo is fine.

PR #1480, opened 2018-03-13
Fallback image for inflictors (Dota 2 Logo)

Argn0:
placeholder


Argn0:
the zIndex change is to revert something I broke in another PR

PR #1493, opened 2018-03-23
Fix inflictor image size

Argn0:


src/components/Match/StyledMatch.jsx
```diff
@@ -276,7 +276,7 @@ export const StyledLogFilterForm = styled.div`
 
 export const StyledDmgTargetInflictor = styled.div`
   .inflictorWithValue {  
-    img {
+    object {
```
howardchung: are we using an object tag now?

Argn0: yes, because img elements only allow text as fallback content.

PR #1500, opened 2018-03-26
Add option to generate unique component keys for table rows

Argn0:
fixes https://github.com/odota/web/issues/1499

this is to prevent unexpected behavior in all tables where we want to shift around rows. like for the Teamfight table.

### DIFF
--- a/src/components/Match/TeamTable.jsx
+++ b/src/components/Match/TeamTable.jsx
@@ -7,7 +7,7 @@ import { connect } from 'react-redux';
 import Table from 'components/Table';
 import PicksBans from './Overview/PicksBans'; // Displayed only on `Overview` page
 
-const uniqueKey = row => row && row.player_slot + 1;
+const keyFn = row => row && row.player_slot + 1;
 
 const getHighlightFn = loggedInId => row => loggedInId && row.account_id === loggedInId;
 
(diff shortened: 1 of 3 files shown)


src/components/Table/Table.jsx
```diff
@@ -176,7 +178,7 @@ class Table extends React.Component {
               </MaterialTableHeader>
               <MaterialTableBody displayRowCheckbox={false} selectable={false}>
                 {data.map((row, index) => (
-                  <MaterialTableRow key={index} style={rowStyle(highlightFn, row)}>
+                  <MaterialTableRow key={(uniqueKey && uuid.v4()) || index} style={rowStyle(highlightFn, row)}>
```
howardchung: perhaps we should use uuid.v4() always?  Index is not a good key for a table row anyway.

coreymaher: Doesn't this mean every render will generate a new key?
I think we probably want to actually come up with unique keys for each row so the DOM nodes can be reused when possible.

Argn0: yea it does. I'm already looking for an alternative. I figured it out too late.

Argn0:
Teamfight is the only component where we shift row positions. So this is a reasonable solution I think.

src/components/Match/TeamTable.jsx
```diff
@@ -7,6 +7,8 @@ import { connect } from 'react-redux';
 import Table from 'components/Table';
 import PicksBans from './Overview/PicksBans'; // Displayed only on `Overview` page
 
+const uniqueKey = row => row && row.player_slot + 1;
```
howardchung: i'd probably call this keyFn for consistency

howardchung: the prop name, anyway, you could name this the same if you like

Issue #1342, opened 2017-11-27
Provide stats and metrics descriptions


mcetkovsky:
There are lot of pages, acronyms and stats which has no explanation what the stat/metric tracks and how it is calculated (if it is not clear from its title) so it is understandable for new comers or casual gamers.

For example, I don't find self-explainable the Comeback, KLA and Map pings.

howardchung:
Sure, we can add tooltips/subtitles for those.  Are there any more terms that are confusing?

mcetkovsky:
There is lots of them where I have a feeling of missing something.

Few more examples:
* I can get nothing from Fantasy tab as I have no idea what it means, what the points are, what they represent, how I stand against others etc,
* I have no idea what does the Rankings number is on heroes' Rankings tab.
* What does the Advantage number on on heroes' Players tab means (ok, it somehow involves a Wilson method but that does not say what it measures what it reflects).
* What does the (people icon) 1x means on Matches tab.
* What does Normal skill means under the Win tab, what are the other options and what they represent, how are calculated or whatever is known about.
* How does the EFF@10 is calculated, I got some idea from the tooltip but that is far from being sure. What happen when I roam between lines during the first 10 minutes of the game.
* Why are my Rankings, Peers tabs empty.

In general, I would welcome if all stats, graphs and numbers has at least the following (if it makes sense):
- What the stats measure. Ideally, even a short (single sentence) description what skill/behavior the stats represents and why it is important or not.
- What is the unit in which the stats is measured.
- Whether lower/higher/mid value is better.
- How it stands against other players / what is the "normal" value / what MMR players usually have if they have this number (median + the range for 25-75% of players having such value); probably for recent matches only as the game evolves and it is rebalanced regularly.

howardchung:
Thanks for the list!  We'll probably gradually add tooltips/subtitles in for these areas.  Feel free to submit a PR if you want to join the effort.

drigols:
I'm new to the project, could anyone guide me with this problem? I wanted to help ...

howardchung:
For the histograms we can probably add it to the subtitle part of the Heading component:
https://www.opendota.com/players/88367253/histograms/comeback

For others we can use tooltips (see existing tooltips for reference code)

bdbch:
We could also implement tooltips for more abbreviations like here for example:
https://www.opendota.com/records

Sure we could also rely on people reading the target URL which explains that **HH** stands for **Hero Healing** but it could be confusing for mobile users.

[...]

PR #1505, opened 2018-03-28
Added histogram descriptions 

maxlefebvre:
In reference to #1342

Added small descriptions to be shown in subtitle component of histograms. Some are left empty because they seem relatively obvious or I couldn't find much info on them but if anyone thinks they should be there it would be easy to implement by just filling them out.


src/lang/en-US.json
```diff
@@ -959,6 +959,36 @@
   "histograms_name": "Histograms",
   "histograms_description": "Percentages indicate win rates for the labeled bin",
 
+  "histograms_kills_description" : "",
```
howardchung: To save space I think if the string is blank then it should not be added to this file.  You can default to empty string.

maxlefebvre:
I implemented the suggested change, let me know if there is anything else I should change or if a different method would be more clear.

PR #1516, opened 2018-04-06
Scenarios page

Argn0:
closes https://github.com/odota/web/issues/1440


src/components/Scenarios/Scenarios.jsx
```diff
@@ -0,0 +1,213 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import strings from 'lang';
```
howardchung: I've swapped to relative paths for this so you'll need to sync latest master

src/components/Scenarios/ScenariosColumns.jsx
```diff
@@ -0,0 +1,80 @@
+import strings from 'lang';
+import { transformations, formatSeconds } from 'utility';
+import { inflictorWithValue } from 'components/Visualizations';
+
+const computeWinRate = row => (row.wins / row.games);
+
+const getTimeRange = (field, metadata) => {
+  let lower;
+  if (metadata.indexOf(field) !== 0) {
```
howardchung: do you want -1 here?

src/lang/en-US.json
```diff
@@ -1213,7 +1214,14 @@
   "activated": "Activated",
   "rune": "Rune",
   "placement": "Placement",
-  "exclude_turbo_matches": "Exclude Turbo matches"
+  "exclude_turbo_matches": "Exclude Turbo matches",
+
+  "item_timings": "Item Timings",
```
howardchung: I suggest prefixing these strings with scenarios_

src/reducers/index.js
```diff
@@ -48,4 +48,7 @@ export default combineReducers({
   ghPulls: reducer('ghPulls'),
   form,
   request,
+  scenariosItemTimings: reducer('scenariosItemTimings', null, false),
```
howardchung: i don't think you need null and false

src/reducers/reducer.js
```diff
@@ -1,5 +1,5 @@
-export default (type, initialData) => (state = {
-  loading: true,
+export default (type, initialData, initialLoading = true) => (state = {
```
howardchung: why is this needed?

Argn0: without this the table loading spinner is displaying before the user made the first query.

src/components/Scenarios/ScenariosColumns.jsx
Argn0: no the method can never return -1. the logic is: "if the time value is not the first one in the array then the lower time range must be 0:00. else the lower range must be the one before this.

src/reducers/reducer.js
howardchung: maybe it should make an initial query by default so the page shows something?

Argn0:
I reverted the initial loading state changes and there is an initial queue now instead

src/components/Scenarios/Scenarios.jsx
```diff
@@ -0,0 +1,214 @@
...
+
+const minSampleSize = row => row.games > 200;
+
+const fields = {
+  itemTimings: ['hero_id', 'item'],
+  laneRoles: ['hero_id', 'lane_role'],
+  misc: ['scenario'],
+};
+
+const menuItems = [{
```
howardchung: with only 3 items maybe we should use a TabBar selector instead?

src/components/Scenarios/Scenarios.jsx
```diff
@@ -0,0 +1,214 @@
...
+    }, this.updateQueryParams);
+  }
+
+  render() {
+    const { scenariosState } = this.props;
+    const { dropDownValue, formFields } = this.state;
+    const { data } = scenariosState[dropDownValue];
+    const { metadata, metadataLoading, metadataError } = scenariosState.metadata;
+
+    return (
```
howardchung: page should probably have a `Heading` component and subtitle

src/components/Scenarios/ScenariosColumns.jsx
```diff
@@ -0,0 +1,80 @@
...
+  } else {
+    lower = 0;
+  }
+  return `${formatSeconds(lower)} - ${formatSeconds(field)}`;
+};
+
+const getColumns = (f, metadata) => {
+  const columns =
+{
+  itemTimings: [{
```
howardchung: should we show the number of matches as well?

src/components/Scenarios/Scenarios.jsx
```diff
@@ -0,0 +1,214 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import ActionSearch from 'material-ui/svg-icons/action/search';
```
howardchung: I quite like the search icon, we should add it to the buttons in Meta and Explorer too :)

src/components/Scenarios/Scenarios.jsx
```diff
@@ -0,0 +1,214 @@
...
+  value: 'itemTimings',
+},
+{
+  text: strings.heading_lane_role,
+  value: 'laneRoles',
+},
+{
+  text: strings.scenarios_misc,
+  value: 'misc',
+},
```
howardchung: swapping scenario types is putting the state into loading but doesn't seem to be firing the request.

src/lang/en-US.json
```diff
@@ -1213,7 +1214,18 @@
...
+  "exclude_turbo_matches": "Exclude Turbo matches",
+
+  "scenarios_item_timings": "Item Timings",
+  "scenarios_misc": "Misc",
+  "scenarios_time": "Time",
+  "scenarios_item": "Item",
+  "scenarios_game_duration": "Game Duration",
+  "scenarios_scenario": "Scenario",
+  "scenarios_first_blood": "Team drew First Blood",
+  "scenarios_courier_kill": "Team sniped the enemy courier before the 3-minute mark",
```
howardchung: string is too long for the container (you can increase the width or shorten the string)

src/components/Scenarios/FormFieldData.js
```diff
@@ -0,0 +1,24 @@
...
+
+    itemList: Object.keys(items).map(k => [items[k], k]).filter(x => x[0].cost >= itemCost).map(x => ({
+      text: x[0].dname,
+      value: x[1],
+    }))
+      .sort((a, b) => a.text.localeCompare(b.text)),
+
+    laneRoleList: [1, 2, 3, 4].map(role => ({ text: strings[`lane_role_${role}`], value: role.toString() })),
+
+    miscList: teamScenariosQueryParams.map(scenario => ({ text: strings[`scenarios_${scenario}`], value: scenario })),
```
howardchung: misc with no filter is only showing 1 scenario

Argn0:
Fixed everything you mentioned in your last review.

src/components/Scenarios/Scenarios.jsx
```diff
@@ -0,0 +1,227 @@
...
+        data: reduceRows(groupByArray(scenariosMisc.data, 'scenario')),
+        loading: scenariosMisc.loading,
+        error: scenariosMisc.error,
+      },
+    },
+  };
+};
+
+const mapDispatchToProps = dispatch => ({
+  scenariosDispatch: {
```
howardchung: don't know if this nesting is necessary, i think you can just put these props in directly

src/components/Scenarios/ScenariosColumns.jsx
```diff
@@ -0,0 +1,83 @@
...
+    displayName: strings.scenarios_item,
+    field: 'item',
+    sortFn: true,
+    displayFn: (row, col, field) => inflictorWithValue(field),
+  }, {
+    displayName: strings.heading_win_rate,
+    field: 'games',
+    sortFn: computeWinRate,
+    percentBarsWithValue: row => row.games,
+    tooltip: strings.tooltip_winrate_samplesize,
```
howardchung: might want to abbreviate the sample size

src/lang/en-US.json
```diff
@@ -1213,7 +1215,19 @@
   "activated": "Activated",
   "rune": "Rune",
   "placement": "Placement",
-  "exclude_turbo_matches": "Exclude Turbo matches"
+  "exclude_turbo_matches": "Exclude Turbo matches",
+
+  "scenarios_subtitle": "Combination of factors that are present in a match",
```
howardchung: "Explore win rates of combinations of factors that happen in matches"

src/components/Scenarios/FormFieldData.js
```diff
@@ -0,0 +1,24 @@
+import heroes from 'dotaconstants/build/heroes.json';
+import items from 'dotaconstants/build/items.json';
+import strings from '../../lang';
+
+export default function getFormFieldData(metadata) {
+  const { teamScenariosQueryParams, itemCost } = metadata;
+
+  return {
+    heroList: Object.keys(heroes).map(id => ({
```
howardchung: maybe provide a filter for game duration as well?

src/components/Scenarios/ScenariosColumns.jsx
Argn0: do you mean the word "sample size" or the number?

src/components/Scenarios/FormFieldData.js
Argn0: game duration is not being tracked for that scenario actually. we could add it but it would increase the table by ~300k rows?

Argn0: or did you mean the lane role scenario? yeah I'll do that.

src/components/Scenarios/Scenarios.jsx
Argn0: I put them inside an object so I can call the functions dynamically. not sure if there is another way to do it.

Argn0:
Fixed everything you mentioned in your last review. Also we should probably let this sit for a week before merging.

src/components/Scenarios/Scenarios.jsx
Argn0: nevermind I changed it

src/components/Scenarios/ScenariosFormField.jsx
```diff
@@ -0,0 +1,127 @@
...
+      searchText,
+    };
+
+    this.resetField = this.resetField.bind(this);
+    this.handleUpdateInput = this.handleUpdateInput.bind(this);
+    this.handleRequest = this.handleRequest.bind(this);
+  }
+
+  componentDidMount() {
+    requestAnimationFrame(() => {
```
howardchung: does this fire many times per second?  might have a perf impact

Argn0:
-fixed issue with abbreviateNumber

src/components/Scenarios/ScenariosFormField.jsx
Argn0: it renders once , skips a frame and then renders each AutoComplete component once again. so they are rendered only one additional time.
I believe the performance loss is negligible.

the frame skip is necessary because browsers merge changes in a single frame together for optimization reasons.

howardchung:
If you prefer, we can "soft launch" this without the link in header, but deploying the code.  Once it's in production for a few weeks and you confirm there are no issues, we put back the header link and add an announcement/blog post.

Argn0:
sounds good to me. I commented out the header.

howardchung:
There's a conflict with the Header (some code got moved around), can you resolve?

Argn0:
resolved merge conflict in Header.jsx

Issue #1503, opened 2018-03-28
replace UNSAFE_ React lifecycles with new ones for React 16.3


howardchung:
As detailed in this post:
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

hawkins:
Looks like this was fixed in #1529, right?

howardchung:
No, that only renamed the names to the UNSAFE_ aliases to make them easier to find and update.

hawkins:
Whoops, my mistake; good catch. Just found the repo and was looking through issues for ways to help out, and I wasn't sure if there was work left to be done on this one. Thanks! 🙂

howardchung:
Looks like componentWillMount should be replaced with componentDidMount and componentWillUpdate should be replaced with getSnapshotBeforeUpdate (or maybe componentDidUpdate)

howardchung:
apparently these new methods are in 16.3, so we can actually start swapping them over now.

howardchung:
Updated the title to more accurately reflect the task

PR #1529, opened 2018-04-14
Rename unsafe lifecycles, and minor stylistic improvements

morgvanny:
In preparation for a more permanent solution to issue #1503 I renamed deprecated lifecycle methods. They will work equally well, but make it more clear that they are meant to be changed eventually, and might cause issues in React 17.

While I was at it, I replaced instances where we were manually binding `this` with arrow functions instead.

I also reordered methods within components, to follow [this ruleset](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.mdl) for organization's sake.


howardchung:
Thanks!  I think you have to upgrade to React 16.3 though

howardchung:
I would also suggest enabling the sort-comp lint rule in eslint config to enforce it.

morgvanny:
lol yeah I literally just realized I forgot to upgrade React. good suggestion about the eslint config as well.

morgvanny:
@howardchung is it unwise to simply run `npm update`? Would you prefer I only update react itself, or should I also update related dependencies?

also, after changing the lifecyle methods, I'm getting a lot of eslint errors like the following:
error  Identifier 'UNSAFE_componentWillUpdate' is not in camel case             camelcase

Do you know a good way to allow for exceptions for prefixes?

src/components/AccountWidget/AccountWidget.jsx
```diff
@@ -63,19 +63,24 @@ const mapStateToProps = (state) => {
   };
 };
 
+function RequestLayer(props) {
```
howardchung: it looks like this is commented out so you can probably just remove it.

howardchung: nvm, I misread the diff

howardchung: but actually you can remove the part that's commented out after the diff

morgvanny: I thought it may have been structured that way to more easily make some planned change to componentWillUpdate, so left it in. I'll go ahead and remove it

howardchung:
https://eslint.org/docs/rules/camelcase

properties: never might work, but you'd have to try it

morgvanny:
I tried a bunch of different variations - the only thing I could get to work is either "camelcase": 1 for warning, or "camelcase": 0 for no warnings

PR #1531, opened 2018-04-15
[WIP] code splitting by route

howardchung:
Splitting code into chunks by route, as described here:
https://serverless-stack.com/chapters/code-splitting-in-create-react-app.html


howardchung:
This is probably not worth the additional load time for each page change.  we could consider dynamically loading dependencies like recharts or wordcloud though

Issue #1530, opened 2018-04-15
dynamically import language files


howardchung:
We should utilize dynamic import (a recent webpack feature) in order to dynamically import only the language file that the user currently requires.

https://webpack.js.org/guides/code-splitting/#dynamic-imports

Currently we bundle all the language files into the JS bundle which leads to extra bytes downloaded/parsed.

PR #1533, opened 2018-04-15
[WIP] dynamically load lang files

howardchung:
partially #1530 

This loads the strings into the redux store, which allows dynamically loading only the necessary language file.

However, it means every component that accesses the strings (many of them) will need to connect() to the Redux store, or have strings passed to it as a prop from its parent.

The problem is that there seems to be no way to do synchronous dynamic import, so we can't just do it with an import at the top of the file like we have been doing.  Regardless of whether we read the strings from Redux store or use import() directly in the components, we'll still have this issue.

Currently implemented for the header, feedback/advice is welcome!


howardchung:
There are currently 114 imports of strings right now.  I've updated this PR to allow the strings to be imported the old way as well as the new way, so we can check in parts of this change at a time rather than have to do it all at once.

We won't get the benefits of the smaller bundle size until it's fully migrated, though.

PR #1539, opened 2018-04-19
Disable filter form toggle on Player > Overview page

Argn0:
fixes https://github.com/odota/core/issues/1636


howardchung:
I don't know if we should do this, the filtering still works on w/l, players, and heroes.

Argn0:
Whoops didn't notice that. Maybe I can come up with a better solution.

PR #1546, opened 2018-04-21
Issue #1541 move proptypes into react component classes 

### LINKED ISSUE #1541: move proptypes into react component classes
As described here:
https://michalzalecki.com/react-components-and-class-properties/

We should do something like the following:
```
// AFTER
import React, { Component, PropTypes } from "react";

class Counter extends Component {
  static propTypes = {
    step: PropTypes.number.isRequired,
  }
}
```

darutoko:
fixes #1541


howardchung:
awesome, thanks!

Issue #1407, opened 2017-12-21
use skeleton loaders to replace spinners


rezk2ll:
hello, 
I think for a non-developer, it is quite confusing to see multiple loading indicators on the profile page. The user won't tell which component is loading, and which loading indicator belongs to which component.

![image](https://user-images.githubusercontent.com/6764881/34256280-7298ccca-e64c-11e7-9c38-d090af08131a.png)

I think it would be better to replace it all by a single loading progress indicator, like a bar at the top or a spinner at the center.

thanks.

howardchung:
@blukai @Argn0 @mike-robertson is there a natural way of combining the loading states for all those components into one loading component that renders a single loading indicator?

gu3st:
Other than an if statement that checks for multiple loading states to conditionally render based on that. 

The alternative would be to use a "skeleton" loading animation rather than a spinner which might be more elegant.

uncledu:
it's better to use skeleton screen to replace the loading indicator

howardchung:
Screenshot/example?

uncledu:
refer to this:  https://github.com/dvtng/react-loading-skeleton

howardchung:
Interesting, we could probably skeleton load the tables with rows of gray boxes (to represent table rows?) and maybe something custom for the header.

Argn0:
I know a page that uses those skeleton loaders

https://www.gutefrage.net/frage/welchen-kaese-benutzt-man-in-pizzerien

gu3st:
We wrote a skeleton implementation at work, but I think I wouldn't be allowed to share that.  It was fairly simple though, when loading, return the skeleton out of the component, else return the data component.

GravityMsc:
https://github.com/danilowoz/react-content-loader
maybe this is better?

farmasek:
I like how blueprint implements it

http://blueprintjs.com/docs/#core/components/progress/skeleton

With that, you dont need to create custom svgs for different components, but it might be more complicated to implement.

Argn0:
I had this idea of filling the tables with dummy data while the real data is loading and blurring it out and overlaying it with a loading animation. I've never seen something like this though :D

howardchung:
This is partially completed (match, player pages) in #1555

howardchung:
Remaining:
* Meta page
* Matches page
* Heroes page
* Teams page
* Ranks page
* Records page
* (possibly) login widget
* Table component

PR #1552, opened 2018-04-26
replace spinners with skeletons on player page

Phrynobatrachus:
references #1407

react-content-loader seems like the most appropriate choice so these changes point in that direction. the PlayerHeader edits can be finished by writing a custom skeleton to more closely match the loaded header. the rest of the spinners on the player page are displayed primarily by the Container component, for the moment there's just the List skeleton substituted in for all children, but perhaps typechecking could be done there to load specific custom skeletons.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11754,6 +11754,11 @@
         "prop-types": "15.6.1"
       }
     },
+    "react-content-loader": {
+      "version": "3.1.1",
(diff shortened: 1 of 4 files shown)


howardchung:
Thanks!  Think the skeletons could be resized a bit to more closely fit the content.

howardchung:
Also would be cool to enable an animation effect on them like Facebook has.

howardchung:
added your commit to #1555 

Let me know if you want to work on this and I can give you write access so you can work on that branch!

PR #1553, opened 2018-04-26
Scenarios: remove recipes from items list

Argn0:
no data for those

### DIFF
--- a/src/components/Scenarios/FormFieldData.jsx
+++ b/src/components/Scenarios/FormFieldData.jsx
@@ -22,7 +22,7 @@ export default function getFormFieldData(metadata) {
     }))
       .sort((a, b) => a.text.localeCompare(b.text)),
 
-    itemList: Object.keys(items).map(k => [items[k], k]).filter(x => x[0].cost >= itemCost && !x[1].includes('recipe')).map(x => ({
+    itemList: Object.keys(items).map(k => [items[k], k]).filter(x => x[0].cost >= itemCost && !x[1].includes('recipe_')).map(x => ({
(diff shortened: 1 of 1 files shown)


src/components/Scenarios/FormFieldData.jsx
```diff
@@ -22,7 +22,7 @@ export default function getFormFieldData(metadata) {
     }))
       .sort((a, b) => a.text.localeCompare(b.text)),
 
-    itemList: Object.keys(items).map(k => [items[k], k]).filter(x => x[0].cost >= itemCost).map(x => ({
+    itemList: Object.keys(items).map(k => [items[k], k]).filter(x => x[0].cost >= itemCost && !x[1].includes('recipe_')).map(x => ({
```
howardchung: indexOf !== 0 might be a little faster

howardchung: maybe you could move the map after the filter to avoid the need to double map?

Argn0: I'm not sure if that's possible. I need to filter by property name and value. The items are stored in an object and not an array.

Argn0: I think startsWith() is even more appropriate to be used here?

howardchung: you can still use items[k] to get the item object

Argn0: whoops you're totally right. I had a total tunnel vision.

PR #1554, opened 2018-04-27
Scenarios: add info with data age

Argn0:


src/lang/en-US.json
```diff
@@ -1262,6 +1262,7 @@
   "exclude_turbo_matches": "Exclude Turbo matches",
 
   "scenarios_subtitle": "Explore win rates of combinations of factors that happen in matches",
+  "scenarios_info": "Data compiled from matches of the last %d weeks",
```
howardchung: "in the last" instead of "of the last"

PR #1556, opened 2018-04-28
Scenarios: fix bug with table page selection state

Argn0:
Finally figured out what was going on.

Page selection state of the table component would not reset on a new query. So if you selected page 3 but your result only had a single page then it would not display anything.

### DIFF
--- a/src/components/Scenarios/Scenarios.jsx
+++ b/src/components/Scenarios/Scenarios.jsx
@@ -185,7 +185,7 @@ class Scenarios extends React.Component {
           />
           <Heading title={strings.explorer_results} subtitle={`${data.filter(minSampleSize).length} ${strings.explorer_num_rows}`} />
           <Table
-            key={selectedTab}
+            key={selectedTab + JSON.stringify(data && data[0])}
             data={data.filter(minSampleSize)}
             columns={getColumns(selectedTab, metadata)}
             loading={scenariosState[selectedTab].loading}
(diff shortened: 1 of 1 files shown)


src/components/Scenarios/Scenarios.jsx
```diff
@@ -185,7 +185,7 @@ class Scenarios extends React.Component {
           />
           <Heading title={strings.explorer_results} subtitle={`${data.filter(minSampleSize).length} ${strings.explorer_num_rows}`} />
           <Table
-            key={selectedTab}
+            key={selectedTab + JSON.stringify(data && data[0])}
```
howardchung: maybe just don't supply a key?

Argn0: I tried that. It doesn't work.

howardchung: or just generate a GUID?  I don't think data[0] makes a very good key

howardchung: I'm concerned that data[0] might be the same for two different queries, which would cause the table not to update on the new query

Argn0: Wouldn't the table re-render anyway because it receives a new data prop? Not entirely sure how it works.

howardchung: not sure, react may use the key as a shortcut to determine that it should not update, even if the props change.

Argn0: Maybe this is a bug with the Table component and how the pagination works.

howardchung:
actually see https://github.com/odota/web/issues/1460

The resetTableState prop might be what you're looking for

PR #1564, opened 2018-05-03
Remove pvgna

albertcui:


src/reducers/index.js
```diff
@@ -22,9 +22,7 @@ export default combineReducers({
   playerRecentMatches: reducer('playerRecentMatches'),
   playerTotals: reducer('playerTotals'),
   metadata: reducer('metadata'),
-  match: reducer('match', {
-    players: [],
-  }),
+  match: reducer('match'),
```
howardchung: there might be code that relies on the players array being present.

PR #1566, opened 2018-05-03
Added Issue Templates

bdbch:
This is a new feature added by Github today.

Example:
https://github.com/babel/babel/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -7,18 +7,19 @@ about: If something isn't working as expected 🤔.
 **Current Behavior**
 A clear and concise description of the behavior.
 
-**Input Code**
-- REPL or Repo link if applicable:
-
-```js
-var your => (code) => here;
-```
 **Expected behavior/code**
 A clear and concise description of what you expected to happen (or code).
 
+**Console Output**
+Please provide your browsers console log if possible. Most browsers will open the console via F12.
+
+```
+please paste console output in here
+```
+
 **Environment**
-- Node/npm version: [e.g. Node 8/npm 5]
-- OS: [e.g. OSX 10.13.4, Windows 10]
+- Browser Version: [e.g. Chrome 64.0, Firefox 57, Edge 17]
+- OS: [e.g. OSX 10.13.4, Windows 10 or Linux Distribution]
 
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: 🐛 Bug Report
+about: If something isn't working as expected 🤔.
+
+---
+
+**Current Behavior**
+A clear and concise description of the behavior.
+
+**Input Code**
```
howardchung: shouldn't this be more end-user friendly?  e.g. they probably won't be able to point to the code in question

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+- REPL or Repo link if applicable:
+
+```js
+var your => (code) => here;
+```
+**Expected behavior/code**
+A clear and concise description of what you expected to happen (or code).
+
+**Environment**
+- Node/npm version: [e.g. Node 8/npm 5]
```
howardchung: browser instead of node version?

.github/ISSUE_TEMPLATE/Support.md
```diff
@@ -0,0 +1,13 @@
+---
+name: 🤗 Support Question
+about: If you have a question 💬, please check out our Discord!
```
howardchung: provide the link?

howardchung: nvm I see it

.github/ISSUE_TEMPLATE/Bug_report.md
bdbch: Yeah makes sense. How about `browser version OR Node version` in case someone has a bug with the build/linting process? Or should we redirect bug reports like this to the specific tool (webpack, loaders, etc.)?

.github/ISSUE_TEMPLATE/Bug_report.md
bdbch: Agree, I'll refactor the text a bit, think about what actually are the usecases etc.

bdbch:
@howardchung updated :+1:

PR #1568, opened 2018-05-04
Scenarios release

Argn0:


howardchung:
@Argn0 do you want to wait for the next release?  (probably in a week or so)

Argn0:
@howardchung ok

bdbch:
Could be merged now, right?

src/components/Header/Header.jsx
```diff
@@ -99,7 +99,7 @@ class Header extends React.Component {
       <Link key="header_heroes" to="/heroes">{strings.header_heroes}</Link>,
       <Link key="header_distributions" to="/distributions">{strings.header_distributions}</Link>,
       <Link key="header_records" to="/records">{strings.header_records}</Link>,
-      // <Link key={strings.header_scenarios} to="/scenarios/itemTimings">{strings.header_scenarios}</Link>,
+      <Link key={strings.header_scenarios} to="/scenarios/itemTimings">{strings.header_scenarios}</Link>,
```
howardchung: should this just be /scenarios?  Also update the key to be a string (it doesn't need  to be localized)

howardchung: yeah but i think the link should be just to /scenarios so if you change the default later you don't have to update this

Argn0:
fixed default URL

PR #1571, opened 2018-05-05
Include attribute bonuses in base stats

Argn0:


src/components/Hero/index.jsx
```diff
@@ -149,7 +150,7 @@ class Hero extends React.Component {
               {this.state.detailsOpen ? strings.hide_details : strings.show_details}
             </HeroDetailsButton>
           </HeroFooter>
-          {this.state.detailsOpen && <AttributesBlock hero={hero} />}
+          {this.state.detailsOpen && <AttributesBlock hero={compileLevelOneStats(hero)} />}
```
bdbch: I'm not sure but shouldn't this happen inside the AttributesBlock component since it's more of a "default behaviour" for stats in general?

src/utility/index.jsx
```diff
@@ -713,3 +713,79 @@ export function groupByArray(xs, key) {
 
 export const sumValues = f => Object.values(f).reduce((a, b) => a + b);
 
+/* eslint-disable camelcase */
+// https://dota2.gamepedia.com/Attributes
+export function compileLevelOneStats(hero) {
+  const statsBonuses = {
```
bdbch: fine with this but maybe we can fetch this automatically from the patch notes (whenever valve decides to change the bonuses again)?

cc @howardchung

src/components/Hero/index.jsx
Argn0: fixed

src/utility/index.jsx
howardchung: yeah idk if that's possible

bdbch: Alright

PR #1572, opened 2018-05-06
more skeleton loaders

Phrynobatrachus:
continues work on #1407, covers what @howardchung mentioned except for the login widget. I'm going to work on cleaning up the rest tomorrow.


howardchung:
were there tweaks you wanted to make?

Phrynobatrachus:
sorry been afk with rl work, getting started on swapping the hero page spinners which I think is mostly it. also if you have a preference on how these should be organized let me know, maybe I should put them in a folder like spinner?

howardchung:
Yeah, I think it makes sense to put all the loader components in a directory.

howardchung:
Also I think the gray looks better than the purple

Argn0:
Did removing the animations from the skeleton loaders actually resolve the performance issues?

PR #1576, opened 2018-05-10
Improved row&col highlighting

Argn0:


src/components/Visualizations/Table/Styled.jsx
```diff
@@ -10,6 +10,7 @@ export const StyledContainer = styled.div`
 export const KDAContainer = StyledContainer.extend`
   width: calc(300% + 50px);
   left: -10px;
+  pointer-events none;
```
howardchung: doesn't this need a colon?

Argn0: lol I actually forgot the colons but it's working without colons too for some reason

PR #1585, opened 2018-05-20
Match search and revive FourOhFour

Argn0:


src/actions/action.js
```diff
@@ -11,17 +11,30 @@ export default function action(type, host, path, params = {}, transform) {
...
+      error,
+    });
     const fetchDataWithRetry = delay => fetch(url, path === 'api/metadata' ? { credentials: 'include' } : {})
       .then((response) => {
         if (!response.ok) {
-          throw new Error('fetch failed');
+          dispatch(getError(response.status));
+          if (response.status === 404) {
+            const err = new Error('api returned 404');
+            err.code = 404;
```
howardchung: probably err.code = response.status.  Technically all 4xx codes indicate user error and should not be retried.

src/lang/en-US.json
```diff
@@ -620,6 +621,7 @@
   "rune_6": "Arcane",
 
   "search_title": "Player Search",
+  "search_title_2": "Player, Match Search",
```
howardchung: probably "Search by player name, match ID..."

howardchung: Any reason you opted to make a new string instead of replacing the existing one?

Argn0:
-do not retry on any 4xx client error
-fixed hint text

PR #1592, opened 2018-05-21
Display wardmaps next to each other 

Argn0:
with ability to display in full size on click.

https://deploy-preview-1592--opendota-stage.netlify.com/players/400603826/wardmap

fixes  https://github.com/odota/web/issues/1591

### DIFF
--- a/src/components/Player/Pages/Wardmap/Wardmap.jsx
+++ b/src/components/Player/Pages/Wardmap/Wardmap.jsx
@@ -8,6 +8,7 @@ import { getPlayerWardmap } from '../../../../actions';
 import Heatmap from '../../../Heatmap';
 import Container from '../../../Container';
 
+const MAX_WIDTH = 1200;
 
 const StyledContainer = styled.div`
   display: flex;
@@ -97,7 +98,7 @@ class RequestLayer extends React.Component {
           >
             <Heatmap
               points={unpackPositionData(data.obs)}
-              width={Math.min(this.state.clicked === 0 ? 1200 : 580, heatmapWidth)}
+              width={Math.min(this.state.clicked === 0 ? MAX_WIDTH : 580, heatmapWidth)}
               key={this.state.clicked} // force update
             />
           </Container>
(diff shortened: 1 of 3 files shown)


src/components/Player/Pages/Wardmap/Wardmap.jsx
```diff
@@ -9,21 +9,25 @@ import Heatmap from '../../../Heatmap';
 import Container from '../../../Container';
 
 const MAX_WIDTH = 1200;
+const HALF_WIDTH = 580;
```
howardchung: I wonder if we can make these percentages so we can adjust the width of the container without modifying this?

src/components/Player/Pages/Wardmap/Wardmap.jsx
```diff
@@ -44,6 +48,26 @@ class RequestLayer extends React.Component {
...
+  handleClick(mapId) {
+    this.setState({
+      clicked: !this.state.clicked && mapId,
+    });
+  }
+
+  getClickProperties(mapId) {
+    return {
+      className: this.state.clicked && this.state.clicked !== mapId ? 'heatmap-clicked' : 'heatmap',
+      onClick: this.handleClick.bind(this, mapId),
```
howardchung: I think you can avoid using bind if you make the function a class property

src/components/Player/Pages/Wardmap/Wardmap.jsx
```diff
@@ -67,26 +91,32 @@ class RequestLayer extends React.Component {
...
-          title={strings.th_ward_sentry}
-          error={error}
-          loading={loading}
-        >
-          <Heatmap
-            points={unpackPositionData(data.sen)}
-            width={Math.min(MAX_WIDTH, heatmapWidth)}
-          />
-        </StyledInner>
+        <div {...this.getClickProperties(1)}>
```
howardchung: perhaps should use strings for the mapId to make it more readable?

PR #1600, opened 2018-05-26
Fix overflowing search hinttext on small screens

Argn0:
fixes https://github.com/odota/web/issues/1598

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -137,16 +137,12 @@ class Header extends React.Component {
       </VerticalAlignToolbar>
     );
 
-    const SearchGroup = ({ small }) => (
+    const SearchGroup = () => (
       <VerticalAlignToolbar style={{ marginLeft: 20 }}>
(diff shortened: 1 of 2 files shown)


src/components/Header/Header.jsx
```diff
@@ -137,13 +137,17 @@ class Header extends React.Component {
       </VerticalAlignToolbar>
     );
 
-    const SearchGroup = () => (
+    const SearchGroup = ({ small }) => (
```
howardchung: are you actually using this prop?

Argn0: yes. SearchGroup is a parent of SearchForm. I use it in SearchForm.jsx

Argn0: or wait let me double check....

Argn0: ok it goes Redux Store -> Header -> SearchGroup > SearchForm

Argn0: this is probably bad. I should not do this :) I'm gonna get the prop in SearchForm directly.

PR #1601, opened 2018-05-28
string refs

Phrynobatrachus:
working on the remainder of #1530


howardchung:
I think it's actually fine to use connect() instead of passing strings down.

Phrynobatrachus:
yeah I suppose it probably doesn't matter performance-wise

PR #1604, opened 2018-06-01
move strings for more files

### LINKED ISSUE #1530: dynamically import language files

howardchung:
Last one!

fixes #1530 

Three files left, matchColumns, MatchStory, and utility.


howardchung:
@Phrynobatrachus The best solution I've come up with is what I did in this commit: https://github.com/odota/web/pull/1604/commits/1ccca241253221b027fa69fce5fc595073c344e5

Basically I'm making matchColumns return a function instead and then everything that uses it calls that function and passes it strings.

Downsides:
* No compile time checking of incorrectly named things
* More expensive since the function has to be invoked on every re-render

A similar approach would probably work for utility, if we want to avoid passing strings into every single function.

tagging @gu3st for any ideas on how to do this

howardchung:
I just remembered this is another option we can use:

```
const { strings } = store.getState().app;
```

This works when we place it within a function that will get called on a render update.  (e.g. it won't work to do this at the top level of a file since it will get never called again).  This could possibly help with reducing the number of times we have to pass strings down through components.

gu3st:
I didn't notice this until now.  I've tried thinking about a way to do what you suggest but I don't have a great way outside of a specific component that gets the strings.

For instance, in another project I work on, we effectively have a component that renders out 
```
({strings, [keypath]}) => { return strings[keypath[0]][keypath[1]] } 
```
Then a redux container around it to get strings from the redux store.  The `keypath` is an array of keys (in case you need to traverse a nested string path, but if it's flat that's ideal).  A simple reduce is usually sufficient.

gu3st:
Not every component, just the actual strings you'd want. But that could
also be tedious (and relies on the ability to have a component rendered
everywhere you'd want to use a string).



On Jun 3, 2018, at 3:25 PM, Howard Chung <notifications@github.com> wrote:

Does that mean every component would need to be wrapped in this new one?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/odota/web/pull/1604#issuecomment-394196509>, or mute
the thread

PR #1610, opened 2018-06-04
[WIP] Notifications for New Matches

albertcui:


src/components/App/App.jsx
```diff
@@ -28,6 +34,64 @@ import Api from '../Api';
...
+    const n = new Notification(
+      'Parsed $match'.replace('$match', payload.data.match_id),
+      {
+        body: 'Check out your performance.',
+      },
+    );
+    n.onclick = (event) => {
+      console.log(event);
+      event.preventDefault(); // prevent the browser from focusing the Notification's tab
+      window.open(`https://opendota.com/matches/${payload.data.match_id}`, '_blank');
```
howardchung: Should the host be configurable?

src/components/App/App.jsx
```diff
@@ -28,6 +34,64 @@ import Api from '../Api';
...
+messaging.onTokenRefresh(getMessagingToken);
+
+messaging.onMessage((payload) => {
+  console.log('Message received. ', payload);
+
+  if (payload.data.type === 'MATCH') {
+    const n = new Notification(
+      'Parsed $match'.replace('$match', payload.data.match_id),
+      {
+        body: 'Check out your performance.',
```
howardchung: localize?

src/components/App/App.jsx
```diff
@@ -93,6 +168,19 @@ class App extends React.Component {
...
+  constructor(props) {
+    super(props);
+
+
+    this.state = {
+      canNotify: 'Notification' in window && Notification.permission,
+      showLoginPrompt: true,
+    };
+
+    this.askForNotifiyPermission = this.askForNotifiyPermission.bind(this);
```
howardchung: Notify

src/components/App/App.jsx
```diff
@@ -114,6 +215,50 @@ class App extends React.Component {
...
+                  style={{ margin: '5px 5px' }}
+                />
+              </Prompt>
+            : <div />
+          }
+          {
+            !user && this.state.showLoginPrompt ?
+              <Prompt>
+                <img src="/assets/images/icons/icon-72x72.png" alt="logo" style={{ height: '25px', margin: '0 5px' }} />
+                <h3 style={{ display: 'inline-block', margin: '0' }}>Get Tracked</h3>
```
howardchung: "Get Tracked" may be poor phrasing with current concerns about data privacy

src/index.js
```diff
@@ -11,8 +11,7 @@ import store from './store';
 import { getMetadata, getStrings, getAbilities, getNeutralAbilities, getAbilityIds } from './actions';
 import App from './components/App';
 import constants from './components/constants';
-// import registerServiceWorker from './registerServiceWorker';
-import { unregister } from './registerServiceWorker';
+import registerServiceWorker from './registerServiceWorker';
```
howardchung: we previously had issues with the default CRA service worker.  not sure if that's still an issue

howardchung: nvm looks like you changed the path to another SW.  I would probably make a copy of that file and change the path there so we can enable the CRA service worker at some later point

src/components/App/App.jsx
albertcui: I don't know how. I was going to ask you. Strings are only within App, and this is an event listener.

howardchung: maybe you could mount this event listener within componentDidMount in App?

PR #1621, opened 2018-06-12
Fix URL to blog post re MMR not being up to date

matalo33:
Datestamp in the URL pointing at the MMR blog post is out by 1 day

Appreciate this is old news now because mmr cant be tracked this way, but the current link is 404


howardchung:
I think the proper fix is to actually remove the date from the front matter in the blog post so it doesn't overwrite the date in the file name:
https://github.com/odota/blog/commit/86d8756ba7eb93d3c6b41bc44d44ddcb9baefa5e

PR #1629, opened 2018-06-21
Parsed benchmarks

Argn0:
fixes odota/web#524


howardchung:
Note: you can add them to the heroes page too:
https://www.opendota.com/heroes/102/benchmarks

PR #1634, opened 2018-06-23
Update hero pages with new benchmarks

Argn0:


src/components/Hero/BenchmarkGraphs.jsx
```diff
@@ -17,7 +17,7 @@ const Wrapper = styled.div`
   }
 `;
 
-const mapData = (data, key) => data.map(item => ({ Percentage: (`${item.percentile * 100}%`), Value: Number(item[key].toFixed(2)) }));
+const mapData = (data, key) => data.map(item => ({ Percentage: (`${item.percentile * 100}%`), Value: typeof item[key] === 'number' && Number(item[key].toFixed(2)) }));
```
howardchung: what's the additional check for?  Are any of them not numbers?

src/components/Match/matchColumns.jsx
```diff
@@ -332,7 +332,7 @@ export default (strings) => {
         if (match.version || !parsedBenchmarkCols.includes(key)) {
           cols.push({
             displayName: strings[`th_${key}`] || strings[`heading_${key}`],
-            tooltip: strings[`tooltip_${key}`],
+            tooltip: strings[`benchmarks_${key}`] || strings[`tooltip_${key}`],
```
howardchung: I think we should remove the benchmarks_ strings and replace them with tooltip_ strings (many already exist)

src/components/Match/matchColumns.jsx
```diff
@@ -332,7 +332,7 @@ export default (strings) => {
         if (match.version || !parsedBenchmarkCols.includes(key)) {
```
howardchung: do you need to update parsedBenchmarkCols for stuns_per_min?

src/components/Hero/BenchmarkGraphs.jsx
Argn0: Without this the page will keep crashing for a while after we add new benchmarks

Argn0:
-removed all the benchmarks strings that were added recently
-updated parsedBenchmarkCols array
-added new string for stuns_per_min

howardchung:
https://stage.opendota.com/heroes/102/benchmarks looks like the table header string is missing

PR #1635, opened 2018-06-23
Counts summary section on overview

Argn0:
fixes https://github.com/odota/web/issues/771


src/components/Player/Pages/Overview/CountsSummary.jsx
```diff
@@ -0,0 +1,28 @@
...
+    border-radius: 10px;
+
+    > div {
+        margin-top: 3px;
+    }
+`;
+
+const Summary = ({ data }) => (
+  <Styled>
+    {data.map(el => <GaugeChart value={el.winPercent} caption={el.category} />)}
```
howardchung: It's not very clear these are winrates, they look more like proportions.  I wonder if there's a good way to show both?

src/components/Player/Pages/Overview/Overview.jsx
```diff
@@ -199,13 +212,17 @@ Overview.propTypes = {
...
+  countsLoading: PropTypes.bool,
+  countsError: PropTypes.bool,
 };
 
 
 const getData = (props) => {
   props.getPlayerRecentMatches(props.playerId);
   props.getPlayerHeroes(props.playerId, props.location.search);
   props.getPlayerPeers(props.playerId, props.location.search);
+  props.getPlayerCounts(props.playerId, props.location.query);
```
howardchung: why does this use .query instead of .search?

src/components/Player/Pages/Overview/Overview.jsx
```diff
@@ -262,6 +279,50 @@ const countValidRecentMatches = matches => matches.filter(match => match.game_mo
...
+  const countMap = {
+    is_radiant: [],
+    game_mode: [],
+    patch: [],
+    region: [],
+  };
+
+  const limitCount = (key, field, lim) =>
+    counts[key].list.sort((a, b) => (
+      a[field] <= b[field] ? 1 : -1
```
howardchung: can you just do b[field] - a[field]?

src/components/Player/Pages/Overview/Summary.jsx
```diff
@@ -26,6 +26,7 @@ const SummOfRecMatches = ({ matchesData, strings }) => {
     tower_damage: [],
     duration: [],
     wins: [],
+    game_mode: [],
```
howardchung: I don't think this makes sense to add to the summary?

src/components/Player/Pages/Overview/Overview.jsx
```diff
@@ -262,6 +279,50 @@ const countValidRecentMatches = matches => matches.filter(match => match.game_mo
...
+    )).slice(0, lim);
+
+  Object.keys(counts).forEach((key) => {
+    switch (key) {
+      case 'is_radiant':
```
howardchung: maybe we should show the top two for each, and put each in its own column?  not sure if that would make it too tall.

src/components/Player/Pages/Overview/Overview.jsx
```diff
@@ -111,6 +113,9 @@ const Overview = ({
   toggleTurboGames,
   showTurboGames,
   strings,
+  countsData,
+  countsLoading,
+  countsError,
 }) => (
   <OverviewContainer>
     <SummaryContainer
```
howardchung: thinking we should add the rounded border to the summary component as well.  they should look similar.

src/components/Player/Pages/Overview/Overview.jsx
Argn0: I think it would be too tall.

src/components/Player/Pages/Overview/CountsSummary.jsx
Argn0: I changed the styling. Hopefully now it's more clear.

Argn0:
I didn't localize "Win" and "Loss" because we cannot guarantee that the strings fit in any language.
I also assume that almost every Dota player understands those strings in English.

Argn0:
I reverted the colors. The green/red was way too colorful/bright.

src/components/Player/Pages/Overview/Overview.jsx
howardchung: Think we could still do top two for each but leave it one row.  That would be 8 gauges (4x2)

src/components/Visualizations/GaugeChart.jsx
```diff
@@ -0,0 +1,134 @@
...
+    return 'rgb(156,148,96)';
+  } else if (value < 53) {
+    return 'rgb(140,159,99)';
+  }
+  return 'rgb(117,176,103)';
+};
+
+const GaugeChart = ({ value, caption }) => (
+  <Styled percent={computeMeterPercent(value)} meterColor={computeMeterColor(value)}>
+    <div className="caption">{caption}</div>
```
howardchung: I think we could display the number of matches as well, maybe as parens after the category, like `DIRE (1020)`

src/components/Visualizations/GaugeChart.jsx
```diff
@@ -0,0 +1,134 @@
...
+    <div className="gauge percentage">
+      <div className="meter" />
+      <div className="percentage-container">
+        <div className="percentage-indicator">
+          {`${Math.round(value * 10) / 10}%`}
+        </div>
+      </div>
+    </div>
+    <div className="win">Win</div>
```
howardchung: maybe W and L to make it more general instead of explicitly English?

src/components/Player/Pages/Overview/Overview.jsx
howardchung: could possibly also fit one more set of gauges, maybe the top two lane_role counts (skipping unknown)

Argn0:
Fixed.

It's really easy to read at a quick glance but also dense in information now.

howardchung:
I wonder if we could put in some kind of separator element between the gauges for each type of count.

Aside from that, I like this a lot!

Argn0:
-added a small visual divider

PR #1639, opened 2018-06-25
Move purchases toggle to the right and improve styling

Argn0:


Argn0:
added some unrelated fix as well

PR #1643, opened 2018-06-29
Feature/randomize crying hero

### LINKED ISSUE #1638: pls randomize which hero is crying on the 404 page
i love me some DP, but a sad <other hero> would be nice, too

tienpham94:
fixes #1638

I added 3 new crying hero images (undying and tide) that has the same dimensions with the original one (64x64px), and randomize the image will be displayed on the 404 page

New result: 
randomize-crying-hero


src/components/FourOhFour/FourOhFour.jsx
```diff
@@ -43,7 +43,12 @@ const FourOhFour = ({ strings, msg }) => (
   <Styled>
     <div className="container">
       <Helmet title={`${strings.error} 404`} />
-      <img src="/assets/images/sad.gif" alt="" />
+      <img
+        src={`/assets/images/${
+          sample(['sad', 'bawl', 'cry', 'charm-cry'])
```
Argn0: it might be possible to just use `Math.random()` instead

tienpham94: Thank you @Argn0 , I have switch from lodash method to Math.random() thingy 💯

tienpham94:
Thank you for reviewing @Argn0 , I have switched from lodash method to Math.random() 💯

tienpham94:
Sorry @Argn0 my bad, was VSCode 1 click commit.  Thank you.

src/components/FourOhFour/FourOhFour.jsx
```diff
@@ -39,11 +38,18 @@ const Styled = styled.div`
   }
 `;
 
+const imageList = ['sad', 'bawl', 'cry', 'charm-cry'];
```
howardchung: I'd label these as the names of the heroes featured to make it more readable. . . unless these are the names of the original files?

howardchung: It looks like cry and charm-cry are both undying, is there another crying gif of a different hero?

tienpham94: Yeah those are the original names of the emoticons, I found them here
https://dota2.gamepedia.com/Emoticons
But these gifs here are 32x32px, not 64x64 so I did some photoshop to make them the same as `sad`. I only found these images, there are couple of different options like WR `facepalm` but not exactly the same so I didn't put.

howardchung: In that case, I'd suggest only keeping one of the undying images so there's an equal chance of each hero.

tienpham94: Updated. I keep the more pretty version of undying and remove another. Also tested, still working ok.

PR #1648, opened 2018-07-04
Fixed heroes column

Argn0:
fixes https://github.com/odota/web/issues/1378

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -112,12 +112,12 @@ class Table extends React.Component {
     this.state = initialState;
   }
 
-  setTableRef = (node) => { 
+  setTableRef = (node) => {
     this.innerContainerRef = node;
     if (this.innerContainerRef) {
       this.innerContainerRef.refs.tableDiv.onscroll = this.handleScroll;
     }
-   }
+  }
 
   setCurrentPage = (pageNumber) => {
     this.setState({
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -112,6 +112,22 @@ class Table extends React.Component {
     this.state = initialState;
   }
 
+  setInnerContainerRef = (node) => { this.innerContainerRef = node; }
+
+  handleScroll = () => {
+    const { scrolled } = this.state;
+    const { scrollLeft } = this.innerContainerRef.childNodes[0].childNodes[0];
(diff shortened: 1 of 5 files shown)


howardchung:
I'm wondering if we should apply this to all tables (always fix the first column). since this is also a problem on match page tables.  Could we make it part of the Table component?

src/components/Table/Table.jsx
```diff
@@ -112,6 +112,18 @@ class Table extends React.Component {
     this.state = initialState;
   }
 
+  setInnerContainerRef = (node) => { this.innerContainerRef = node; }
+
+  handleScroll = () => {
+    const { scrolled } = this.state;
+    const { scrollLeft } = this.innerContainerRef.childNodes[0].childNodes[0];
```
howardchung: this seems very specific.  could it fail if the table structure changes?

Argn0:
I found an alternative method. Maybe this is a better solution?

howardchung:
Yes, I think this is better.  Do you have a version where the sponsor button disappears on scroll?  That might actually be OK.

Argn0:
setting the event listener in componentDidUpdate didnt work for the tables below. (ability builds).

So I have to set it in componendDidMount too

howardchung:
Is there a reason to do it in componentDidUpdate?  can't you just do it in mount?

Argn0:
see my discord message.:

ref callbacks only get called before componentDidMount if the element you want set the ref for actually gets rendered. so if there is a loading condition like ```{!loading && <div ref={.....}>...</div>}``` it doesnt work

Argn0:
also see https://stackoverflow.com/a/50019873

howardchung:
But if the component isn't rendering, do we need the table ref to be assigned?  shouldn't the table not be showing in that case?

Argn0:
We need the ref to bind the onscroll event listener to tableDiv. tableDiv is the container of material-ui tables.

Argn0:
Actually forget what I said. It works just fine like this...

Argn0:
I fixed the first column on the large heroes table by default on mobile. I'm worried about bad performance on low tier mobiles.

howardchung:
I think I actually prefer it not to default to fixed, it shouldn't look different unless the user interacts with the table.

Argn0:
undid commit with default scrolled

PR #1658, opened 2018-07-11
Added lane outcome to the hero tooltips!

### LINKED ISSUE #1655: show lane outcome in overview
Dotabuff shows lane outcome in the match overview page, which is the only reason I open matches both there and in Opendota.

wanderrful:
Fixed #1655 per user request. Waving the mouse over a hero's icon in the match overview's `BuildingMap` component will now have a new line at the bottom of the tooltip pop-up to say either "Win", "Loss", or "No result" based on the outcome of that lane. It is calculated the same way as in the LaneStory class contained in the MatchStory.jsx file.

It is probably better practice to take that part of the LaneStory class out into its own file so that we can refactor and stick to the DRY principle and use indirection, but I don't know what your rules are about adding new files or components, so I just played it safe and ported the functionality directly into this component via hard coding.

I hope this PR request is worthy of acceptance!

### DIFF
--- a/src/components/Match/BuildingMap/BuildingMap.jsx
+++ b/src/components/Match/BuildingMap/BuildingMap.jsx
@@ -439,7 +439,25 @@ const BuildingMap = ({ match, strings }) => {
     const roaming = (
       <span className="roaming">{strings.roaming}</span>
     );
+
+    const getLaneScore = players => (Math.max(...players.map(player => player.gold_t[10] || 0)) || 0);
+    const laneScoreDraw = 500;
+
     for (let i = 0; i < match.players.length; i += 1) {
+      const { lane } = match.players[i];
+      const radiantPlayers = match.players.filter(player => player.lane === parseInt(lane, 10) && player.isRadiant && (!player.is_roaming));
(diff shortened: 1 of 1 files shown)


howardchung:
Thanks! I think this might be easier to use if part of the overview table, as discussed in the issue.

Optimally we would reuse the utility functions (move the relevant code to the utility file) for calculating win/loss, but if you're not sure about refactoring the code, you don't have to.

wanderrful:
I can refactor it, sure. I'm pretty sure that it's only used by LaneStory in the MatchStory file so it should be relatively simple.

wanderrful:
Just an update for you -- I've refactored everything into the utility file as requested, but I'm having trouble untangling the TeamTable component. It seems like everything is very much abstracted away so that the Table's contents appear to be read directly from the API object, which makes it hard to see where exactly I'd modify the table's data itself to include the lane outcome.

howardchung:
The particular cell you want to change is rendered here: https://github.com/odota/web/blob/a6a5c7664157d50b925b2e78280b60558331a47e/src/components/Match/matchColumns.jsx#L25

`row` is the player's row in the match, you could set properties on it (e.g. row.laneResult) and render those out in TableHeroImage.

PR #1662, opened 2018-07-13
Hero tooltip

Argn0:


src/utility/index.jsx
```diff
@@ -394,6 +394,10 @@ export const sumValues = f => Object.values(f).reduce((a, b) => a + b);
 /* eslint-disable camelcase */
 // https://dota2.gamepedia.com/Attributes
 export function compileLevelOneStats(hero) {
+  if (!hero) {
+    return null;
```
howardchung: should this be {} to avoid crash if hero can't be found?

Argn0:
-return empty object instead of null

PR #1675, opened 2018-07-19
New Explorer fields

Argn0:
gold advantage and win against mega creeps


src/components/Explorer/fields.js
```diff
@@ -230,6 +230,23 @@ const getFields = (players = [], leagues = [], teams = []) => {
...
   }));
+
+  function setGoldAdvantageText(element) {
+    if (!element) {
+      return String(element);
+    }
+    if (element > 0) {
+      return `${strings.general_radiant} ${element * 1000}`;
+    }
+    return `${strings.general_dire} ${Math.abs(element * 1000)}`;
```
howardchung: Can't you use math.abs in both places?

howardchung: i.e. you don't need the if block

howardchung: i guess you still need it for the string to display so this is probably ok

src/components/Explorer/fields.js
```diff
@@ -492,7 +509,10 @@ ${props.player && props.player.value ? '' : 'AND player_matches.account_id < pla
       const laneRoleId = Number(str.substring('lane_role_'.length));
       return { text: strings[str], value: laneRoleId, key: String(laneRoleId) };
     }),
-    isTi8Team: [{ text: 'Yes', value: true, key: 'true' }],
+    isTi8Team: [{ text: '\u2714', value: true, key: 'true' }],
```
howardchung: what is \u2714?

Argn0: it's a check mark

src/components/Explorer/fields.js
```diff
@@ -230,6 +230,23 @@ const getFields = (players = [], leagues = [], teams = []) => {
...
+    if (!element) {
+      return String(element);
+    }
+    if (element > 0) {
+      return `${strings.general_radiant} ${element * 1000}`;
+    }
+    return `${strings.general_dire} ${Math.abs(element * 1000)}`;
+  }
+
+  const goldAdvantage = Array(21).fill().map((_, i) => 50 - (i * 5)).map(element => ({
```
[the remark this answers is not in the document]
Argn0: polyfill for fill() was added by gu3st

src/components/Explorer/queryTemplate.js
```diff
@@ -127,6 +130,9 @@ ${conjoin`matches.start_time >= extract(epoch from timestamp '${new Date(minDate
 ${conjoin`matches.start_time <= extract(epoch from timestamp '${new Date(maxDate.value).toISOString()}')`}
 ${conjoin`leagues.tier = '${tier}'`}
 ${isTi8Team ? 'AND teams.team_id IN (5, 15, 39, 67, 2163, 350190, 543897, 726228, 1375614, 1838315, 1883502, 2108395, 2586976, 5026801, 5027210, 5066616, 5228654, 5229127)' : ''}
+${megaWin ? 'AND ((matches.barracks_status_radiant = 0 AND matches.radiant_win) OR (matches.barracks_status_dire = 0 AND NOT matches.radiant_win))' : ''}
+${radiantMaxGoldAdvantage ? `AND (SELECT matches.radiant_gold_adv[array_upper(matches.radiant_gold_adv, 1)]) ${radiantMaxGoldAdvantage.value >= 0 ? '<=' : '>='} ${radiantMaxGoldAdvantage.value}` : ''}
```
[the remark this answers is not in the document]
Argn0: A max Dire advantage of 10000 gold would mean that radiant_gold_adv has to be bigger than -10000.

```(SELECT matches.radiant_gold_adv[array_upper(matches.radiant_gold_adv, 1)]) >= -10000```

https://deploy-preview-1675--opendota-stage.netlify.com/explorer?minDate=2018-06-24T18%3A02%3A47.420Z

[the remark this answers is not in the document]
Comment (Argn0) [reply src/components/Explorer/queryTemplate.js]: whoops just noticed the url query is not working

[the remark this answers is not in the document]
Comment (Argn0) [reply src/components/Explorer/queryTemplate.js]: nevermind...the query string is updated when you do a query https://deploy-preview-1675--opendota-stage.netlify.com/explorer?minDate=2018-06-24T18%3A02%3A47.420Z&radiantMaxGoldAdvantage=-10000.

But the "Dire" string until you click on the field.

File: src/components/Explorer/fields.js [hunk 5]
```diff
@@ -230,6 +230,23 @@ const getFields = (players = [], leagues = [], teams = []) => {
     value: element,
     key: String(element),
   }));
+
+  function setGoldAdvantageText(element) {
+    if (!element) {
+      return String(element);
+    }
+    if (element > 0) {
+      return `${strings.general_radiant || 'Radiant'} ${element * 1000}`;
```
howardchung: What is the default string for?

src/components/Explorer/fields.js
howardchung: I would recommend text since it's more understandable

src/components/Explorer/queryTemplate.js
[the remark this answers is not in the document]
howardchung: Maybe the field/text should just always be "radiant gold advantage" and use negative/positive values in the field.  That might be less confusing.

src/lang/en-US.json
```diff
@@ -211,6 +211,10 @@
   "explorer_having": "At Least This Many Matches",
   "explorer_limit": "Limit",
   "explorer_match": "Match",
+  "explorer_is_ti_team": "Is TI{number} team",
```
howardchung: capitalize each word for consistency with the other fields

src/lang/en-US.json
```diff
@@ -211,6 +211,10 @@
...
+  "explorer_mega_comeback": "Win against mega creeps",
```
howardchung: "Won" I think is better since it's always past

src/components/Explorer/queryTemplate.js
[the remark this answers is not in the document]
howardchung: actually now that I think about it, it seems to make more sense for this field to be "max gold advantage" regardless of the team.  The dropdown should not have Radiant and Dire, it should just be negative and positive values.  If a user is looking for e.g. wins where a team was down by 10000 gold or more, they probably don't care whether the team was Radiant or Dire.

src/components/Explorer/fields.js
```diff
@@ -230,6 +230,23 @@ const getFields = (players = [], leagues = [], teams = []) => {
...
+    }
+    return `${strings.general_dire || 'Dire'} ${Math.abs(element * 1000)}`;
+  }
+
+  const goldAdvantage = Array(21).fill().map((_, i) => 50 - (i * 5)).map(element => ({
```
howardchung: this range seems a bit large. . . maybe -20000 to 20000 in increments of 1000?

Argn0:
-replaced check marks with text
-capitalized each word
-changed how Max and Min advantage fields work
-range is now -20000 to 20000 in increments of 1000

src/components/Explorer/queryTemplate.js
```diff
@@ -127,6 +130,9 @@ ${conjoin`matches.start_time >= extract(epoch from timestamp '${new Date(minDate
 ${conjoin`matches.start_time <= extract(epoch from timestamp '${new Date(maxDate.value).toISOString()}')`}
 ${conjoin`leagues.tier = '${tier}'`}
 ${isTi8Team ? 'AND teams.team_id IN (5, 15, 39, 67, 2163, 350190, 543897, 726228, 1375614, 1838315, 1883502, 2108395, 2586976, 5026801, 5027210, 5066616, 5228654, 5229127)' : ''}
+${megaWin ? 'AND ((matches.barracks_status_radiant = 0 AND matches.radiant_win) OR (matches.barracks_status_dire = 0 AND NOT matches.radiant_win))' : ''}
+${radiantMaxGoldAdvantage ? `AND (SELECT matches.radiant_gold_adv[array_upper(matches.radiant_gold_adv, 1)]) <= ${radiantMaxGoldAdvantage.value}` : ''}
```
howardchung: why is there a SELECT here, shouldn't it just be matches.radiant_gold_adv?

src/lang/en-US.json
```diff
@@ -211,6 +211,10 @@
   "explorer_having": "At Least This Many Matches",
   "explorer_limit": "Limit",
   "explorer_match": "Match",
+  "explorer_is_ti_team": "Is TI{number} Team",
+  "explorer_mega_comeback": "Won Against Mega Creeps",
+  "explorer_max_gold_adv": "Max Gold Advantage",
```
howardchung: Does this string need to indicate radiant gold advantage?

howardchung: never mind this should be either team

src/components/Explorer/queryTemplate.js
```diff
@@ -127,6 +130,9 @@ ${conjoin`matches.start_time >= extract(epoch from timestamp '${new Date(minDate
...
+${radiantMinGoldAdvantage ? `AND (SELECT matches.radiant_gold_adv[array_upper(matches.radiant_gold_adv, 1)]) >= ${radiantMinGoldAdvantage.value}` : ''}
```
howardchung: do you need to do absolute value here so you pick up matches where radiant or dire were leading?

Argn0: This is getting too confusing. Do you want me to do Max Radiant Advantage / Min Radiant  Advantage and Max Dire Advantage / Min Dire Advantage?

So 4 new fields in total.

Argn0: I think this would also be the least confusing to the user.

Argn0: or we could use a range slider? https://codepen.io/davidchin/full/GpNvqw

howardchung: I think just max gold advantage is most useful to the user, but that means the query needs to use absolute values so it can pick up on both negative (dire leading) and positive (radiant leading) values.

My concern with splitting it into 4 fields is that the user can't combine them, e.g. if they want to look at games where a team led by 10K gold or more, they would need to make one query for the radiant and one for the dire.  If the query handles absolute values, then it can just look for matches where the last element of radiant_gold_adv was abs greater than 10K

Argn0:
Using absolute values now. I was a bit confused because initially I had a completely different implementation in mind.

src/components/Explorer/fields.js
```diff
@@ -493,6 +500,9 @@ ${props.player && props.player.value ? '' : 'AND player_matches.account_id < pla
       return { text: strings[str], value: laneRoleId, key: String(laneRoleId) };
     }),
     isTi8Team: [{ text: 'Yes', value: true, key: 'true' }],
+    megaWin: [{ text: 'Yes', value: true, key: 'true' }],
+    radiantMinGoldAdvantage: goldAdvantage,
```
howardchung: should probably rename this field to minGoldAdvantage to avoid confusion

Argn0: done

PR #1678, opened 2018-07-24
Remove webpack references

dartharaknus:
And also deleted unneeded scripts folder

**Squash the PR when merging**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ notifications:
   - aqc2109@columbia.edu
   - mjr127@case.edu
 script:
+- npm run lintnofix
 - npm run build
 # blocklist
 branches:
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
   "license": "MIT",
   "scripts": {
     "start": "react-scripts start",
-    "install-start": "npm install && npm run lintnofix && npm run start",
-    "build": "npm install && npm run lintnofix && npm run rsbuild",
-    "rsbuild": "react-scripts build",
+    "start:install": "npm install && npm run lintnofix && npm run start",
+    "build": "react-scripts build",
+    "build:install": "npm install && npm run lintnofix && npm run build",
     "snapshot": "react-snap",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -5,18 +5,17 @@
   "license": "MIT",
   "scripts": {
     "start": "react-scripts start",
+    "install-start": "npm install && npm run lintnofix && npm run start",
     "build": "npm install && npm run lintnofix && npm run rsbuild",
     "rsbuild": "react-scripts build",
```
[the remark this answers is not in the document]
dartharaknus: I would propose `start`, `start:install`, `build`, `build:install`. But I understand if people want to leave them as it is now.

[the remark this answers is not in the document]
gu3st: I like the `:` separator for indicating command heirarchy/variants.  as long as there's only 1 level.  I don't want to see no `start:dev:howard` 😛

package.json
```diff
@@ -5,18 +5,17 @@
...
     "snapshot": "react-snap",
     "test": "react-scripts test --env=jsdom",
-    "eject": "react-scripts eject",
```
howardchung: I would keep this since it's a standard part of CRA.  I have no plans to eject, but we should leave it there

dartharaknus: I don't want people to accidentally run it. It's not reversible. We should keep it hidden (since it will always be there as part of react-scripts)

.travis.yml
```diff
@@ -7,6 +7,7 @@ notifications:
   - aqc2109@columbia.edu
   - mjr127@case.edu
 script:
+- npm run lintnofix
 - npm run build
```
dartharaknus: We shouldn't do `npm build:install` here, because `npm install` takes a lot more time while CI already does `npm ci` to install stuff.

howardchung:
I still recommend keeping eject in the scripts, otherwise this LGTM

dartharaknus:
@howardchung Updated. Squash the PR while merging into one single commit please. Thanks.

PR #1699, opened 2018-07-25
Optimize travis config

dartharaknus:


.travis.yml
```diff
@@ -1,6 +1,7 @@
+sudo: false
```
howardchung: does this make the VM start up faster?

.travis.yml
```diff
@@ -1,6 +1,7 @@
...
 language: node_js
 node_js:
-- 'node'
+- '10'
```
howardchung: what's the purpose of this?  shouldn't we always run on the latest?

dartharaknus: I couldn't find which node version everyone was running this on. I think it's better if we upgrade it manually rather than letting it be automatic. So, if 12 comes out, then we can replace this with 12 and then say that the app runs on 12.

.travis.yml
dartharaknus: No, but waiting times are less in travis.

PR #1716, opened 2018-07-31
Reduce extrema on wardmap

Argn0:
fixes https://github.com/odota/web/issues/1714

https://www.opendota.com/players/123596051/wardmap

### DIFF
--- a/src/components/Heatmap/Heatmap.jsx
+++ b/src/components/Heatmap/Heatmap.jsx
@@ -11,15 +11,15 @@ import DotaMap from '../DotaMap';
  * Returns the adjusted heatmap data.
  */
 function scaleAndExtrema(points, scalef, max, shift) {
-  const averageValue = (points.reduce((a, b) => a + b.value, 0) / points.length);
-  const maxDeviationFromAvg = 25;
+  // the max values should not deviate from the average by more than a factor of 25
+  const maxValue = (points.reduce((a, b) => a + b.value, 0) / points.length) * 25;
 
   const newPoints = points.map(p => ({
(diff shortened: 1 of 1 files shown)


src/components/Heatmap/Heatmap.jsx
```diff
@@ -11,13 +11,16 @@ import DotaMap from '../DotaMap';
  * Returns the adjusted heatmap data.
  */
 function scaleAndExtrema(points, scalef, max, shift) {
+  const averageValue = (points.reduce((a, b) => a + b.value, 0) / points.length);
```
howardchung: why are we using b.value but not a.value?

src/components/Heatmap/Heatmap.jsx
```diff
@@ -11,13 +11,16 @@ import DotaMap from '../DotaMap';
...
+  const maxDeviationFromAvg = 25;
+
   const newPoints = points.map(p => ({
     x: Math.floor(p.x * scalef),
     y: Math.floor(p.y * scalef),
-    value: p.value + (shift || 0),
+    value: Math.min(p.value, (averageValue * maxDeviationFromAvg)) + (shift || 0),
```
howardchung: I would just compute averageValue * maxDeviationFromAvg once and reuse that variable (maybe call it maxValue)

src/components/Heatmap/Heatmap.jsx
howardchung: ah you're right, I'm used to using reduce on arrays of numbers but this is an array of objects

PR #1729, opened 2018-08-03
Scroll bug

Argn0:
fixes https://github.com/odota/web/issues/1725


howardchung:
could we fix the column but not shrink it?

howardchung:
that might be more complicated though

Argn0:
Yes, I'll try.

Argn0:
Done.

PR #1754, opened 2018-08-16
Item tooltip

Argn0:
will display incorrectly until dotacontants is updated.


src/components/ItemTooltip/index.jsx
```diff
@@ -0,0 +1,261 @@
...
+import propTypes from 'prop-types';
+import styled from 'styled-components';
+import items from 'dotaconstants/build/items.json';
+import constants from '../constants';
+
+const itemAbilities = {
+  active: {
+    color: '#4183d7',
+    headerColor: 'rgba(65, 131, 215, 0.25)',
+    text: 'Active',
```
howardchung: localize?

Argn0: but the rest of the tooltip is in english too. I thought it would be pointless

howardchung: yeah that's true.  up to you if you want to do this, it might make it easier if we one day localize the tooltip fully

howardchung:
you can upgrade dotaconstants to 6.0.0

Argn0:
OK, updated dotaconstants, but something is broken. match pages are loading endlessly.

trying to figure out why. the item tooltips are working though if you check them on the scenarios page.

PR #1775, opened 2018-08-27
Grimstroke

Argn0:


src/utility/index.jsx
```diff
@@ -180,6 +180,9 @@ const getTitle = (row, col, heroName) => {
 };
 
 export const getHeroImageUrl = (heroId, imageSizeSuffix) => {
+  if (heroId === 121) {
+    return '/assets/images/grimstroke_sb.png';
```
howardchung: is this necessary? doesn't the dotaconstants update include this path?

Argn0: returns 404 right now. this path points to the image I uploaded manually.

Argn0: oh sorry. i mean, the other image is the big version. it's used for the heroes page for example.

PR #1792, opened 2018-09-04
Start extracting common code into a separate repo

dartharaknus:
I have started trying to extract the common code out of this repo. This is the first of many PRs.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/common"]
```
howardchung: are there alternatives to git submodules?

src/store/index.js
```diff
@@ -1,33 +0,0 @@
-import {
```
howardchung: is there a reason we're moving store?  I thought we only wanted to reuse certain components, like the Table

.gitmodules
dartharaknus: Having it as a normal dependency means we need to do all the webpack configuration there too and then build the distribution files and then publish to npm. Which I thought is a bad thing because of maintaining 2 different configuration files. This is definitely better.

src/store/index.js
dartharaknus: Yeah, because the other web app also uses the same store code. I am basically moving all the common stuff, not only components.

howardchung: The store structures probably won't match though, since there are definitely dota-specific things in there like abilities and heroes?

.gitmodules
howardchung: OK, as long as the CI/build handles these, I'm ok with this :)

src/store/index.js
dartharaknus: Yup, store structures won't match, which is why there still is `import app from '../reducers';`

howardchung: Could we keep separate store files? I think it's possible that we might want to make changes to one or the other.

howardchung: I think we should try to share only components and not framework code

dartharaknus: We can always put the separate store file when we need it. Until then we can use it as a common part. When you want to share code between 2 websites, it doesn't have to be components only.

howardchung:
I'd like to hold this until after the next web release (probably this weekend).

I'd still suggest keeping separate store files in case we want to change something/add a middleware to only one or the other store (maybe something experimental that we only want to test on OpenArtifact).  If we are planning to switch to TypeScript at some point for OpenArtifact that might also make it less likely to cause an issue with OpenDota.

howardchung:
Also, if we are using submodules then we need to update the README with the instructions if it won't work out of the box:
https://stackoverflow.com/questions/3796927/how-to-git-clone-including-submodules

dartharaknus:
I don't want to do typescript in openartifact. As I said, let's keep most of the code for both the websites same.

howardchung:
My only concern with this is that it will require all developers to update their repos (running the submodule download command).

We should update the README to reflect this.

Alternatively, would it be possible to keep the common code in this repo and import this repo into the artifact web UI to reuse relevant files?

howardchung:
Can you show an example of an import that fails if we just submodule odota/web into openartifact/web?

dartharaknus:
Small example is this line in https://github.com/odota/web/blob/master/src/store/index.js#L12 would always refer to the opendota's reducers unless we take the `store/index.js` into a separate submodule. There are other places this is happening too.

Or let's forget about this PR and I will start OpenArtifact in vue.

howardchung:
We could probably create a separate store for openartifact, I don't know if there's significant value in reusing that. The code that I primarily wanted to reuse are the components and utility functions, which I think should be importable from a submodule of the project.

I'd say it's fine to bootstrap the project possibly with some duplication, then work on sharing common code later as we need it.

dartharaknus:
> We could probably create a separate store for openartifact, I don't know if there's significant value in reusing that

There is a significant value because it is part of the architecture of the project and sharing architecture is why things like CRA exist in the first place.

> The code that I primarily wanted to reuse are the components and utility functions

The store is an utility, is it not?

This has been delayed long enough. I have removed the store for now. Please merge it ASAP

howardchung:
Removing the store means we can remove submodules from this PR? You should be able to import components by adding submodule to openartifact/web

dartharaknus:
No, components have the same issues as describe above. The store was just an example since I wanted to keep the description simple.

howardchung:
If a component has this issue it should be updated so that it can be imported.  Looks like this is just service worker at the moment so it shouldn't break anyone's workflow to merge

Issue #1596, opened 2018-05-23
add smoke tests for render crashes


howardchung:
We should figure out a way to have CI run the react app and hit all the routes to catch any runtime JS exceptions we might have.

howardchung:
we should really prioritize this

howardchung:
we can also set up property bags/pre-defined API results so we're not hitting the live API all the time.

howardchung:
testcafe is a tool we use at Microsoft for testing React applications end-to-end:
http://devexpress.github.io/testcafe/

howardchung:
fixed in #1804

Issue #144, opened 2016-09-20
Unit tests


mike-robertson:
We need to add unit tests since the code base is becoming quite large, and since we all have our own 'domains' that we work on, changes to components across domains can cause subtle bugs (if we don't have unit tests). I think that the #1 priority after completing MVP should be to go back and add useful unit tests for everything since we don't have any dedicated QA contributors or QA process.

Roshanjossey:
@mike-robertson: I'd love to work on this. I'm very much aligned to TDD and unit testing. Do you have a testing framework in mind? I've used mocha, chai and enzyme to test components. I also had a really bad time trying to use Jest.

howardchung:
odota/core uses mocha.  I think it'd be helpful to use the same one on both.

mike-robertson:
I agree that mocha/chai/enzyme is the easiest testing 'stack' to use with react.

dartharaknus:
We also need e2e testing, http://nightwatchjs.org/

howardchung:
https://www.browserstack.com/ is another option.  I've never tried setting this kind of thing up, so feel free to tackle this if you have any experience.

Roshanjossey:
@howardchung: As far as I understand, browserstack is a paid service, do we have enough bandwidth to use it?
@dartharaknus: nightwatchjs looks pretty cool. Have you used it? Looking forward to work with it

howardchung:
They are free for open source projects (need to submit a request):
https://www.browserstack.com/pricing

howardchung:
@Roshanjossey are you still interested in improving test coverage?  @mike-robertson laid out the foundation and wrote a couple of quick tests, but the majority of components still don't have them.

masad-frost:
https://facebook.github.io/jest/ is worth considering

Roshanjossey:
@howardchung, @mike-robertson, I'm sorry I couldn't take a look at this earlier. I'll get started on this.

How should I divide work? I'd like to submit PRs in smaller chunks rather than doing it all at once

howardchung:
Maybe just add one test for one component at a time.

masad-frost:
I wrote on discord, but might be better to say it here. We should probably be using selenium as our web driver now that browserstack is looking to support us https://www.browserstack.com/automate

Tb0:
Hello everyone,
[...]

PR #1799, opened 2018-09-09
Testcafe Example

Phrynobatrachus:
This is a quick example for #1596 using Testcafe. The second test in the file is probably closer to what we're looking for, if the goal is just to run through available routes. I suppose the third is more #144? Anyways I figured I'd post this PR for anyone who might be interested in working on either of those issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -421,6 +421,18 @@
         }
       }
     },
+    "@types/error-stack-parser": {
+      "version": "1.3.18",
+      "resolved": "https://registry.npmjs.org/@types/error-stack-parser/-/error-stack-parser-1.3.18.tgz",
+      "integrity": "sha1-4ByfjIXKg7YQMgxiJYsMkCat4Pc=",
+      "dev": true
+    },
+    "@types/lodash": {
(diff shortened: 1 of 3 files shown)


howardchung:
Thanks. I believe @Argn0 also has a branch with testcafe included, perhaps we can integrate that with this tests?

Luchanso:
Can you look on [puppeteer](https://github.com/GoogleChrome/puppeteer) too?

Argn0:
Does anyone know why you can't do named imports in the test1.js file?

like if I wanted to do ```import { isRadiant } from '../src/utility';``` it doesnt work.

Argn0:
and if you do ```import transformMatch from '../src/actions/transformMatch';``` it imports transformMatch but then it can't import utility in transformMatch.

Phrynobatrachus:
@Luchanso I probably will try it out, but Testcafe looks like it might be easier to integrate with BrowserStack, and it supports running tests through FF as well.

howardchung:
I think we can combine this with https://github.com/odota/web/pull/1804 which is similar

PR #1804, opened 2018-09-12
Testcafe

Argn0:


howardchung:
LGTM, just sync latest to fix lockfile

.travis.yml
```diff
@@ -1,7 +1,22 @@
-sudo: false
+dist: trusty
+sudo: required
 language: node_js
 node_js:
 - '10'
+addons:
+  firefox: latest
```
howardchung: are we testing on firefox?

.travis.yml
```diff
@@ -1,7 +1,22 @@
...
+     - google-chrome
+    packages:
+     - google-chrome-stable fluxbox
+before_script:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3
+  - fluxbox >/dev/null 2>&1 &
+before_install:
+  - stty cols 80
```
howardchung: curious what this is for

.travis.yml
Argn0: no we only test on chrome. I think I should remove that

.travis.yml
Argn0: I'm not sure. I think it has something to do with unix shell settings. it's from the testcafe documentations

testcafe/tests/players.js
```diff
@@ -0,0 +1,40 @@
...
+} from 'testcafe';
+import {
+  fixtureBeforeHook,
+  fixtureBeforeEachHook,
+  fixtureAfterHook,
+  fixtureRequestHooks,
+  host,
+} from './../testsUtility';
+
+const paths = [
```
howardchung: optional: I like putting all the test paths in one file

.travis.yml
Argn0: actually...I dont think we can remove it. the chrome stuff is part of it i think?

Argn0:
fixed the merge conflict

PR #1810, opened 2018-09-14
small bug fixes

morgvanny:
some extra arguments were being passed in and not used, and some conditionals weren't useful. I'm a bit confused about what's going on in src/components/Match/Vision/index.jsx since there are a bunch of functions defined in the Vision class that aren't actually called as far as I can tell. seems like a work in progress, so I didn't touch much there.

### DIFF
--- a/src/components/Heading/Heading.jsx
+++ b/src/components/Heading/Heading.jsx
@@ -29,7 +29,7 @@ const Heading = ({
         <span className="sponsor-button">
           <RaisedButton
             label={buttonLabel}
-            icon={buttonIcon ? <img src={buttonIcon} alt="" /> : ''}
+            icon={<img src={buttonIcon} alt="" />}
             href={buttonTo}
             target="_blank"
             rel="noopener noreferrer"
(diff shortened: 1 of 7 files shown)


src/components/Heading/Heading.jsx
```diff
@@ -29,7 +29,7 @@ const Heading = ({
         <span className="sponsor-button">
           <RaisedButton
             label={buttonLabel}
-            icon={buttonIcon ? <img src={buttonIcon} alt="" /> : ''}
```
morgvanny: already inside a conditional that checked this

src/components/Heatmap/Heatmap.jsx
```diff
@@ -17,7 +17,7 @@ function scaleAndExtrema(points, scalef, max, shift) {
   const newPoints = points.map(p => ({
     x: Math.floor(p.x * scalef),
     y: Math.floor(p.y * scalef),
-    value: Math.min(p.value, maxValue) + (shift || 0),
```
morgvanny: shift appears to be hardcoded as 25

src/components/Match/Vision/index.jsx
```diff
@@ -205,7 +205,7 @@ class Vision extends React.Component {
 
   visibleData() {
     const self = this;
-    const filter = ward => alive(ward, self.state.currentTick) && isTeam(ward, self.state.teams) && self.state.players[ward.type][ward.player];
+    const filter = ward => alive(ward, self.state.currentTick) /* && isTeam(ward, self.state.teams) */ && self.state.players[ward.type][ward.player];
```
howardchung: i'd just remove this instead of commenting out

PR #1811, opened 2018-09-16
Allow out-of-order translation parameters

### LINKED ISSUE #627: allow positional parameters in template strings
https://github.com/odota/ui/commit/c0acd3d002a43e1e716a8c923541876fdf3cd540

Thank you for try to support formatting text.  
But, `utils.format` is not support choice order.  
(like `printf("%2$d", 1, 2) // =>2`)  
Need to use some format library.

gu3st:
Fixes #627 and allows translators to change the order of any parameters in a templated string.  Also converted all existing translations to utilize new template string format (`{1}` instead of `%s`).  Included script used for transforming translations from 1 format to other format.


gu3st:
Apparently I need to run lint on my conversion script 🙃

howardchung:
Are the diffs to package-lock and shrinkwrap expected?

src/components/Match/MatchStory.jsx
```diff
@@ -142,7 +141,7 @@ const formatApproximateTime = (timeSeconds) => {
   // If the time is at least two hours, describe it in hours
   if (timeMinutes > 120) {
     const timeHours = parseInt(timeSeconds / (60 * 60), 10);
-    return `${strings.advb_over} ${util.format(strings.time_hh, timeHours)}`;
+    return `${strings.advb_over} ${formatTemplate(strings.time_hh, timeHours)}`;
```
howardchung: doesn't this need to be tostring?

gu3st:
I think I committed the shrinkwrap by accident (web doesn't currently have one in master).  Not sure why lock would have changed as I only ran npm install 🤔.  I'll go kill the shrinkwrap and revert lock file.

src/lang/bg-BG.json
```diff
@@ -86,7 +86,7 @@
   "app_pvgna": "Find a Guide",
   "app_pvgna_alt": "Find a Dota 2 Guide on Pvgna",
   "app_rivalry": "Bet on Pro Matches",
-  "app_rivalry_team": "Bet on %s Matches",
+  "app_rivalry_team": "Bet on {1} Matches",
```
howardchung: shouldn't we start at 0 for consistency with c#?

src/lang/en-US.json
```diff
@@ -1,7 +1,6 @@
 {
   "yes": "yes",
   "no": "no",
-
```
howardchung: i'm not sure if the script can preserve whitespace but it would be nice to keep the separation lines

src/utility/index.jsx
```diff
@@ -580,7 +589,7 @@ export function fromNow(time) {
 
     if (diff < unit.limit || !unit.limit) {
       const val = Math.floor(diff / unit.in_seconds);
-      return util.format(strings.time_past, val > 1 ? util.format(unit.plural, val) : unit.name);
+      return formatTemplate(strings.time_past, val > 1 ? formatTemplate(unit.plural, val) : unit.name);
```
howardchung: doesn't this need to be tostring?

src/lang/en-US.json
gu3st: If I read the file line by line then it could.  I'm just reading the file in as JSON, doing the replace, and then writing it out (using JSON.stringify(obj, undefined, "  ") to preserve some whitespace.

src/lang/bg-BG.json
gu3st: I can fix that.  Some reason I got the impression C# started at 1 🙃

src/lang/en-US.json
howardchung: I think you could just do a replace on the raw text, I don't think %s and %d ever show up in the keys or delimiters.

src/lang/bg-BG.json
howardchung: here are the docs for reference:
https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=netframework-4.7.2

PR #1814, opened 2018-09-20
Add tooltips to Records page

seanfagan:
This partially addresses issue #1342, specifically [this comment](https://github.com/odota/web/issues/1342#issuecomment-386817287).

I enable the TabBar component to show tooltips, if a tooltip is provided. I then add tooltips to the tabs on the "Records" page (i.e. "K", "D", "A", "GPM", "XPM", ...).

I'm a newbie to React -- let me know if anything looks fishy. Cheers.

### DIFF
--- a/src/components/TabBar/TabBar.jsx
+++ b/src/components/TabBar/TabBar.jsx
@@ -63,7 +63,7 @@ const TabBar = ({ tabs, info, match }) => (
           disabled={tab.disabled}
           hidden={tab.hidden && tab.hidden(match)}
         >
-          <div data-tip={tab.tooltip && true} data-for={`tooltip_${tab.key}`}>
+          <div data-tip={tab.tooltip} data-for={`tooltip_${tab.key}`}>
             {tab.name}
             {tab.tooltip &&
             <ReactTooltip id={`tooltip_${tab.key}`} place="top" effect="solid">
--- a/src/components/TabBar/TabBar.jsx
+++ b/src/components/TabBar/TabBar.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Link } from 'react-router-dom';
+import ReactTooltip from 'react-tooltip';
 import constants from '../constants';
 
 const StyledMain = styled.main`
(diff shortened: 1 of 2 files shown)


src/components/TabBar/TabBar.jsx
```diff
@@ -62,7 +63,14 @@ const TabBar = ({ tabs, info, match }) => (
           disabled={tab.disabled}
           hidden={tab.hidden && tab.hidden(match)}
         >
-          {tab.name}
+          <div data-tip={tab.tooltip && true} data-for={`tooltip_${tab.key}`}>
```
howardchung: what's the purpose of the && true?

seanfagan: @howardchung I saw something similar for the tooltips in TableHeaderColumn.jsx . I think the idea is to have a simple value in the attribute that you can use for CSS selecting, like: &[data-tip="true"] {...}

Argn0: @howardchung  ReactTooltip finds the tooltips with the data-tip attribute and it cannot be undefined.

Argn0: actually nevermind. it has to be ```data-tip={tab.tooltip}``` or else the mouse cursor will change on tabs where there is no tooltip. see the heroes page for example

seanfagan: @Argn0 is there value in me keeping the value as `{tab.tooltip && true}` so that rules like `[data-tip=true]` can still work? Or would the simpler code `data-tip={tab.tooltip}` be preferred.

Argn0: I'd just use ```data-tip={tab.tooltip}``` because the [data-tip=true] rule was removed.

seanfagan: I've removed the `&& true`.

seanfagan:
Changes made, lmk.

howardchung:
LGTM

I was actually thinking about removing the records page since it gets very little use and the nav bar is getting cluttered.

RDJ14:
What is the best way to add a new string to the .json files? There doesn't seem to be one for tooltips relating to Roshan kills for the fantasy page in matches. Is it alright to just use the strings.farm_roshan as a tooltip instead?

PR #1819, opened 2018-09-24
Contributor badge

### LINKED ISSUE #1808: add contributor badge to player profiles
suggested by @gu3st

gu3st:
fixes #1808 

This is an implementation of the contributor badge.  I used the OpenDota logo SVG.   Attached is a preview of the badge, but I hope that people have some design suggestions to make it better.  I had a greyscale version as well, but it looked too subtle. 

There's a tooltip that says `This user has contributed to the development of the OpenDota project`

It looks somewhat similar to the Cheese icon (my acct on local dev doesn't have Cheese).

screen shot 2018-09-23 at 4 27 29 pm


Argn0:
Looks nice. But I think the O should not be in pure black (0, 0, 0) but something like (17, 17, 17).

Argn0:
or maybe rgb(10, 10, 10)

Argn0:
also how about something like this? 
gu3st

gu3st:
So I implemented what @Argn0 (with their help) suggested but it has a corresponding API piece for implementation.

src/components/Match/matchColumns.jsx
```diff
@@ -39,6 +39,7 @@ export default (strings) => {
     return (<TableHeroImage
       title={row.name || row.personaname || strings.general_anonymous}
       registered={row.last_login}
+      contributor={row.isContributor}
```
howardchung: update to snake_case

howardchung: eh, let's just keep it consistent with the other is_contributor.  I think that's better to be consistent rather than with other properties.

src/components/Player/Header/PlayerBadges.jsx
```diff
@@ -174,6 +198,15 @@ export const PlayerBadgesIcons = ({
               <IconCheese className="cheese icon" />
             </div>
           )}
+          {isContributor && (
+          <div
+            className="iconButton iconContributor"
+            data-hint={`${strings.app_contributor}`}
+            data-hint-position="top"
+          >
+            <IconContributor className="icon" dColour="#21be93" oColour="#212121" />
```
howardchung: I think we usually use the spelling "color" in the existing code

src/lang/en-US.json
```diff
@@ -946,6 +947,7 @@
   "tooltip_damage_dealt": "Damage dealt to heroes by items/abilities",
   "tooltip_damage_received": "Damage received from heroes by items/abilities",
   "tooltip_registered_user": "Registered user",
+  "tooltip_contributor": "Contributor to the OpenDota project",
```
howardchung: I think we can just reuse the other string

PR #1833, opened 2018-09-27
Fix issue with extremely long cached AJAX filenames

Argn0:
Cap the super long paths at 30 characters and attach a hash number of the full path at the end.
This allows to uniquely identify each path and keep it human readable.


howardchung:
remove changes to package-lock?

package.json
```diff
@@ -74,6 +74,7 @@
     "eslint-plugin-react": "^7.7.0",
     "sanitize-filename": "^1.6.1",
     "simple-vdf": "^1.1.1",
+    "string-hash": "^1.1.3",
```
howardchung: I think you can do this with built in nodejs libraries

howardchung: https://nodejs.org/api/crypto.html#crypto_class_hash

PR #1840, opened 2018-09-28
Update testcafe

Argn0:
testcafe runs for 2m 48s on travis with these changes


package.json
```diff
@@ -9,8 +9,8 @@
     "build": "react-scripts build",
     "build:install": "npm install && npm run lintnofix && npm run build",
     "snapshot": "react-snap",
-    "test": "react-scripts test --env=jsdom",
-    "testcafe": "testcafe -c 1 chrome testcafe/ --app \"react-scripts start\" --app-init-delay 30000",
+    "test": "testcafe -c 4 chrome:headless testcafe/ --app \"serve -s build\"",
```
howardchung: headless doesn't work on travis?

src/actions/action.js
```diff
@@ -17,14 +17,17 @@ export default function action(type, host, path, params = {}, transform) {
...
+          err.fetchError = true;
           dispatch(getError(response.status));
           if (response.status >= 400 && response.status < 500) {
-            const err = new Error('client error');
-            err.clientError = true;
+            err.message = 'fetch failed - client error';
             throw err;
           } else {
-            throw new Error('fetch failed');
+            setTimeout(() => fetchDataWithRetry(delay + 3000), delay);
```
howardchung: if we're retrying why are we also throwing the error?

package.json
Argn0: headless mode had slightly worse or same performance on travis when I tested it.

src/actions/action.js
Argn0: I throw here so the promise chain jumps to the catch handler.

package.json
howardchung: if it's approx the same I'd rather keep one version of the test to avoid fragmentation

src/actions/action.js
howardchung: I think we should only throw if we exceed some retry threshold.

Argn0: But if I don't throw here, then the next then()  ``` .then(transform || (json => json))```  will throw anyway

howardchung: hmm, ok.  In that case I think we should move the retry back to the catch block.

PR #1848, opened 2018-09-30
Option to collapse UI elements

Argn0:
https://deploy-preview-1848--opendota-stage.netlify.com/players/85937380

react-transition-group is a material-ui dependency. so it might not have an impact on bundle size at all. note sure.

it allows us to leave hidden components unmounted on initial page load and still have an animation when you show the component the first time.


howardchung:
I think we need to show a bit more when the section is collapsed since it's kind of hard to see otherwise.

Also, any reason the players played with section is collapsable but not the heroes?

src/lang/en-US.json
```diff
@@ -1230,5 +1230,7 @@
   "gosu_vision": "Get how many heroes were killed under your wards",
   "gosu_actions": "Get your lost time from mouse usage vs hotkeys",
   "gosu_teamfights": "Get who to target during teamfights",
-  "gosu_analysis": "Get your real MMR bracket"
+  "gosu_analysis": "Get your real MMR bracket",
+  "show": "Show",
```
howardchung: I suggest namespacing these under "general"

PR #1851, opened 2018-10-01
update Create react app scripts to v2

### LINKED ISSUE #1753: Use CRA@next
We need to move to CRA@next which has webpack4 and babel7. Fixes a few issues and helps us moving forward development wise.

farmasek:
Hi, so CRA is at its latest beta, ready to testing. 

fixes #1753


src/components/Match/Vision/index.jsx
```diff
@@ -82,7 +82,7 @@ const Styled = styled.div`
   }
 }
 `;
-
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
howardchung: we should just make this a warning in eslint config

howardchung:
2.0.4 is the latest bugfix

src/components/Match/Vision/index.jsx
farmasek: normally it is a warning, but in CI build it fails beacause warnings are treated as errors. Interesting is, that it only fails at this warning, there are several more in the project which build ignores.

dartharaknus:
What about the polyfills @gu3st ?

gu3st:
I'll test it later and see if those are redundant.

PR #1852, opened 2018-10-01
add old match to testcafe

Argn0:
this should probably only get merged after the crash is fixed

### DIFF
--- a/testcafe/tests/matches.js
+++ b/testcafe/tests/matches.js
@@ -9,29 +9,29 @@ import {
   host,
 } from './../testsUtility';
 
-
-const paths = [
-  '/matches/4080856812/overview',
-  '/matches/4080856812/benchmarks',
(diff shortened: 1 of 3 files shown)


testcafe/tests/matches.js
```diff
@@ -40,48 +46,39 @@ paths.forEach((p) => {
...
-  const checkboxes = Selector('.innerContainer').find('input');
-  const checkboxCount = await checkboxes.count;
-  for (let i = 0; i < checkboxCount; i++) {
-    await t
-      .click(checkboxes.nth(i))
-      .wait(100);
-  }
-});
+const legacy = [
+  '/matches/2472899185/overview',
```
howardchung: we should probably define the suffixes in a list once and reuse them

Argn0:
I included the patch for the crash in this PR.

Issue #1843, opened 2018-09-29
Remove MMR tab from player profiles


mobyte0:
The MMR tab should be removed since MMR isn't really a stat that is used anymore after being replaced by medals. It could stay in the files for archival purposes if desired but I don't see the point in having it on display in a profile when it doesn't serve any functionality anymore.

howardchung:
I think it's still useful for looking up old data/graphs

mobyte0:
Like I said, it could stay in as code for archival but I think when looking at the profile page, it doesn't make sense to have it as a link since the last updated data is from almost a year ago.

xynazog:
I think people still refer to MMR on the Profile/Stats page on their DoTA accounts. It is used to calculate approximately the number of matches to win to reach the next level.

mobyte0:
> I think people still refer to MMR on the Profile/Stats page on their DoTA accounts. It is used to calculate approximately the number of matches to win to reach the next level.

It hasn't been a public stat in a very long time, though. The only person that can actually see it is you unless you share a screenshot or something like that. It's effectively useless data-wise, now, in my opinion.

evillman:
Isn't MMR still being tracked for people that didn't change in-game profile and show SOLO MMR: -----???

mobyte0:
> Isn't MMR still being tracked for people that didn't change in-game profile and show SOLO MMR: -----???

No, the value is no longer updated publicly. It shows up as blank because there is not an option to show Solo MMR on your profile anymore.

howardchung:
I think the data is actually still coming back on the backend so people who never changed it can still have it updated.  The client might hide it separately

PR #1855, opened 2018-10-02
MMR tab is commented out

hackyjs:
#1843


howardchung:
I'm not sure we should do this. . . may need to discuss it some more

PR #1862, opened 2018-10-05
1843 remove mmr tab

### LINKED ISSUE #1843: Remove MMR tab from player profiles

bdbch:
This resolves #1843 

## Why did I remove the MMR page completely?

I think it doesn't make sense to keep the tab on the profile and removing it from the repo should clearify that it's obsolete and replaced by player rankings.


mobyte0:
I think this is a duplicate of #1855 except this one removes the code for the MMR page entirely.

bdbch:
I see, I think we should decide to remove it or not to remove it. I think just uncommenting it out just doesn't gives a fixed decision on this topic.

howardchung:
Yeah, we should discuss this some more. My opinion is that there's still value in providing this data in case someone wants to check MMR progression from several years ago, but I'm interested in hearing what others think. Perhaps we can de-emphasize it somehow

bdbch:
Didn't think of it from the history viewpoint. Maybe it makes sense to keep it.

C-Kenny:
Could always mention that it's only historic season 1? MMR tracking, right now it's a bit confusing as the data just drops off.

PR #1867, opened 2018-10-06
remove polyfills

howardchung:
save some bytes from the bundle

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,16 +1,4 @@
 /* eslint-disable react/jsx-filename-extension */
-import 'core-js/fn/object/entries';
-import 'core-js/fn/object/values';
-import 'core-js/fn/array/includes';
-import 'core-js/fn/array/find';
-import 'core-js/fn/array/find-index';
-import 'core-js/fn/array/fill';
-import 'core-js/fn/array/from';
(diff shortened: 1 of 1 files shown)


howardchung:
does current version load? (stage.opendota.com)  I thought it broke after CRA2 without this change

gu3st:
It's missing a Promise library now (in ie11).  

I wonder about trying to use polyfill.io (again) instead of compiling in the polyfills.  At least with CRA2

gu3st:
So..Looking at this I think we should make a decision.  

* We either look at implementing [this](https://github.com/facebook/create-react-app/blob/master/packages/react-app-polyfill/README.md) which I think is the proper way to include the IE11/IE9 polyfills in CRA2

* We look at killing IE11 support and adding in a screen (localized if possible) that indicates that IE11 is no longer supported and you should upgrade.

I've noticed we have a few more UI issues in IE11 than we did a few months ago when I fixed it last.  It's still usable but there's some serious flex bugs going on (although not unfixable, just needs to look at this: https://stackoverflow.com/questions/28627879/flexbox-not-filling-height-in-ie11).

If we do the former, it might be worth placing a notification bar above the header (in IE only) asking users to upgrade for a better experience.

PR #1868, opened 2018-10-07
Updated Spanish (US) localization. es-US

thepante:
Updated to Spanish some of untranslated strings written in English.


howardchung:
Thanks! This language currently isn't enabled in the selector though, only es-ES is available.  Are there significant differences?

PR #1872, opened 2018-10-10
Fix totals card. Add localization for abbreviated year(s)/month(s)/da…

gu3st:
…y(s)/hour(s)/minute(s)/second(s) to support totals update.

This fixes the issue reported in Discord where the formats used for totals were relying on old formatting style.


src/lang/en-US.json
```diff
@@ -845,17 +845,30 @@
...
   "time_m": "a minute",
+  "time_abbr_m": "{0}m",
   "time_mm": "{0} minutes",
+  "time_abbr_mm": "{0}m",
   "time_h": "an hour",
+  "time_abbr_h": "{0}h",
   "time_hh": "{0} hours",
+  "time_abbr_hh": "{0}h",
   "time_d": "a day",
+  "time_abbr_d": "{0} day",
```
howardchung: think we can just use d here

howardchung: and I don't think we need both singular and plural abbreviations, just one should be good

src/lang/en-US.json
```diff
@@ -845,17 +845,30 @@
...
   "time_dd": "{0} days",
+  "time_abbr_dd": "{0} days",
   "time_M": "a month",
+  "time_abbr_M": "{0} mo.",
```
howardchung: I think just mo without the period for abbreviation is fine

src/lang/en-US.json
```diff
@@ -845,17 +845,30 @@
...
   "time_MM": "{0} months",
+  "time_abbr_MM": "{0} mo.",
   "time_y": "a year",
+  "time_abbr_y": "{0} yr",
```
howardchung: prob just y is fine

src/lang/en-US.json
gu3st: I was just considering multiple languages in case they need to express singular/mutliple days differently.

howardchung: ok.  I think day should still be d for both in english

src/lang/en-US.json
```diff
@@ -845,17 +845,30 @@
...
   "time_dd": "{0} days",
+  "time_abbr_dd": "{0} days",
   "time_M": "a month",
+  "time_abbr_M": "{0} mo",
   "time_MM": "{0} months",
+  "time_abbr_MM": "{0} mo",
   "time_y": "a year",
+  "time_abbr_y": "{0} y",
   "time_yy": "{0} years",
+  "time_abbr_yy": "{0} yrs",
```
howardchung: for english y is fine

src/lang/en-US.json
```diff
@@ -845,17 +845,30 @@
...
   "time_h": "an hour",
+  "time_abbr_h": "{0}h",
   "time_hh": "{0} hours",
+  "time_abbr_hh": "{0}h",
   "time_d": "a day",
+  "time_abbr_d": "{0} day",
   "time_dd": "{0} days",
+  "time_abbr_dd": "{0} days",
   "time_M": "a month",
+  "time_abbr_M": "{0} mo",
```
howardchung: for consistency remove the space for english

PR #1880, opened 2018-10-11
Moved MMR tab to clickable medal on profile

mobyte0:
Idea from: https://github.com/odota/web/pull/1862#issuecomment-428416324


howardchung:
My concern with this is that it's not very discoverable

Argn0:
It's not working for me.

gu3st:
@howardchung Given MMR (as a numerical value at least) isn't public, I think it makes sense to somewhat move it from the top level profile navigation. 

I suppose the other solution would be to start tracking historical rank tiers and move that page from "mmr" to "rank", where you can see both the historical MMR data and then rank tier changes as well. 

Also FWIW It didn't work for me either.  The page loads forever.  It also does a hard route change rather than a SPA route change.

PR #1885, opened 2018-10-13
Overview Tables

Argn0:
https://deploy-preview-1885--opendota-stage.netlify.com/matches/4165759465


howardchung:
general comments:

* I think the table itself should be left aligned so it aligns with the ability tables
* sort arrow should probably be on the right side?

Argn0:
-if we put the sort arrow on the right side then it will be hidden behind the mouse cursor

-I think removing the material-ui Table dependency should be done in a separate PR. It will break some things. For example the fixed first column when scrolling . And I think the check boxes in the Laning tab are a part of material-ui Table ?

src/utility/index.jsx
```diff
@@ -518,7 +518,7 @@ export function abbreviateNumber(num) {
   if (!num) {
     return '-';
   } else if (num >= 1000 && num < 1000000) {
-    return `${Number((num / 1000).toFixed(1))}${strings.abbr_thousand}`;
+    return `${Number((num / 1000).toFixed(1))}k`;
```
howardchung: was this intentional?

PR #1889, opened 2018-10-13
Feature/check same team

ichbinsalman:
This is an improvement to player profiles where a user can see if they have played any matches with a specific player. The current match list does not show if the other player was in the same team as the user or not.

With this changeset, it is visible if the player was with or against the user under the match result.

image

### DIFF
--- a/src/actions/transformPlayerMatches.js
+++ b/src/actions/transformPlayerMatches.js
@@ -1,3 +1,7 @@
+import {
+  isRadiant,
+} from '../utility';
+
 export default function transformPlayerMatches(fields) {
   return (response) => {
     // Check if player match parameters have an included account id, and add an extra variable which specifies if that player id is in the same team
@@ -5,7 +9,7 @@ export default function transformPlayerMatches(fields) {
       return response.map((match) => {
         let sameTeam = false;
         const partnerHero = Object.values(match.heroes).find(hero => hero.account_id && hero.account_id.toString() === fields.included_account_id);
-        if ((partnerHero.player_slot < 5 && match.player_slot < 5) || (partnerHero.player_slot > 127 && match.player_slot > 127)) {
+        if ((isRadiant(partnerHero.player_slot) && isRadiant(match.player_slot)) || (!isRadiant(partnerHero.player_slot) && !isRadiant(match.player_slot))) {
           sameTeam = true;
         }
         return { ...match, sameTeam };
--- a/src/actions/transformPlayerMatches.js
+++ b/src/actions/transformPlayerMatches.js
@@ -9,7 +9,7 @@ export default function transformPlayerMatches(fields) {
       return response.map((match) => {
         let sameTeam = false;
         const partnerHero = Object.values(match.heroes).find(hero => hero.account_id && hero.account_id.toString() === fields.included_account_id);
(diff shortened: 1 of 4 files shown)


src/actions/transformPlayerMatches.js
```diff
@@ -0,0 +1,16 @@
+export default function transformPlayerMatches(fields) {
+  return (response) => {
+    // Check if player match parameters have an included account id, and add an extra variable which specifies if that player id is in the same team
+    if (fields.included_account_id && !Array.isArray(fields.included_account_id)) {
+      return response.map((match) => {
+        let sameTeam = false;
+        const partnerHero = Object.values(match.heroes).find(hero => hero.account_id && hero.account_id.toString() === fields.included_account_id);
+        if ((partnerHero.player_slot < 5 && match.player_slot < 5) || (partnerHero.player_slot > 127 && match.player_slot > 127)) {
```
howardchung: there is an isRadiant utility function that makes this easier

src/actions/transformPlayerMatches.js
```diff
@@ -0,0 +1,16 @@
+export default function transformPlayerMatches(fields) {
```
howardchung: to avoid confusion (all transformations should be functions that operate on the input) i'd just make this file the inner function that operates on and returns the input.  You can put the wrapper function elsewhere (maybe in the actions file) and name it `generateTransformPlayerMatches` or something

src/lang/en-US.json
```diff
@@ -696,6 +696,8 @@
   "td_win": "Won Match",
   "td_loss": "Lost Match",
   "td_no_result": "No Result",
+  "td_partner_with": "With",
```
howardchung: these strings already exist in the file, you can reuse them

src/actions/transformPlayerMatches.js
ichbinsalman: Sorry, I don't quite understand what you mean by this, currently I just followed how it was done for transformTrends.
What I understand from this is the wrapper function should be the actual transform function (as that is the code transforming the json response). But I only want to transform the json if the url query has an included account id (this information won't be available from the json response), hence the transformPlayerMatches function which is getting the query params from actions.

howardchung: yeah, I forgot about transformtrends, this is fine. Thanks!

The naming is a bit confusing but we can fix that later.

src/actions/transformPlayerMatches.js
```diff
@@ -0,0 +1,20 @@
...
+} from '../utility';
+
+export default function transformPlayerMatches(fields) {
+  return (response) => {
+    // Check if player match parameters have an included account id, and add an extra variable which specifies if that player id is in the same team
+    if (fields.included_account_id && !Array.isArray(fields.included_account_id)) {
+      return response.map((match) => {
+        let sameTeam = false;
+        const partnerHero = Object.values(match.heroes).find(hero => hero.account_id && hero.account_id.toString() === fields.included_account_id);
+        if ((isRadiant(partnerHero.player_slot) && isRadiant(match.player_slot)) || (!isRadiant(partnerHero.player_slot) && !isRadiant(match.player_slot))) {
```
howardchung: you can just shorten this to isRadiant() == isRadiant

PR #1890, opened 2018-10-13
Fix #1883 - Add descending sort to tooltips on match graphs

### LINKED ISSUE #1883: Graph popups are not sorted by value in descending order
**Current Behavior**
In the Graphs page of a match, mousing over any of the networth/xp/last-hits graphs produces a tooltip showing each player's value at that time in the game. These entries are ordered by the default player order (blue->teal->purple->...->brown)

**Expected behavior/code**
The entries should be ordered in descending order. ie. if the yellow player has the highest networth at 15 mins, mousing over the 15 min mark on the graph should create a tooltip where he is at the top.

This is how the graphs used to work but seems to be bugged for some time now. It would also mirror the functionality of the in-game graphs.

**Console Output**
none

**Environment**
- Browser Version: Chrome Version 70.0.3538.54 (Official Build) beta (64-bit)
- OS: Windows 7 Ultimate 64-bit

**Additional context/Screenshots**

In the following pic: TA (olive) is No.1 networth at 20 mins, so it should be at the top; Grimstroke (purple) should be at the bottom. The entries are instead sorted by player slot, from blue to brown
image

Also: I am trying to start contributing to open-source, and would like to tackle this issue myself!

jagmeetb:
This pull request fixes #1883: Graph popups are not sorted by value in descending order

The tooltip when hovering over graphs will now sort correctly.

### DIFF
--- a/src/components/Visualizations/Graph/MatchGraph.jsx
+++ b/src/components/Visualizations/Graph/MatchGraph.jsx
@@ -40,7 +40,7 @@ const CustomizedTooltip = ({ label, payload }) => (
     (
       <div value={data.value} className={`data ${i < 5 && 'isRadiant'}`} style={{ borderLeft: `8px solid ${data.color}` }}>
         {data.dataKey}: {data.value}
-      </div>)).sort((a, b) => a.props.value < b.props.value)
+      </div>)).sort((a, b) => b.props.value - a.props.value)
     }
   </StyledCustomizedTooltip>
 );


howardchung:
For some reason I can't see the original issue, e.g. https://www.opendota.com/matches/4156391118/graphs appears to have sorted tooltips. Do you have a match where this occurs?

jagmeetb:
This is happening on all matches for me, including the one you linked. 

I tested it some more now and was able to recreate it on Chrome (on Win7 and Android) but not on Firefox.

PR #1892, opened 2018-10-14
ward killer

evaldobratti:
show ward killer (if there were one)

in combinantion with odota/parser#24

### DIFF
--- a/src/actions/transformMatch.js
+++ b/src/actions/transformMatch.js
@@ -227,6 +227,8 @@ function transformMatch(m) {
       });
       newPlayer.abilities = arr;
     }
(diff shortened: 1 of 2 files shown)


src/actions/transformMatch.js
```diff
@@ -227,6 +227,8 @@ function transformMatch(m) {
       });
       newPlayer.abilities = arr;
     }
+    newPlayer.hero_name = heroes[player.hero_id].name;
```
howardchung: make sure heroes[player.hero_id] exists, otherwise this will fail
```suggestion
    newPlayer.hero_name = heroes[player.hero_id] && heroes[player.hero_id].name;
```

PR #1900, opened 2018-10-18
 Added Timeout if strings are empty on page load

karthicrajakumar:
- Added Timeout to wait until the strings objects are present. 
- Checks if strings object is empty and rerenders the component.


howardchung:
I think the correct fix to use mapStateToProps to put the strings into the component as a prop so it automatically reloads when strings changes

karthicrajakumar:
Removed the timeout and changed the ExplorerField. The componentWillUpdate was blocking the new searchString to be set.

src/components/Explorer/ExplorerFormField.jsx
```diff
@@ -35,6 +35,19 @@ class ExplorerFormField extends React.Component {
           searchText,
         });
       }
+    } else if (this.autocomplete && this.autocomplete.state.searchText) {
```
howardchung: isn't this the same code as the other block?

karthicrajakumar: No , the previous block is called when the search text is not set. But here the search text is set when we pass from the url directly.

howardchung: yes, but if the inner code is the same, can't you just change the boolean condition to execute the same block instead of duplicating the code?

karthicrajakumar: Yes it is the same . I'll update it.

src/components/Explorer/ExplorerFormField.jsx
```diff
@@ -22,7 +22,7 @@ class ExplorerFormField extends React.Component {
   }
 
   componentDidUpdate(newProps) {
-    if (this.autocomplete && !this.autocomplete.state.searchText) {
+    if (this.autocomplete) {
```
howardchung: probably want to check for this.autocomplete.state as well

PR #1902, opened 2018-10-19
Chatwheel update

mdiller:
Added support for the new chatwheel messages


howardchung:
You can also remove the old audio files

howardchung:
We might want to consider loading the chatwheel data on-demand using webpack dynamic import, it adds 31KB to the bundle.

mdiller:
could we make all of dotaconstants load its localized strings data via dynamic webpack import? maybe then we could add support for other languages.

PR #1911, opened 2018-10-25
activity page

Argn0:
https://deploy-preview-1911--opendota-stage.netlify.com/players/123485937/activity

The expanded table will be cut off because I adjusted the height to work with the overhauled table in the other PR where the table rows are 3px shorter due to removed borders.

It will work fine when both changes are merged.


howardchung:
LGTM. This looks great!

PR #1915, opened 2018-11-01
recent matches filter

Argn0:
fixes https://github.com/odota/web/issues/1914

https://deploy-preview-1915--opendota-stage.netlify.com/players/117421467?hero_id=102

I believe this is the most user friendly and intuitive solution.


howardchung:
should probably hide the last 20 matches summary when filter is applied

Argn0:
> should probably hide the last 20 matches summary when filter is applied

ah yes I think that happens when you go through some URL where there was a filter already applied so the redux store for recent matches is empty.

Let me see how I can fix this.

Argn0:
OK, I tried to address the issue in the last commit.

PR #1926, opened 2018-11-14
Explorer: default to "kills"

Argn0:
make it default to "kills" if someone selects something on the group by but not the select field


howardchung:
Opinion on just making a preselected default like min date? Might also be more obvious than having a silent default

Argn0:
> Opinion on just making a preselected default like min date? Might also be more obvious than having a silent default

To be honest if I didn't know about it my very first instinct would be to try to remove the "kills" select.

PR #1935, opened 2018-11-19
Activity: localization functions

Argn0:
we can just let the browser do the localization of the strings


src/components/Player/Pages/Activity/ActivityCalendar.jsx
```diff
@@ -43,7 +47,7 @@ const getNormalizationRatio = (data) => {
 const getTooltip = (date, data) => {
   const dateString =
     date &&
-    date.toLocaleDateString('en-US', {
+    date.toLocaleDateString(langCode, {
```
howardchung: what happens if we pass a language code that the browser doesn't recognize? Should we fallback to en-us?

Argn0: passing an unknown language code will not cause any issues. the parameter is optional and it uses the browser's language preference by default.

PR #1950, opened 2018-11-22
Fix duration columns in /matches and /teams/{teamId}/matches

spudde123:
At some point the duration transformation was changed to show also the side the player is on (for example https://opendota.com/players/87278757/matches). Because the transformation is used also in

https://opendota.com/matches
https://opendota.com/teams/1838315

it just shows 'Dire' in situations when it is either incorrect or it doesn't make sense to show anything. I changed it to show the side also on the page that belongs to a specific team. In the matches page when the side info doesn't make sense, it doesn't show anything.

### DIFF
--- a/src/utility/index.jsx
+++ b/src/utility/index.jsx
@@ -790,12 +790,18 @@ export const transformations = {
   last_played: (row, col, field) => <FromNowTooltip timestamp={field} />,
   duration: (row, col, field) => {
     const { strings } = store.getState().app;
+    const playerSideExists = row && typeof row.player_slot !== 'undefined';
+    const playerIsRadiant = playerSideExists && isRadiant(row.player_slot);
+    const teamSideExists = row && typeof row.radiant !== 'undefined';
+    const teamIsRadiant = teamSideExists && row.radiant;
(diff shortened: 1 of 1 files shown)


Argn0:
I noticed that we don't display anywhere when the match took place. That's not so good.

spudde123:
Ah right, I think that info was previously shown under the duration but was changed in the process of the columns in the player pages being rearranged. Perhaps in the pro games we could drop the game id and put the league name (with a link to the match page) and the match time under it to the first column.

Argn0:
Maybe put the league below result. and date below match ID. then it would be similar to the player matches page.

Argn0:
We can also just put league and date next to each other separated with a slash

spudde123:
I put the date now next to the league.

If it's okay like this, I could make a separate function for the display logic so it doesn't get repeated in 3 places.

spudde123:
I added also the updated player images from TI8.

In Discord a user reported that different match tabs show the winner badge on different sides. For example compare https://opendota.com/matches/4236856959/overview to https://opendota.com/matches/4236856959/benchmarks . I found that the problem was that the match winner information was missing from the tabs (other than Overview)

PR #1976, opened 2018-11-30
fix table bug in safari

Argn0:
fixes https://github.com/odota/web/issues/1973

please wait until I can confirm it's fixed before merging

### DIFF
--- a/src/components/App/GlobalStyle.jsx
+++ b/src/components/App/GlobalStyle.jsx
@@ -141,7 +141,8 @@ li {
 }
 
 table {
-  border-collapse: collapse;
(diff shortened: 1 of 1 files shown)


Argn0:
OK, it seems to be fixed now.

src/components/App/GlobalStyle.jsx
```diff
@@ -141,7 +141,8 @@ li {
 }
 
 table {
-  border-collapse: collapse;
+  border-collapse: seperate;
```
howardchung: https://www.w3schools.com/cssref/pr_border-collapse.asp

spelling?

PR #1986, opened 2018-12-07
Search fix

Argn0:


Argn0:
but where do I put the if-condition then?

Argn0:
oh you mean a regular function? let me check if that works

src/actions/index.js
```diff
@@ -29,7 +29,12 @@ export const getSearchResultAndPros = query => dispatch => Promise.all([
   dispatch(setSearchQuery(query)),
   dispatch(getSearchResult(query)),
   dispatch(getProPlayers()),
-  dispatch(getMatch(query)),
+  (() => {
```
howardchung: couldn't this just be wrapped in an if statement instead of adding the IIFE?  Maybe turn this into a full function, build the array and then Promise.all that

Argn0:
How about this one?

PR #1988, opened 2018-12-09
Last Played Sorting

nisargshh:
Made it so you can sort name and last played when searching for a player.


howardchung:
I preferred the look of the previous version, but not sure how to get sorting to work there.

The sort by date doesn't seem to work:
https://deploy-preview-1988--opendota-stage.netlify.com/search?q=howardc

PR #1989, opened 2018-12-09
Adjust larger font-size for TabBar component.

hungvt13:
Since the additional tabs are added to the TabBar making the text hard to view. I adjusted it so that they are more visible. Also I kept the same font size for small screen devices.

### DIFF
--- a/src/components/TabBar/TabBar.jsx
+++ b/src/components/TabBar/TabBar.jsx
@@ -20,7 +20,7 @@ const StyledSection = styled.section`
     /* Tab */
     text-align: center;
     font-weight: ${constants.fontWeightLight};
-    font-size: 13px;
+    font-size: 16px;
     color: ${constants.primaryTextColor};
     padding: 10px 9px 16px;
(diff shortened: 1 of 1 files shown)


howardchung:
I believe my primary reason for the font size change was to avoid the tabs running onto two lines:
https://deploy-preview-1989--opendota-stage.netlify.com/matches/4260113474

hungvt13:
Hello Howard, on my laptop it's all on one line, also the resolution is 1280 x 800. I tested a range of widths and the tabs will eventually running running into two lines unless we make use of calc().
screen shot 2018-12-09 at 19 58 48

howardchung:
I think on 1920 x 1080 it goes two lines.  I think we should try for getting it on one line for that (it's the most common resolution) and not worry about smaller viewports.

howardchung:
does #1998 somewhat resolve this?

Argn0:
> does #1998 somewhat resolve this?

In my opinion it does.

Issue #1842, opened 2018-09-28
View damage dealt/received to each hero by type


mobyte0:
I think it could be useful to put damage types (physical, magical, pure) somewhere in the combat tab. It wouldn't replace any of the current systems set in place but I think that having that data at a glance could be beneficial.

PR #1991, opened 2018-12-10
Add damage per minute column on Match overview

hbsock:
This is for #1842, I've added an extra column to display DPM (Damage per minute).
Please tell me if there's other changes that should be done in order for this PR to be accepted.

Thanks.

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -242,16 +242,13 @@ export default (strings) => {
         tooltip: 'Damage per minute',
         sortFn: true,
         sumFn: true,
-        displayFn: row => { 
-            let dmg = row.damage;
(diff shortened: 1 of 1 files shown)


Argn0:
This is damage done on all units including neutral and lane creeps etc.

I think it would make more sense to have this in the benchmarks tab and not match overview.

And if this will be in the match overview then it should not be grouped separately but together with the other damage type columns. (HD/ TD/ HH).

Argn0:
I actually feel strongly that it should not be in the overview table.

howardchung:
Yeah I'm not sure it's the best fit on the overview, especially since we're already very tight for space.

I think the damage object only tracks damage to heroes.  Maybe this could go in combat?

PR #2006, opened 2019-01-26
get user language from navigator

spitfire77:


src/actions/index.js
```diff
@@ -66,12 +66,14 @@ export const getPlayerTotals = (accountId, params) => action('playerTotals', pro
...
+  const defaultLang = getLang(userLang) || langs[0];
   const savedLang = window.localStorage && window.localStorage.getItem('localization');
-  const defaultLang = langs[0];
-  const selectedLang = langs.find(lang => lang.value === savedLang) || langs[0];
-  const defData = await import(`../lang/${defaultLang.value}.json`);
+  const selectedLang = getLang(savedLang) || defaultLang;
   const selData = await import(`../lang/${selectedLang.value}.json`);
-  dispatch({ type: 'strings', payload: { ...defData, ...selData } });
+
+  dispatch({ type: 'strings', payload: { ...selData } });
```
howardchung: I think we still need defData to default to english strings since some languages aren't translated completely

spitfire77: we have defData here `const selectedLang = getLang(savedLang) || defaultLang;` and here `const defaultLang = getLang(userLang) || langs[0];`

spitfire77: well, I rewrote it in next commit to make it easier to understand what is happening)

src/actions/index.js
```diff
@@ -66,12 +66,14 @@ export const getPlayerTotals = (accountId, params) => action('playerTotals', pro
...
+  const userLang = window.navigator.language;
   const defaultLang = langs[0];
-  const selectedLang = langs.find(lang => lang.value === savedLang) || langs[0];
-  const defData = await import(`../lang/${defaultLang.value}.json`);
-  const selData = await import(`../lang/${selectedLang.value}.json`);
-  dispatch({ type: 'strings', payload: { ...defData, ...selData } });
+  const lang = getLang(savedLang) || getLang(userLang) || defaultLang;
+  const data = await import(`../lang/${lang.value}.json`);
+
+  dispatch({ type: 'strings', payload: { ...data } });
```
howardchung: isn't this still going to import one set of strings? the problem is that we need to merge the translated strings with the english strings as the files don't get the new keys immediately

spitfire77: hm really, i forgot about fallback :)

howardchung:
looks like a lint issue

PR #2009, opened 2019-01-27
Solving Issue Page is too wide for browser 1200x800

HenriqueFadoni:
I've found the issue about the widescreen. It was in line 86 in App.jsx, for the most surprising reasons I deleted it and all the application keeps working just fine.


howardchung:
It looks like this is superseded by #2010

PR #2010, opened 2019-01-27
Screen to wide for browser - Solved

### LINKED ISSUE #1609: Page is too wide for browser
**Current Behavior**
Match content overflows side of screen
image

**Environment**
ChromeOS Version 66.0.3359.158 (Official Build) (64-bit)

**Possible Solution**

Min width is set as 1200px which seems to be too large.

HenriqueFadoni:
I solved the issue on line 86, App.js

fixes #1609

### DIFF
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -83,6 +83,7 @@ const StyledBodyDiv = styled.div`
   flex-grow: 1;
 
   @media only screen and (min-width: ${constants.appWidth}px) {
+    width: ${constants.appWidth === 1200 ? 'auto' : constants.appWidth};
     margin: auto;
   }
 `;
@@ -139,7 +140,7 @@ class App extends React.Component {
     document.body.scrollTop = 0;
     document.documentElement.scrollTop = 0;
   }
-  
+
   render() {
     const {
       width, location, strings,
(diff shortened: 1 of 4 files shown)


src/components/App/App.jsx
```diff
@@ -83,7 +83,7 @@ const StyledBodyDiv = styled.div`
   flex-grow: 1;
 
   @media only screen and (min-width: ${constants.appWidth}px) {
-    width: ${constants.appWidth}px;
+    width: ${constants.appWidth === 1200 ? 'auto' : constants.appWidth};
```
howardchung: shouldn't this be a <= rather than strict equality?

HenriqueFadoni: Actually, it shouldn't, because it could cause changes on other screen scales. The main problem with this application is that when the screen is exactly at 1200px one of the divs on the panel go beyond the screen. In order to fix that, we need to get the exact value where it bugs and put in "auto". I didn't find any bugs with this code. Nonetheless,  you can test if you want to and try to find any other mistake, but I am pretty sure it is not the case.

HenriqueFadoni: Ok, changing the value doesn't affect other layouts. But, if you must insist, I will do better. Hold on a moment.

HenriqueFadoni: I tested all the available layouts next to 1200px, and other sizes as well. The layout did not broke, you can see for yourself if you want to.

HenriqueFadoni: I even changed the value of "auto" to 1150px. This way when the browser redirect to 1200 x 800, it will always have a small border in any case of issues.

PR #2015, opened 2019-01-29
Revert "Screen to wide for browser - Solved"

howardchung:
Reverts odota/web#2010

@HenriqueFadoni Apparently this change is causing some layout issues on other pages, see https://github.com/odota/web/pull/2014/files

Will revert for now until we find a more generic solution.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,90 +13,24 @@
       }
     },
     "@babel/core": {
-      "version": "7.1.6",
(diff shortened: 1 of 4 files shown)


HenriqueFadoni:
But, this changes occur in the 1200x800? If it is a yes, of course it is going to happen! Your site is not responsive, all the sizes are set in "px", therefore the program will need to fix that in order to fit to the screen.
A more "generic" solution would be to update your architecture and fix those 'px' measurements. It would be way more organized, nonetheless, people would find bugs even faster.

PR #2023, opened 2019-02-02
Sum pings correctly in performances tab

LeoRiether:
I noticed that when a player doesn't ping during the game the sum of all pings is not shown correctly, as seen below.
Shows "-" instead of number
This pull request aims to fix that.


howardchung:
Got a match ID to verify with?

LeoRiether:
This one [4381393441](https://www.opendota.com/matches/4381393441/performances)

howardchung:
Thanks, LGTM!

https://deploy-preview-2023--opendota-stage.netlify.com/matches/4381393441/performances

PR #2044, opened 2019-02-20
Added Dota Plus badge to player profiles

builder-247:
As per https://github.com/odota/core/pull/1865.


howardchung:
Do you have a sample player ID with plus to see how this looks once core is deployed to production?

builder-247:
> Do you have a sample player ID with plus to see how this looks once core is deployed to production?

Here: `115042342`.

builder-247:
@howardchung how often does the mmr service run per user usually? I've only tested that the function doing the data population works, so it might take a while to update.

howardchung:
Should be after each ranked match

howardchung:
I still think it looks too big, but open to opinions

builder-247:
Yeah looking at it again I have to agree

howardchung:
https://deploy-preview-2044--opendota-stage.netlify.com/players/115042342

Looks a bit squished?

Argn0:
width must be 65px for the correct image ratio.

builder-247:
oops I didn't notice the ratio wasn't 1:1,

PR #2047, opened 2019-02-23
Deferred tabs

Argn0:
https://deploy-preview-2047--opendota-stage.netlify.com/matches/4451380824/

-This improves perceived performance by highlighting the clicked tab immediately.
-Display skeleton loader for expensive pages.

### DIFF
--- a/src/components/TabbedContent/TabbedContent.jsx
+++ b/src/components/TabbedContent/TabbedContent.jsx
@@ -1,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { List } from 'react-content-loader';
+import TabBar from '../TabBar';
+
+class RenderContent extends React.Component {
+  static propTypes = {
+    skeleton: PropTypes.bool,
+    content: PropTypes.shape({}),
+  };
+
+  state = {
+    render: false,
+  };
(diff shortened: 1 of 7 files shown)


src/components/TabbedContent/TabbedContent.jsx
```diff
@@ -0,0 +1,56 @@
...
+    skeleton: PropTypes.bool,
+    content: PropTypes.shape({}),
+  };
+
+  state = {
+    render: false,
+  };
+
+  componentDidMount() {
+    requestAnimationFrame(() => {
```
Argn0: I'm not entirely sure why, but requestAnimationFrame has to be called twice for this to work.

howardchung: weird, but if it works :)

Curious to know what's going on though

howardchung: wonder if just deferring it with a setTimeout would work as well

Argn0: yeah it would, but I think using requestAnimationFrame twice has the least amount of delay.
If I figure out how exactly this works, I'll let you know.

PR #2048, opened 2019-02-25
Lazy load AbilityBuild table and remove unnecessary imports

Argn0:
Those files are in the redux store already. I don't know why they are being imported again. Maybe I'm missing something.


howardchung:
I think those imports pre-dated them being in the Redux store. I think it's fine to remove now that they're loaded at app init.

PR #2066, opened 2019-03-09
Refactor ability build table logic

Argn0:
And simplify item tooltip colors. I think they look much better with less playful colors.

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -375,19 +375,32 @@ export default (strings) => {
     return cols;
   };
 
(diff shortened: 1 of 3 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -375,19 +375,32 @@ export default (strings) => {
     return cols;
   };
 
+  const abilityMapping = (index, upgradesArr, hero) => {
+    const mapping = {
```
howardchung: probably could use a comment, not sure what this is doing

PR #2079, opened 2019-03-23
Combos

Argn0:
closes https://github.com/odota/web/issues/724

### DIFF
--- a/src/components/Combos/Combos.jsx
+++ b/src/components/Combos/Combos.jsx
@@ -1,0 +1,242 @@
+import React from 'react';
+import RaisedButton from 'material-ui/RaisedButton';
+import heroes from 'dotaconstants/build/heroes.json';
+import querystring from 'querystring';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import HeroImage from '../Visualizations/HeroImage';
+import TableSkeleton from '../Skeletons/TableSkeleton';
+import Heading from '../Heading/Heading';
+import ExplorerOutputSection from './../Explorer/ExplorerOutputSection';
+import getQueryString from './getQueryString';
+import HorizontalMenu from '../Visualizations/HorizontalMenu';
+import {
+  StyledHeroSelector,
+  StyledSelectedHeroes,
+  StyledCombos,
+} from './Styles';
+
+const heroesArray = Object.keys(heroes).map(id => heroes[id]);
+
+const HeroSelector = ({
+  id,
+  handleHeroSelection,
+  selected,
+  teamAFull,
+  teamBFull,
+  strings,
+}) => (
+  <StyledHeroSelector>
+    <HeroImage id={id} />
+    <div className={`ts-container ${selected ? 'selected' : ''}`}>
+      <div
+        className={`ts ts-left ${teamAFull ? 'no-event' : ''}`}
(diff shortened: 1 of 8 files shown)


src/components/Combos/getQueryString.js
```diff
@@ -0,0 +1,108 @@
...
+      ${!teamB[i] && i > 0 && !teamB[i - 1] ? `AND pmB${i}.hero_id > pmB${i - 1}.hero_id` : ''}            
+    `).join(' ');
+
+  const joinTeamBConditions = [0, 1, 2, 3, 4]
+    .map(i => `
+      AND ( ( pmA0.player_slot < 5 ) = ( pmB${i}.player_Slot > 5 ) ) 
+    `).join(' ');
+
+  return (
+    `SELECT matches.match_id, matches.start_time,
```
howardchung: we probably want to limit the number of results returned

src/components/Combos/Combos.jsx
```diff
@@ -0,0 +1,265 @@
...
+          subtitle={`${
+            (this.state.queryResult.rows || []).length
+          } ${strings.explorer_num_rows}`}
+        />
+        {this.state.loading ? (
+          <TableSkeleton />
+        ) : (
+          <ExplorerOutputSection
+            rows={this.state.queryResult.rows}
+            fields={this.state.queryResult.fields}
```
howardchung: it would also be nice to show whether team A or team B won that match (can use the trophy icon from matches page)

src/components/Combos/Combos.jsx
```diff
@@ -0,0 +1,265 @@
...
+);
+
+SelectedHeroes.propTypes = {
+  handleHeroDeSelection: PropTypes.func,
+  strings: PropTypes.shape({}),
+  teamA: PropTypes.arrayOf(PropTypes.number),
+  teamB: PropTypes.arrayOf(PropTypes.number),
+};
+
+function asArray(value) {
```
howardchung: This is a fancy trick I used in odota/core to turn things into arrays. It's a bit less readable

`[].concat(input || [])`

Argn0:
added a query select limit of 500 and the trophy symbol next to the winning team

howardchung:
The public matches search by heroes feature is deployed on the backend. Perhaps we can have a switch that toggles between the two but keep the same UI?

Argn0:
> The public matches search by heroes feature is deployed on the backend. Perhaps we can have a switch that toggles between the two but keep the same UI?

yes. I'll get that done.

howardchung:
Do you want to do this in this release, or ship with pro games first and add it later?

Argn0:
> Do you want to do this in this release, or ship with pro games first and add it later?

This release

howardchung:
I'm kind of tempted to store the pro games data in this format as well because it will make the client code a lot less messy

Argn0:
> I'm kind of tempted to store the pro games data in this format as well because it will make the client code a lot less messy

Could that be applied to the old rows too? I think it would be  bad if only data in the new format would be available to query

howardchung:
yeah but I'd have to write a script to fetch the existing data and put it back in the new format

Argn0:
Added the new endpoint

PR #2106, opened 2019-04-23
Update .env

albertcui:


howardchung:
fixed this, forgot CRA enforces that custom variables start with REACT_APP

PR #2112, opened 2019-05-12
Update Percent.jsx

Argn0:
use a fake fill element to stretch the table cell.

fix https://github.com/odota/web/issues/2110

https://deploy-preview-2112--opendota-stage.netlify.com/explorer?minDate=2019-05-05T22%3A00%3A00.000Z&=&group=player&select=kills&limit=&tier=&league=10681&league=10482&league=10810&result=&side=&megaWin=&maxDate=2019-05-09T22%3A00%3A00.000Z&maxDuration=

### DIFF
--- a/src/components/Visualizations/Table/Percent.jsx
+++ b/src/components/Visualizations/Table/Percent.jsx
@@ -7,23 +7,28 @@ import constants from '../../constants';
 const Percent = ({
   percent, altValue, valEl, inverse = false,
 }) => (
-  <StyledContainer>
-    <TitleContainer>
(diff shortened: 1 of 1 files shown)


howardchung:
I was hoping this could be achieved just be setting min-height on the tr or td element

Argn0:
In the past we made efforts to reduce the table height, so I don't know if we should globally add a min-height to table cells.

That being said, the only way I was able to fix this was by adding a greater padding to table cells. I don't know a way to add this padding to only cells that contain this percentage bar that's not super elaborate

PR #2114, opened 2019-05-17
attempt #3 fix percent cell

Argn0:
fix https://github.com/odota/web/issues/2110

https://deploy-preview-2114--opendota-stage.netlify.com/explorer?minDate=2019-05-05T22%3A00%3A00.000Z&=&group=player&select=kills&limit=&tier=&league=10681&league=10482&league=10810&result=&side=&megaWin=&maxDate=2019-05-09T22%3A00%3A00.000Z&maxDuration=


howardchung:
The JSX on TablePercent translates to a React.createElement call which puts the original type into the type property. I don't think that API will change, so it should be OK.

https://reactjs.org/docs/react-api.html#createelement

PR #2134, opened 2019-06-14
simplified tables

Argn0:
https://deploy-preview-2134--opendota-stage.netlify.com/matches/4836456903/overview


src/components/Match/matchColumns.jsx
```diff
@@ -128,7 +128,7 @@ export default (strings) => {
         displayFn: (row, col, field) => field || '-',
         sortFn: true,
         sumFn: true,
-        color: constants.colorGreen,
+        color: 'hsla(123, 25%, 57%, 1)',
```
howardchung: why hsla instsead of rgba or hex?

Argn0: hsla allows to adjust saturation values more easily.

PR #2142, opened 2019-06-20
remove self-closing div

howardchung:
Since HTML doesn't actually support self closing div this is more correct?


Argn0:
yes, this will behave the same behavior and correct syntax.

PR #2151, opened 2019-06-29
Adding laning graph

litch:
I wanted to work a bit on the analytics of laning, and thought this would be a nice starter task.

image

### DIFF
--- a/src/actions/transformMatch.js
+++ b/src/actions/transformMatch.js
@@ -156,16 +156,19 @@ function transformMatch(m) {
       analysis: analyzeMatch(m, player),
     };
 
-    // compute a cs_t as a sum of lh_t & dn_t
-    player['cs_t'] = (player.lh_t || []).map((v, i) => v + (player.dn_t || [])[i]);
-
     // filter interval data to only be >= 0
     if (player.times) {
-      const intervals = ['lh_t', 'cs_t', 'dn_t', 'gold_t', 'xp_t', 'times'];
+      const intervals = ['lh_t', 'dn_t', 'gold_t', 'xp_t', 'times'];
       intervals.forEach((key) => {
         newPlayer[key] = player[key].filter((el, i) => player.times[i] >= 0);
       });
+
+      // compute a cs_t as a sum of lh_t & dn_t
+      const csT = (player.lh_t || []).map((v, i) => v + (player.dn_t || [])[i]);
+      newPlayer.cs_t = csT.filter((el, i) => player.times[i] >= 0);
     }
+
+
     // compute damage to towers/rax/roshan
     if (player.damage) {
       // npc_dota_goodguys_tower2_top
--- a/src/actions/transformMatch.js
+++ b/src/actions/transformMatch.js
@@ -164,7 +164,7 @@ function transformMatch(m) {
       });
 
       // compute a cs_t as a sum of lh_t & dn_t
-      const csT = (player.lh_t || []).map((v, i) => v + ((player.dn_t || [])[i] || 0) );
+      const csT = (player.lh_t || []).map((v, i) => v + ((player.dn_t || [])[i] || 0));
       newPlayer.cs_t = csT.filter((el, i) => player.times[i] >= 0);
     }
 
--- a/src/actions/transformMatch.js
+++ b/src/actions/transformMatch.js
@@ -158,13 +158,13 @@ function transformMatch(m) {
 
     // filter interval data to only be >= 0
     if (player.times) {
-      const intervals = ['lh_t', 'dn_t', 'gold_t', 'xp_t', 'times'];
+      const intervals = ['lh_t', 'gold_t', 'xp_t', 'times'];
       intervals.forEach((key) => {
         newPlayer[key] = player[key].filter((el, i) => player.times[i] >= 0);
       });
 
       // compute a cs_t as a sum of lh_t & dn_t
-      const csT = (player.lh_t || []).map((v, i) => v + (player.dn_t || [])[i]);
+      const csT = (player.lh_t || []).map((v, i) => v + ((player.dn_t || [])[i] || 0) );
       newPlayer.cs_t = csT.filter((el, i) => player.times[i] >= 0);
     }
 
--- a/src/actions/transformMatch.js
+++ b/src/actions/transformMatch.js
@@ -155,9 +155,13 @@ function transformMatch(m) {
       dn_ten: (player.dn_t || [])[10],
       analysis: analyzeMatch(m, player),
     };
+
+    // compute a cs_t as a sum of lh_t & dn_t
+    player['cs_t'] = (player.lh_t || []).map((v, i) => v + (player.dn_t || [])[i]);
+
     // filter interval data to only be >= 0
     if (player.times) {
-      const intervals = ['lh_t', 'gold_t', 'xp_t', 'times'];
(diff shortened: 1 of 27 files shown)


litch:
I think that the layout of this map and the table is weird, maybe we should move the table above the map & graph.  

If the actual execution of the page looks reasonable, and it's desirable, I'm happy to extract the new graph code that I added to this page to a different file (but it's all in the `Laning/index.js` file for convenience right now).

src/components/Match/matchColumns.jsx
```diff
@@ -833,6 +833,13 @@ export default (strings) => {
       relativeBars: true,
       sumFn: true,
     },
+    {
+      displayName: "LH First 10 Mins",
+      tooltip: strings.lh_ten,
+      field: 'lh_t',
+      sparkline: true,
+      width: 240,
```
litch: I'm a bit weirded out that I *HAVE* to add the width *here*.  Otherwise any `recharts` chart that I add winds up putting in a big "pseudo-margin-right" that I can't get rid of.

howardchung: I took a look but couldn't find the cause. I'm fine with this as a workaround.

howardchung:
Note: You don't have to edit any of the loc files except English.

src/components/Match/Laning/index.jsx
```diff
@@ -33,12 +33,18 @@ class Laning extends React.Component {
       match, strings, sponsorURL, sponsorIcon,
     } = this.props;
     const { laningColumns } = mcs(strings);
+    const annotatedPlayers = match.players.map((p) => {
+      const cs_t = p.lh_t.map((v, i) => v + p.dn_t[i])
+
+      return {...p, cs_t}
+    })
+
```
litch: Alright, another question.  (I have a lot of questions - thanks for bearing with me).

I would prefer to graph CS over time, rather than just LH.  It's possible to do that if I add this code.


In an ideal world, there would be a `Table` thing that we could pass something like, "the pairwise sum of lh_t and dn_t".  Or *EVEN* better, we could add something to the process in core (which I know where to add it), that would just put a cs_t field onto the `match.players` items.  Then, I guess we would need to re-process all of the matches that we want to see that sparkline thing on, and then ideally, we would do something where we show the `cs_t` if it is available, but if not, fall back to `lh_t`.

litch: If you could give a general 👍 or 👎 to this question, then I'll finish cleanup in general, and have this ready to merge this evening (or tomorrow morning).

Argn0:
It displays a vertical scrollbar in my browser. Can we get rid of it?

src/components/Match/Laning/index.jsx
howardchung: Since cs_t is a pure function of lh_t and dn_t, I think it's more space efficient to just compute it client side instead of adding it to the database in core.

litch:
Alright, this should be good to go

Screenshot 2019-06-30 11 10 06

src/components/Match/Laning/Graph.jsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+  </StyledCustomizedTooltip>
+);
+CustomizedTooltip.propTypes = {
+  payload: PropTypes.arrayOf(PropTypes.shape({})),
+  label: PropTypes.string,
+};
+
+const iconUrlFromHeroKey = (heroKey) => {
+  const heroIds = Object.keys(heroes).filter(k => heroes[k].name === heroKey);
```
howardchung: I think you can just use find() instead of filter() here

howardchung: also consider moving this function to the utility file, it seems pretty generic

src/components/Match/Laning/index.jsx
```diff
@@ -32,8 +33,29 @@ class Laning extends React.Component {
       match, strings, sponsorURL, sponsorIcon,
     } = this.props;
     const { laningColumns } = mcs(strings);
+    const annotatedPlayers = match.players.map((p) => {
```
howardchung: We have a central "post-fetch" transform function where we do these computations on the match object.

See https://github.com/odota/web/blob/master/src/actions/transformMatch.js#L146

src/components/Visualizations/Table/Sparkline.jsx
```diff
@@ -0,0 +1,82 @@
...
+import { StyledCustomizedTooltip } from '../../Visualizations/Graph/Styled';
+import constants from '../../constants';
+
+const CustomizedTooltip = ({ label, external }) => (
+  <StyledCustomizedTooltip>
+    <div className="label">{label} - {label + 1}</div>
+    {external && external[label] && (
+      <div>
+        <div>
+          CS This Minute: {external[label].delta}
```
howardchung: These strings should be localized

howardchung: I'd also recommend indicating which minute it was on the tooltip

src/components/Visualizations/Table/Sparkline.jsx
```diff
@@ -0,0 +1,82 @@
...
+        </div>
+        <div>
+          Cumluative CS: {external[label].cumulative}
```
howardchung: spelling on Cumulative

src/lang/bg-BG.json
```diff
@@ -326,6 +326,7 @@
   "heading_graph_gold": "Злато",
   "heading_graph_xp": "Опит",
   "heading_graph_lh": "Последни удари",
+  "heading_graph_cs": "Последни удари + Denies",
```
howardchung: You can revert changes to all lang files besides English, strings will be added automatically later.

src/actions/transformMatch.js
```diff
@@ -155,9 +155,13 @@ function transformMatch(m) {
...
       analysis: analyzeMatch(m, player),
     };
+
+    // compute a cs_t as a sum of lh_t & dn_t
+    player['cs_t'] = (player.lh_t || []).map((v, i) => v + (player.dn_t || [])[i]);
+
     // filter interval data to only be >= 0
     if (player.times) {
-      const intervals = ['lh_t', 'gold_t', 'xp_t', 'times'];
+      const intervals = ['lh_t', 'cs_t', 'dn_t', 'gold_t', 'xp_t', 'times'];
```
litch: I added the dn_t to the list of things to get filtered.

src/components/Match/matchColumns.jsx
```diff
@@ -806,6 +806,14 @@ export default (strings) => {
             </span>}
           </div>),
     },
+    {
+      displayName: strings.cs_over_time,
+      tooltip: strings.tooltip_cs_over_time,
+      field: 'cs_t',
+      sparkline: true,
+      strings: strings,
```
litch: It feels weird to pass the whole strings object here, but in order to localize the text of the tooltip in the graph that's a table element, it seems necessary. =/

src/components/Visualizations/Table/Sparkline.jsx
litch: The heading in the tooltip box is the minute number (range really).  I can expand this somewhat, but it starts to get a bit cumbersome

src/actions/transformMatch.js
litch: And of course adding the `dn_t` threw an error because some matches don't have the dn_t data calculated.  Reverted.

PR #2153, opened 2019-07-04
Update local docker instructions

dartharaknus:
Just update some stuff on how to develop using docker. Also this is much more faster

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,12 +4,4 @@ ENV NPM_CONFIG_LOGLEVEL warn
 
 WORKDIR /usr/src
 
-COPY package.json /usr/src/
-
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -4,12 +4,4 @@ ENV NPM_CONFIG_LOGLEVEL warn
...
 
-COPY package.json /usr/src/
-
-RUN npm install
-
-COPY . /usr/src
-
-ENV PATH /usr/src/node_modules/.bin:$PATH
-
 CMD ["npm", "start"]
```
dartharaknus:
```suggestion
CMD ["npm", "start:install"]
```

PR #2174, opened 2019-08-11
Fix color of gold amount in story pages

### LINKED ISSUE #2173: Text-Color issue for Gold in post-match Story
**Current Behavior**
Hello, when you look at the **Story** of the post-match the **Gold** is text colored in Black with a dark background making it very difficult te read.

Capture

**Expected behavior/code**
Gold text should use a yellow colored value.

**Environment**
- Browser Version: Brave (0.67.124) based on Chromium: 76.0.3809.100
- OS: Windows 10

Regards,

102:
fixes #2173

### DIFF
--- a/src/components/Match/MatchStory.jsx
+++ b/src/components/Match/MatchStory.jsx
@@ -13,7 +13,7 @@ import {
   formatTemplate,
   formatTemplateToString,
 } from '../../utility';
-import { StyledEmote, StyledStoryNetWorthBar, StyledStoryNetWorthText, StyledStorySpan, StyledStoryWrapper } from './StyledMatch';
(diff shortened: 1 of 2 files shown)


src/components/Match/MatchStory.jsx
```diff
@@ -29,7 +29,9 @@ const GoldSpan = (amount) => {
   const { strings } = store.getState().app;
   return (
     <StyledStorySpan key={`gold_${amount}`}>
-      <font color={constants.colorGolden}>{amount.toLocaleString()}</font>
```
102: image

[source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font)

PR #2183, opened 2019-08-17
Updated player images from TI9

spudde123:


spudde123:
Actually I seem to be missing some images, had some sort of an error extracting them which I didn't notice. Will try again in a bit

PR #2187, opened 2019-08-18
50% indicator for net worth bar

Argn0:
fix https://github.com/odota/web/issues/2186

https://deploy-preview-2187--opendota.netlify.com/matches/4973241356/story

### DIFF
--- a/src/components/Match/StyledMatch.jsx
+++ b/src/components/Match/StyledMatch.jsx
@@ -233,6 +233,17 @@ export const StyledStoryWrapper = styled.div`
 export const StyledStoryNetWorthBar = styled.div`
 
   display: flex;
+  position: relative;
+
+  &::after {
+    content: "";
+    position: absolute;
+    width: 3px;
+    height: 13px;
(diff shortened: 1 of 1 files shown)


src/components/Match/StyledMatch.jsx
```diff
@@ -233,6 +233,17 @@ export const StyledStoryWrapper = styled.div`
...
+  position: relative;
+
+  &::after {
+    content: "";
+    position: absolute;
+    width: 4px;
+    height: 13px;
+    background-color: white;
+    left: 50%;
+    margin-left: -2px;
```
howardchung: what's this for?

Argn0: moves it 2px to the left so it's precisely centered

howardchung: Oh I see it's half of the width value. I generally dislike CSS that's bound to itself like this because it's easy to accidentally break. I don't have a better solution though aside from styling this in JS.

Issue #2184, opened 2019-08-17
Histograms/Trends: Duration display as seconds instead of minutes


GoldMath:
**Current Behavior**
Hello, when you look at your Duration in **Histograms** or **Trends** the labels, median and tooltips are only display as seconds making it difficult to quickly read and understand the values. 

**Expected behavior/code**
A second line in the label, or simply replaced with the converted values to minutes would be welcome.
(38:53 instead of 2312)

**Environment**
- Browser Version: Brave (0.67.125) based on Chromium: 76.0.3809.100
- OS: Windows 10

Regards,

howardchung:
I think recharts has formatting functions we can implement. We should be able to do this.

bdbch:
I created a fix for this, see #2212

PR #2196, opened 2019-09-07
#2184

slavabelb:
Functionality #2184


howardchung:
Thanks, I think this will work but it should be possible to do this with Recharts config options to reduce the amount of code necessary? Haven't had a chance to investigate yet.

slavabelb:
@howardchung So we are using BarChar (http://recharts.org/en-US/api/BarChart) as Histogram. So i found nothing in props that can help reduce amount of code

howardchung:
Maybe we could do the transform in TrendGraph/HistogramGraph instead? I like the idea of keeping the original units in the data model and only transforming it at display time.

howardchung:
I think for TrendGraph we can use the tickFormatter on the YAxis component:
http://recharts.org/en-US/api/YAxis#tickFormatter

PR #2209, opened 2019-10-04
Updated french language translation 

londonjob:
Added french language features


howardchung:
Typically translations go through https://translate.opendota.com, but I assume this is for Hacktoberfest.

PR #2213, opened 2019-10-06
adds max-width to damage inflictor columns

### LINKED ISSUE #404: Radian/Dire damage columns should be equal width
The following looks bad.

opendota

bdbch:
fixes #404

This should fix the issue that damage inflictor columns will take more space in one table than the other. inflicted hero icons should now break after 5 items and should never take more space.

image

image

### DIFF
--- a/src/components/Match/TargetsBreakdown.jsx
+++ b/src/components/Match/TargetsBreakdown.jsx
@@ -87,7 +87,12 @@ const damageTargetIcons = (t) => {
   });
 
   return (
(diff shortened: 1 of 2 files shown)


krazi3:
Nice work @bdbch , But i wanted to submit a PR for this. Spent a couple of hours today identifying the issue .. anyways good job 👍

bdbch:
Sorry @krazi3, was already working on the repo atm. Make sure to still open your PR since you could've chosen a different approach which could make more sense! I also like to see code from others.

PR #2215, opened 2019-10-06
2214 navigation rework

### LINKED ISSUE #2214: Navigation Bar gets to crowded on 13" Macbook Pro
**Current Behavior**
The navigation bar items get the whole navigation very crowded and it's almost overflowing on my 13" Macbook Pro. When I resize the window, it breaks the searchbar + settings button completely.

**Expected behavior/code**
The navigation items could be resized on smaller resolutions or could even collapse automatically.

**Environment**
- Browser Version: Edge Beta 78
- OS: macOS 10.14.5
- Resolution: 2560 x 1600
- Screen size: 13"

**Possible Solution**
We could move the overflowing items into a dropdown menu similar to this: https://codepen.io/kvana/full/ZbENqr

**Additional context/Screenshots**

Default fullscreen view:
image

Resized view:
image

bdbch:
closes #2214 

With this PR I'm starting to rebuild the navigation/headerbar component and introduced some package updates so we can move onwards with functionalities and get around dirty ways like using a Route for the app to have access to the router.

Updated packages:
- React => 16.10.2
- ReactDOM => 16.10.2
- ReactRouterDOM => 5.1.2

I also started implement @material-ui/core since the old material-core npm package was deprecated a long time ago so we should start to reimplement components with the new library from now on + rewrite old components. What do you think about this @howardchung ?

I also implemented the `use-react-router` hook to get direct router access from any functional component.

Screenshots:

Menu default:
image

Menu open:
image

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -5,15 +5,12 @@ import { Link } from 'react-router-dom';
 import useReactRouter from 'use-react-router';
 import ActionSearch from 'material-ui/svg-icons/action/search';
 import IconButton from '@material-ui/core/IconButton';
-import MoreVert from '@material-ui/icons/MoreVert';
+import { MoreVert, Settings, BugReport } from '@material-ui/icons';
 import { Toolbar, ToolbarGroup } from 'material-ui/Toolbar';
-import ActionSettings from 'material-ui/svg-icons/action/settings';
-import Bug from 'material-ui/svg-icons/action/bug-report';
 import LogOutButton from 'material-ui/svg-icons/action/power-settings-new';
 import { Menu, MenuItem } from '@material-ui/core';
 import styled from 'styled-components';
 import LocalizationMenu from '../Localization';
-import Dropdown from '../Header/Dropdown';
 import constants from '../constants';
 import AccountWidget from '../AccountWidget';
 import SearchForm from '../Search/SearchForm';
@@ -24,12 +21,6 @@ import { GITHUB_REPO } from '../../config';
 const REPORT_BUG_PATH = `//github.com/${GITHUB_REPO}/issues`;
 
 const VerticalAlignToolbar = styled(ToolbarGroup)`
-  display: flex;
-  align-items: center;
-  justify-content: center;
-`;
-
-const VerticalAlignDropdown = styled(Dropdown)`
   display: flex;
   align-items: center;
   justify-content: center;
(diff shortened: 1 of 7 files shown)


bdbch:
I also reworked the SettingsGroup component with the new dropdown component:

image

image

So both work the same way

bdbch:
I also updated redux and react-redux and fixed migration issues so we can now use react-redux hooks like `useSelector` or `useDispatch` instead of `Connect`

howardchung:
Old material-ui is supposed to be somewhat difficult to migrate due to incompatibilities. I'd actually lean toward using a new library if we want to move old components off the deprecated library.

src/components/Header/Header.jsx
```diff
@@ -73,6 +81,64 @@ const ToolbarHeader = styled(Toolbar)`
   }
 `;
 
+const LinkGroup = ({ navbarPages }) => {
+  const router = useReactRouter();
```
howardchung: I believe this is the first React Hook we're using. I don't have a ton of experience with them, do they provide a significant benefit to developer experience?

102: Hooks are quite convenient to use & are also awesome to be used with type analysis tools like typescript/flow

bdbch:
@howardchung awesome!

> Old material-ui is supposed to be somewhat difficult to migrate due to incompatibilities. I'd actually lean toward using a new library if we want to move old components off the deprecated library.

Yes, it seems like they rewrote the whole thing and moved things into different packages. We could look for old rewriteable components sporadically (like for example this navigation in this PR) and update them with the new ui library.

> I believe this is the first React Hook we're using. I don't have a ton of experience with them, do they provide a significant benefit to developer experience?

Hooks are awesome when it comes to reusing logic. While in class components your logic is a part of your class or some util function, a hook can bring logic with itself and handle states, update functions etc. itself.

I've wrote some hook libs myself which you could take a look at to see what they're meant to do:

- https://github.com/bdbch/react-localstorage
- https://github.com/bdbch/react-github
- https://github.com/bdbch/react-devto

Also what this allows us to do is to stop transfering router props and the mapping of redux props with new hooks for example

```js
// Push to a specific route
const router = useReactRouter();
router.history.push('/Route');
```

or

```js
// Get strings from our redux store
const strings = useSelector(state => state.app.strings);
```

or

```js
// dispatch redux actions
const dispatch = useDispatch();
dispatch(setSearchQuery(query));
```

---

in general we can replace class components with functional components and extract logic out of them into separated hooks or functions. If you're interested, I could add the next functionality as a FC and ping you in the PR so you can take a look. :)

PR #2219, opened 2019-10-07
refactored navbar styling and logo spacings

bdbch:
Minor adjustments on the navbar including less menu items (Matches, Heroes, Combos) for more focused navigation and moved all other links to the expand menu.

I also increased the left and right padding of the navbar in total and gave the logo a bit more spacing.

Before:
image

After:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "react-ga": "^2.5.5",
     "react-helmet": "^5.2.0",
     "react-markdown": "^4.0.3",
-    "react-redux": "^5.1.1",
+    "react-redux": "^7.1.1",
     "react-router-dom": "^5.1.2",
     "react-scripts": "^2.1.1",
     "react-stripe-checkout": "^2.6.3",
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -162,13 +162,28 @@ class App extends React.Component {
       },
       {
         key: 'header_teams',
-        to: '/combos',
-        label: strings.combos,
+        to: '/teams',
+        label: strings.header_teams,
+      },
+      {
+        key: 'header_explorer',
+        to: '/explorer',
+        label: strings.header_explorer,
+      },
+      {
+        key: 'header_api',
+        to: '/api-keys',
+        label: strings.header_api,
       },
     ];
 
     const drawerPages = [
       ...navbarPages,
+      {
+        key: 'header_combos',
+        to: '/combos',
+        label: strings.combos,
+      },
       {
         key: 'header_distributions',
         to: '/distributions',
@@ -185,19 +200,9 @@ class App extends React.Component {
         label: strings.header_meta,
       },
       {
-        key: 'header_explorer',
-        to: '/explorer',
-        label: strings.header_explorer,
-      },
-      {
         key: 'header_scenarios',
         to: '/scenarios',
         label: strings.header_scenarios,
-      },
-      {
-        key: 'header_api',
-        to: '/api-keys',
-        label: strings.header_api,
       },
     ];
 
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -203,10 +203,12 @@ class App extends React.Component {
 
     const includeAds = !['/', '/api-keys'].includes(location.pathname);
     return (
-      <navigationContext.Provider value={{
-        navbarPages,
-        drawerPages,
-      }}>
+      <navigationContext.Provider
+        value={{
+          navbarPages,
+          drawerPages,
+        }}
+      >
         <MuiThemeProvider muiTheme={getMuiTheme(darkBaseTheme, muiTheme)}>
           <GlobalStyle />
           <StyledDiv {...this.props} location={location}>
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -26,6 +26,7 @@ import Meta from '../Meta';
 import Api from '../Api';
 import Footer from '../Footer';
 import FourOhFour from '../FourOhFour';
+import navigationContext from '../../context/navigationContext';
 import constants from '../constants';
 import muiTheme from './muiTheme';
 import GlobalStyle from './GlobalStyle';
@@ -38,6 +39,7 @@ const StyledDiv = styled.div`
   flex-direction: column;
   height: 100%;
   left: ${props => (props.open ? '256px' : '0px')};
+  margin-top: ${props => (props.location.pathname === '/' ? '-56px' : '0px')};
   background-image: ${props => (props.location.pathname === '/' ? 'url("/assets/images/home-background.png")' : '')};
   background-position: ${props => (props.location.pathname === '/' ? 'center top' : '')};
   background-repeat: ${props => (props.location.pathname === '/' ? 'no-repeat' : '')};
@@ -163,6 +165,10 @@ class App extends React.Component {
         to: '/combos',
         label: strings.combos,
       },
+    ];
+
+    const drawerPages = [
+      ...navbarPages,
       {
         key: 'header_distributions',
         to: '/distributions',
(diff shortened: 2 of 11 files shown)


bdbch:
I also refactored the login button. It won't show the spinning indicator on slow connections but just hide the widget until we know the authorization state.

I also used the @material-ui/core/Button component.

Before:
image

After:
image

It now has the default ripple effect.

src/components/Header/Header.jsx
```diff
@@ -88,9 +88,12 @@ const LinkGroup = ({ navbarPages }) => {
     setAnchorEl(undefined);
   }, [anchorEl]);
 
+  const mainNavbarPages = useMemo(() => navbarPages.slice(0, 3), [navbarPages]);
```
howardchung: Rather than slicing by index, I think we should add a property in each navbar object that indicates whether it's a main one or a drawer one, and render the drawer ones in a burger menu as described in https://github.com/odota/web/issues/2222

howardchung:
If you want to merge the login button changes now, you can revert the navbar changes and submit them in a later PR?

bdbch:
@howardchung sure thing. will do that asap in the evening today.

> Rather than slicing by index, I think we should add a property in each navbar object that indicates whether it's a main one or a drawer one, and render the drawer ones in a burger menu as described in

So we would render the burger menu on desktop too right?

bdbch:
@howardchung this is how it looks like now:
image

image

I guess we still need the User login/logout in the drawer?

bdbch:
The app bar is now also fixed so you can open the drawer wherever you are:
image

src/components/Header/Header.jsx
```diff
@@ -139,105 +144,130 @@ const SettingsGroup = ({ children }) => {
...
+
+const Header = ({
+  location, disableSearch,
+}) => {
+  const [Announce, setAnnounce] = useState(null);
+  const [menuIsOpen, setMenuState] = useState(false);
+  const navigationState = useContext(navigationContext);
+  const small = useSelector(state => state.browser.greaterThan.small);
+  const user = useSelector(state => state.app.metadata.data.user);
+  const strings = useSelector(state => state.app.strings);
```
bdbch: So just for clarity for people who are not used to write components with hooks yet I've decided to add an explanation here on what is going on.

```js
// Here I create a state for the Announce component
// It's the equivalent to this.state.Announce and this.setState({ Announce })
const [Announce, setAnnounce] = useState(null);

// Here I create a second state for the menu drawer state (we can have multiple states with hooks)
// This is the equivalent to this.state.menuIsOpen and this.setState({ menuIsOpen })
const [menuIsOpen, setMenuState] = useState(false);
```

// Here I use a context to pass down navigation pages from the App.jsx
// The useContext hook is a hook coming from React to access data from a context without
// the use of any HoC
// inside we have navigationState.navPages and navigationState.drawerPages
const navigationState = useContext(navigationContext);

// Here I get the small value from redux
// useSelector is a hook from react-redux which allows us to select any value from the current redux state
const small = useSelector(state => state.browser.greaterThan.small);

// Same here, this gets the current user from redux
const user = useSelector(state => state.app.metadata.data.user);

// Same here, we get the apps strings
const strings = useSelector(state => state.app.strings);

src/components/Header/Header.jsx
```diff
@@ -139,105 +144,130 @@ const SettingsGroup = ({ children }) => {
...
+
+  useEffect(() => {
+    import('../Announce').then(ann => setAnnounce(ann.default));
+  }, []);
```
bdbch: This is special. `useEffect` can be used to handle stuff like initialization, value updates etc.

This effect basically is a `componentDidMount` because it will only run once on the first render call and afterwards won't called again. Thats because we passed an empty dependency array to it (see line 222).

Inside the effect we set the announceState to the imported Announce component.

This would look like this in a class:

```js
componentDidMount() {
  import('../Announce').then(ann => this.setState({
    Announce: ann.default,
  }))
}
```

src/components/Header/Header.jsx
```diff
@@ -70,57 +74,58 @@ const DropdownMenuItem = styled(MenuItem)`
...
     </VerticalAlignToolbar>
   );
 };
 
-LinkGroup.propTypes = {
-  navbarPages: PropTypes.arrayOf(PropTypes.shape({})),
-};
-
 const SettingsGroup = ({ children }) => {
   const [anchorEl, setAnchorEl] = useState(null);
```
bdbch: Here I set an anchorEl state which I will use to define were the dropdown menu should be attached.

src/components/Header/Header.jsx
```diff
@@ -70,57 +74,58 @@ const DropdownMenuItem = styled(MenuItem)`
...
   const handleClose = useCallback(() => {
```
bdbch: `useCallback` defines a callback function which will be called whenever handleClose is called. We need useCallback here so the function is not redefined everytime on render.

package.json
```diff
@@ -48,7 +48,7 @@
     "react-ga": "^2.5.5",
     "react-helmet": "^5.2.0",
     "react-markdown": "^4.0.3",
-    "react-redux": "^5.1.1",
+    "react-redux": "^7.1.1",
```
102: > refactored navbar styling and logo spacings 

I think we should do such upgrades via the separate PRs due to high chance of breaking something (2 major versions is not a joke)

src/components/Header/Header.jsx
102: just a nitpick: arrow function inside `useCallback` will still be created on each render & `handleClose` will also be redefined on each render, but `handleClose` will keep consistent reference until dependencies array items changes

package.json
howardchung: I looked over the changelogs, doesn't seem to be anything breaking that we use: https://github.com/reduxjs/react-redux/releases

src/components/App/App.jsx
```diff
@@ -163,16 +165,10 @@ class App extends React.Component {
         to: '/combos',
```
howardchung: I think this somehow got mixed up in a previous change. Combos and Teams are different paths, teams should go in the main navbar and combos in the drawer.

Also, I think we should add Explorer and API to the main navbar.

bdbch:
@howardchung should be fine now

src/context/navigationContext.js
```diff
@@ -0,0 +1,8 @@
+import { createContext } from 'react';
+
+const navigationContext = createContext({
+  navbarPages: [],
```
howardchung: Why do these need to be in a context? Aren't they static and can be imported as constants?

bdbch: We can't move these into easy to import functions without adding extra logic since we use the strings from our Redux store, thats why I'm defining them in a context so we can easily pass them down to any component below.

Otherwise we could export them from the constants file and add the strings key instead of the translated string.

howardchung: The old version just passed the navbarPages down to the Header component. Perhaps we can use the same approach and pass two props?

bdbch: I removed the context, see my latest commit.

howardchung:
Couple of lint errors, otherwise looks good

howardchung:
@bdbch are you able to fix the lint issue? If you're busy, I can do it!

bdbch:
@howardchung I can do it in a few minutes! :) I'll let you know when I updated my PR. thanks!

Argn0:
there seems to be a bug

if you click on this link you land on the wrong tab

https://deploy-preview-2219--opendota.netlify.com/matches/5079685117/combat

bdbch:
Good find @Argn0 I'll try to fix it together with the lint issues now. Had a rough day today so I could just get back right now.

Seems to be an issue with the tab index.

bdbch:
@Argn0 I found the issue and fixed it with my last commit. The problem was that I was mapping through all tabs before filtering them and using the index to navigate/set active states of the tab. Since some tabs are not visible sometimes I moved the filtering out of the rendering block and moved it into a memorized variable.

bdbch:
@howardchung the PR should be fine now

howardchung:
I'm still seeing the issue with the link going to the wrong tab?

src/components/TabBar/TabBar.jsx
```diff
@@ -44,6 +45,8 @@ const TabBar = ({ tabs, match }) => {
     setTabValue(index);
   }, [history]);
 
+  const visibleTabs = useMemo(() => tabs.filter(tab => (!tab.hidden || (tab.hidden && !tab.hidden(match)))), [tabs]);
```
howardchung: isn't the only change here moving the logic? I'm not sure why this would change the behavior.

Argn0:
> I'm still seeing the issue with the link going to the wrong tab?

me too

src/components/TabBar/TabBar.jsx
howardchung: Possibly an issue with the previous commit https://github.com/odota/web/commit/fc10a7cfc401a3f6727271a5d17b041fea331271#diff-571eda7df4f281922d74a5ed5b6a60c7?

bdbch: Let me check again. For me it was working fine.

>isn't the only change here moving the logic? I'm not sure why this would change the behavior.

No it also changes the index while iterating through the list of tabs. That was what caused the issue in the first place since I used the "setActive" index of the unfiltered list, but the filtered list was rendered.

So the active tab was for example `3` while the actual tab index was `2` because one tab was not visible.

Moving it in front of the rendering map allows us to get the correct index. Let me check why it is not working

bdbch: Ah so the problem is fixed but the issue with initially loading the correct tab is broken.
fldCVHrxPe

As you can see the tab is right when you actually click around.

But on initial load the tab is wrong:
YCgskM1WGL

I'll fix the problem!

bdbch: Found the problem. A few lines above in `TabBar.jsx` I used `useEffect` to set the initial tab index without using the filtered tabs list.

Changing it to this:

```jsx
useEffect(() => {
  visibleTabs.forEach((tab, i) => {
    if (location.pathname === tab.route) setTabValue(i);
  });
}, [visibleTabs]);
```

worked fine:
ARJom5Ey3k

bdbch:
> > I'm still seeing the issue with the link going to the wrong tab?
> 
> me too

could you retest this with the latest netlify deploy preview?

Argn0:
> > > I'm still seeing the issue with the link going to the wrong tab?
> > 
> > 
> > me too
> 
> could you retest this with the latest netlify deploy preview?

It seems to work correctly now.

bdbch:
@howardchung can we merge this pr?

PR #2225, opened 2019-10-07
implements new material tooltip

bdbch:
This PR implements the new `@material-ui/core` tooltips for all simple tooltips which don't need special markup or styling and removes several data-tip and ReactTooltip uses.

### DIFF
--- a/src/components/Table/TableHeaderColumn.jsx
+++ b/src/components/Table/TableHeaderColumn.jsx
@@ -1,11 +1,26 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import { Tooltip } from '@material-ui/core';
 import { StyledHeaderCell } from './Styled';
+import { getSortIcon } from './tableHelpers';
 import { getColStyle } from '../../utility';
 
+const HeaderCellContent = styled.div`
+  align-items: center;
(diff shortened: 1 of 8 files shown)


src/components/Table/TableHeaderColumn.jsx
```diff
@@ -27,17 +24,15 @@ const TableHeaderColumn = ({
...
           style={{ color: column.color, width: '100%', textAlign: getColStyle(column).textAlign }}
         >
-          {column.displayName}
-          {column.sortFn && getSortIcon(sortState, sortField, column.field, { height: 14, width: 14 })}
-          {column.tooltip &&
-          <ReactTooltip id={tooltipId} place="top" type="light" effect="solid">
-            {column.tooltip}
-          </ReactTooltip>
-          }
+          { !column.tooltip ? column.displayName : (
```
howardchung: Do we need the sort icon here?

howardchung:
Looks good, one question that I commented on

src/components/Table/TableHeaderColumn.jsx
bdbch: True! will re-add it

bdbch:
could be reviewed again @howardchung

PR #2228, opened 2019-10-09
Fix eslint warning about array key in Abilities.jsx

hooahest:
Eslint gives the following warning:

\workspace\Github\web\src\components\Hero\Abilities.jsx
  27:21  warning  Do not use Array index in keys  react/no-array-index-key

Changed 'abilities' to contain 'id' instead of relying on array index, as advised by https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

### DIFF
--- a/src/components/Hero/Abilities.jsx
+++ b/src/components/Hero/Abilities.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable object-curly-spacing */
 import React from 'react';
 import { connect } from 'react-redux';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


src/components/Hero/Abilities.jsx
```diff
@@ -23,16 +23,16 @@ const AbilityItem = styled.div`
   padding-right: 4px;
 `;
 
-const renderAbilities = abilities => abilities.map((ability, key) => (
-  <AbilityItem key={key}>
-    <Ability {...ability} abilityID={key} />
+const renderAbilities = abilities => abilities.map(ability => (
+  <AbilityItem key={ability.key}>
+    <Ability {...ability.data} abilityID={ability.key} />
```
howardchung: Doesn't this change the abilityID prop being passed?

howardchung: Oh, I see we're adding the key property.

howardchung:
I think this is still using an array index as an ID though, we're just passing it from a layer up to avoid the warning.

bdbch:
@howardchung can we receive the dota2 internal spell id for the skill which we could for iteration keys?

howardchung:
Yeah, presumably `key` is the key of the object in https://github.com/odota/dotaconstants/blob/master/build/abilities.json

PR #2241, opened 2019-10-20
Add tooltip to Hero Rankings' Score column

nguyenphusi:
In reference to #1342, specifically in [the comment](https://github.com/odota/web/issues/1342#issuecomment-347510522), the `Score` seems confusing so I added a tooltip. 
Not sure if this is the best wording so please let me know if you want to change it. Thanks.


howardchung:
LGTM!

PR #2261, opened 2019-11-03
Git Hook running eslint on commit

bdbch:
This implements a pre-commit git hook that runs eslint before a commit. If the linting fails the commit won't be accepted.

This will prevent linting related commits since the linting now is required for a valid commit. It also decreases the CI usage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "npm run lint"
+      "pre-commit": "npm run lint"
     }
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -20,6 +20,11 @@
     "storybook": "start-storybook -p 9009 -s public",
     "storybook:build": "build-storybook -s public"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint"
```
102: https://github.com/okonet/lint-staged can be really helpful to improve developer experience (via running `lint-staged` on `pre-commit` to only check against staged files & `lint` on `pre-push` to ensure ESLint is passing for the whole)

bdbch: This would be a great addition to this check since the current eslint also takes a bit of time to finish. I will create an issue for that later.

PR #2278, opened 2019-11-09
fix for not appearing items

Argn0:
fixes https://github.com/odota/web/issues/2230


howardchung:
I don't even know what this is supposed to do, so a comment might be useful

PR #2285, opened 2019-11-12
general refactor & deprecating CastTable

hooahest:
This is a general refactor of some code I came across. Some variable renaming, a 'for' loop that's replaced with ES6, and a component that doesn't seem to be used at anyplace in the code so I wrote it as deprecated (better to delete?)


src/components/Match/CastTable.jsx
```diff
@@ -34,19 +34,19 @@ const castsColumns = strings => [{
   displayFn: (row, col, field) => field || '-',
 }];
 
-const getCastArray = (pm) => {
+const getCastArray = (player) => {
   // Get from ability_uses, item_uses
   const resultArray = [];
   const targets = ['ability_uses', 'item_uses'];
```
bdbch: Should we also rename this? targets isn't really describing what this is used for. Should be called `castKeys` or something I guess?

hooahest: targetedCastTypes? kinda long but describes it well

hooahest: though I'm also okay with castKeys

hooahest:
retest this please

src/components/Match/CastTable.jsx
bdbch: I'm fine with longer names as long as it makes sense and is easy to understand :)

bdbch:
> hm, I'm not sure why travis failed
> 
> image
> 
> should I just rerun it? how do I do that?

I think it's an issue with the CI, not your code.

src/components/Match/CastTable.jsx
```diff
@@ -55,16 +55,17 @@ const getCastArray = (pm) => {
   return resultArray;
 };
 
+// Deprecated & unused
```
howardchung: If it's not used, I'd just get rid of it. We can always pull it back out of the history.

src/components/Match/matchColumns.jsx
```diff
@@ -1236,10 +1236,8 @@ export default (strings) => {
         totalDuration.push(duration);
       }
     });
-    let total = 0;
-    for (let i = 0; i < totalDuration.length; i += 1) {
-      total += totalDuration[i];
-    }
+
+    const total = totalDuration.reduce((a, b) => a + b, 0);
```
howardchung: I'm fine with this, but the original might be a little easier to read. Not everyone is familiar with the .reduce function.

hooahest: True. This is as basic as a reduce function gets though, so even if someone is not familiar with it, he can probably still understand it (also probably a good way to learn it?)

src/components/Match/CastTable.jsx
hooahest: deleted it

PR #2288, opened 2019-11-13
run tests in firefox

Argn0:
fixes https://github.com/odota/web/issues/2287


howardchung:
i guess let's try this out. I'd prefer to test in chrome/chromium since most of our users are using chrome, but should be easy to change this back if we want.

PR #2290, opened 2019-11-15
Added roles to player tables

hooahest:
Noticed that we're getting each player's role in the data, so I've added it to the table

image

here's how it looks like, might want to add some padding/margin

image

solves(?) this issue https://github.com/odota/core/issues/1844

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -56,6 +56,7 @@ export default (strings) => {
       predictedVictory={row.pred_vict}
       leaverStatus={row.leaver_status}
       hero={compileLevelOneStats(heroes[row.hero_id])}
+      role={row.desc}
     />);
   };
 
--- a/src/components/Visualizations/Table/HeroImage.jsx
+++ b/src/components/Visualizations/Table/HeroImage.jsx
@@ -481,6 +481,7 @@ class TableHeroImage extends React.Component {
       leaverStatus,
       strings,
       hero = {},
+      role,
     } = this.props;
     const { tooltipVisible } = this.state;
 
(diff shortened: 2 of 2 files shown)


Argn0:
I don't like how this makes the tables taller. Maybe we should use the lane icons instead and put them in their own column?

hooahest:
> I don't like how this makes the tables taller. Maybe we should use the lane icons instead and put them in their own column?
> 
> Also we would need support/core icons.

some of the rows are actually the same length.

image

Any hero with a backpack item is the same height as before

though this is only for the 'overview' tab, a lot of other tables are indeed too tall now

image



> Maybe we should use the lane icons instead and put them in their own column?

I kind of like how you can see a hero's role in each tab, that way you can have more information for wards/tower damage/etc depending on role. 
We can use the lane icons and have a tooltip when hovering on them with the role description, what do you think?

Argn0:
I think there is enough space in the overview table to have both icons next to each other in their own separate column.

If we made the icons a part of the hero-image component and displayed them in all tables then it would look really awkward in some of the tables.

howardchung:
Agreed that it looks a bit cramped/busy on its own line. Maybe we can try icons, or possibly inlining it with the rank line, but that might be confusing.

ahi-el:
@howardchung do you have any icons in mind?

We can do a simple number (safe core 1, mid 2, off core 3, 4 and 5 for off/safe supports, but that won't account for trilanes / junglers

how about a simple icon for core/supp (or text?) and having that be a tooltip for the more descriptive role?

howardchung:
We have role icons here, e.g. https://github.com/odota/web/blob/master/public/assets/images/dota2/lane_1.svg

That includes off/mid/safe/jungle

PR #2292, opened 2019-11-16
Removed coach buttons and text

hooahest:
as per issue https://github.com/odota/web/issues/2289


hooahest:
didn't see that this was already opened in https://github.com/odota/web/pull/2291, can close this then as he added other stuff too

PR #2294, opened 2019-11-17
deferred rendering in overview tab

Argn0:
this improves performance quite a bit.


src/components/Match/Overview/Overview.jsx
```diff
@@ -6,10 +6,13 @@ import AbilityDraftTable from '../AbilityDraftTable';
...
 import BuildingMap from '../BuildingMap';
 import Collapsible from './../../Collapsible/index';
-import AbilityBuildTableSkeleton from './../../Skeletons/AbilityBuildTableSkeleton';
+import AbilityBuildTable from '../AbilityBuildTable';
+import deferredComponent from './../../DeferredComponent/index';
 
-const AbilityBuildTable = React.lazy(() =>
-  import(/* webpackChunkName: 'AbilityBuildTable' */ '../AbilityBuildTable'));
+
+const DeferredAbilityBuildTable = deferredComponent(AbilityBuildTable);
```
howardchung: Any reason we can't just make this a JSX component?

Argn0: There is no particular reason. I guess we could use a render prop.

howardchung: I think I prefer that pattern a little more.

howardchung: or even just this.props.children?

Argn0:
Tested on this page https://deploy-preview-2294--opendota.netlify.com/matches/5117971814/overview

going from the Benchmarks to Overview tab is measured ~300ms faster for me.

Argn0:
Also it seems like the biggest performance hit seems to come from the number of icons in the Ability Build tables.

PR #2303, opened 2019-11-24
Implements fixed gradient stops to prevent the background color jumping around

### LINKED ISSUE #2298: Background brightness changes when changing tabs
There is this weird bug where switching between the Benchmarks and Overview tab very slightly changes the background brightness.

https://www.opendota.com/matches/5118002212

EDIT: it's on the right side

**Environment**
- Browser Version: chromium
- OS: win 10

bdbch:
closes #2298

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+    "singleQuote": true,
+    "trailingComma": "all"
+}
--- a/src/components/App/GlobalStyle.jsx
+++ b/src/components/App/GlobalStyle.jsx
@@ -1,7 +1,9 @@
 import { createGlobalStyle } from 'styled-components';
+
 import constants from '../constants';
 
-const GlobalStyle = createGlobalStyle([`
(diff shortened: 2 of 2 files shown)


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
howardchung: we don't need this if we aren't using prettier?

src/components/App/GlobalStyle.jsx
```diff
@@ -1,7 +1,9 @@
 import { createGlobalStyle } from 'styled-components';
+
```
howardchung: don't think we need this line

howardchung:
@bdbch were you able to resolve the comments?

bdbch:
I'll close this one since @Argn0 already fixed this in the new color scheme PR (seems so).

I finally have more time for OS projects over the holidays so expect some more work from my side. :)

PR #2331, opened 2019-12-15
Add immortal to rank tier

NoPhaseNoKill:
Added immortal to rank tier. Let me know if you'd like me to regularly contribute.

This project should also be using typescript IMO


NoPhaseNoKill:
The match count numbers, are these meant to be reflective of MMR? Initially it was 1000-5000, then got refactored, and now it's in this current state. Looks to be like that's what it was meant to be initially, but this doesn't quite make sense in the current state (divine's are not 7k MMR). I can tidy this up if so.

NoPhaseNoKill:
Closing due to bug found with table not rendering extra columns

PR #2332, opened 2019-12-15
Add immortal to rank tier, and icons, to heroes page 

NoPhaseNoKill:
Added immortal rank tier and rank tier icons to hero page.

There's currently a bug I'm having troubles with - is someone able to help? The table isn't rendering the last few columns. This is due to the CSS on 'innerContainer', but I have no idea how to fix this and don't want to break all table. When disabling the beneath, table renders, but doesn't appear to adjust accordingly.

@media only screen and (min-width: 1200px)
<style>
.ioVduP .innerContainer {
    /* overflow-x: hidden !important; */
}
<style>
.ioVduP .innerContainer {
    /* overflow-y: hidden !important; */
    /* overflow-x: auto !important; */
}

### DIFF
--- a/src/components/Heroes/columns.jsx
+++ b/src/components/Heroes/columns.jsx
@@ -18,6 +18,8 @@ export default (strings) => {
     sortFn: row => (heroes[row.hero_id] && heroes[row.hero_id].localized_name),
   };
 
+  const getRankIcon = number => `/assets/images/dota2/rank_icons/rank_icon_${number}.png`;
+
   return {
     pro: [heroColumn, {
       displayName: strings.hero_pick_ban_rate,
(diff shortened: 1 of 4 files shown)


NoPhaseNoKill:
DO NOT merge until the above bug is fixed, as it will not display tables correctly.

howardchung:
We'd love to have regular contributions!

The match count is not MMR but the number of matches where that hero was picked in the sample.

The table is supposed to scroll in viewports where it's too narrow to see all the columns.

howardchung:
I think that CSS you pasted makes it not have the scrollable ability at 1200px or greater, since previously that was wide enough to hold all the columns. That's no longer true with the wider columns, so bumping that number would probably work.

NoPhaseNoKill:
I'm not quite following what you mean by match count. Correct me if I'm wrong here, but isn't each of the following a representation of the rank someone is in? I know the code says matchCount, but to me the data looks like it's for each rank? `const matchCountPro = json.map(heroStat => heroStat.pro_pick || 0).reduce(sum, 0) / 10;
    const matchCount8 = json.map(heroStat => heroStat['8_pick'] || 0).reduce(sum, 0) / 10;
    const matchCount7 = json.map(heroStat => heroStat['7_pick'] || 0).reduce(sum, 0) / 10;`

NoPhaseNoKill:
Which number specifically should I bump? Where abouts do I change the value of the 1200px?

howardchung:
See https://github.com/odota/web/blob/fa922635acc3308947dc988c90559537e7a730b8/src/components/Table/Styled.jsx#L90

Yes, we partition the number of times each hero was picked by rank (so you can see what heroes are more popular in certain tiers)

Argn0:
this table should not be horizontally scrollable else it has horrible usability.

Argn0:
at least not at full window size

Argn0:
reduced padding. still doesnt fit :(

NoPhaseNoKill:
Sorry I've just added some clarification around that with my previous post. Yep my thoughts exactly around the horizontal width of the table. 1500 I'd imagine would do it, which shouldn't make it take up the full screen should it?

Argn0:
try making hero names shorter

NoPhaseNoKill:
And where would I do that sorry?

howardchung:
I believe 8_pick is generated by the server code dynamically, so you'd have to modify that too if you want to change that.

I'm actually not sure if modifying that 1200 is the best approach. We may have other containers that max out at 1200px to avoid the page getting too wide.

The left column of heroes does stay fixed/frozen when the table scrolls., which makes it a bit more usable.

Argn0:
lets see if it fits now

NoPhaseNoKill:
That's close, but it's about 4/5 pixels out still

NoPhaseNoKill:
That now looks perfect. Nice work :)

NoPhaseNoKill:
I also found a bug that I just pushed a fix for. I think this is ready to go if you guys are happy

NoPhaseNoKill:
Off topic question: This codebase looks like it could benefit from a slow migration to typescript. At the moment it's hard to decipher what's being passed around where, particularly around the structure you've chosen for your dynamic tables. 

Traditionally I would have either: 

1) Create a table which takes rows, columns, headers etc (the data) - and then create components which have the responsibility of doing all the work of the formatting, and then just calling the table. Was there a reason you chose not to do this? 

2) The documentation of having props typed, alongside the benefit that the IDE gives you - would make this project a whole lot easier to contribute to. Thoughts?

howardchung:
TypeScript: Agreed. https://github.com/odota/web/pull/2211 started this effort, but hasn't been finished yet. If you want to help with that, it might be a good starting point!

The data is an array, which is passed to the table. Since each field in the data might be formatted differently, we also pass it a "columnDefs" array which defines how to transform each row into the data. Components can be used in the displayFn of each columnDef.

howardchung:
Getting an error on https://deploy-preview-2332--opendota.netlify.com/heroes

NoPhaseNoKill:
Ah gotcha, makes sense. Will take a look at the typescript thread.

NoPhaseNoKill:
Fixed has just been pushed for error. Apologies

PR #2333, opened 2019-12-15
fix column colors

Argn0:


Argn0:
I'm closing this for now. I'll think of something prettier after the other PR is merged.

Argn0:
OK, reopening this. This solution has the best readability.  Striped columns in only 2 different colors was really confusing.

PR #2357, opened 2019-12-21
Typescript integration

NoPhaseNoKill:
This commit adds typescript to the service.

Currently you can see that linting rules are working correctly (getting two warnings for my test files).


NoPhaseNoKill:
Had to bump ESLint to minimum 6, which requires an upgrade of react-scripts (so that we can override the linting rules to completely seperate JS/JSX files from being linted using new typescript parser). Fixing now

NoPhaseNoKill:
Please note: I explicitly chose not to copy all of the linting rules over from the JS/JSX to TS/TSX files. These imo need discussion, as I have a few qualms with the way the current linting rules are setup. Namely the use of default export, rather than named. There may be others, but this has been the most prominent I've noticed.
. 

1. It encourages the mass index.js pattern, which has two problems:
 - IDE: 'ctrl + click' manoeuvrability is lost
 - Code becomes borderline unreadable when you've got no idea where the file is being used. Also related to the above

2. You also lose the ability to dynamically import from the IDE. I'm having to manually type the file paths, and that just wastes so much time.

What are people's thoughts on this?

package.json
```diff
@@ -66,19 +72,45 @@
     "reselect": "^4.0.0",
     "seamless-immutable": "^7.1.3",
     "styled-components": "^4.4.1",
+    "typescript": "^3.7.3",
```
howardchung: This should be a devDependency?

package.json
```diff
@@ -33,6 +33,12 @@
     "@material-ui/core": "^4.5.0",
     "@material-ui/icons": "^4.5.1",
     "@material-ui/styles": "^4.5.0",
+    "@types/jest": "^24.0.23",
```
howardchung: all the types should be devdependencies?

src/components/Header/Header.jsx
```diff
@@ -116,7 +116,7 @@ const SettingsGroup = ({ children }) => {
   const [anchorEl, setAnchorEl] = useState(null);
   const handleClose = useCallback(() => {
     setAnchorEl(undefined);
-  }, [anchorEl]);
+  }, []);
```
howardchung: What does this change?

bdbch: Removes the dependency of anchorEl which was wrong.

But I think we need `setAnchorEl` as a dependency since we use it in the callback function.

NoPhaseNoKill: Sorry I don't have that much experience with hooks. It sounds like from reading the docs/your answer, that I need to pass through setAnchorEl to the dependency list in that case? So the original one must have been a typo. Am I correct?

NoPhaseNoKill: Once that's verified I can push the change

NoPhaseNoKill: Actually sorry I've misinterpretted the docs. The code looked funtionality correct before - as we're waiting for any changes to anchorEl done through the callback function which is passed to the onClose of the dropdown menu. I will undo these changes and put in a linter ignore for the time being until someone with a bit more experience can help

bdbch: `[setAnchorEl]` should be fine since it's the only dependency we require in the callback function itself

NoPhaseNoKill: Ah yep ok now I get what you mean. Apologies, I'm relatively new to hooks and am still learning them. Appreciate the help :), pushing fix for the linting error now

bdbch: It's all cool, hooks are a pretty new concept. I'm also still learning. :)

package.json
```diff
@@ -33,6 +33,8 @@
     "@material-ui/core": "^4.5.0",
     "@material-ui/icons": "^4.5.1",
     "@material-ui/styles": "^4.5.0",
+    "@typescript-eslint/eslint-plugin": "^2.12.0",
```
howardchung: I'm pretty sure these can go in devDependencies also

src/components/Heroes/TestTypescriptReact.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const TestTypescriptReact = () => (
+  <div>Test</div>
+);
+
+export default TestTypescriptReact;
```
102: Should we remove this file?

src/components/Heroes/TestTypescript.ts
```diff
@@ -0,0 +1,2 @@
+export const TestTypescript = () => 0;
```
102: Should we remove this file?

src/react-app-env.d.ts
```diff
@@ -0,0 +1,2 @@
+// / <reference types="react-scripts" />
```
102: Can we remove it?

https://github.com/odota/web/pull/2211#discussion_r331854185

src/components/Heroes/TestTypescriptReact.tsx
NoPhaseNoKill: Yes was just testing all of the linting with this.

src/components/Heroes/TestTypescript.ts
NoPhaseNoKill: See above

src/react-app-env.d.ts
NoPhaseNoKill: We require this file, but we should probably .gitignore it

NoPhaseNoKill:
Removed test files, added the .d.ts to gitignore, updated the file with a better explanation of why we need it, and added a types directory at the top level. This should be good to go when you are now

102:
@NoPhaseNoKill do you have an idea or plan about the TS integration following steps after this PR being merged? Am I correct, that the only way we can do it is just translating `index.js` to `index.ts` & go down the file tree step by step?

NoPhaseNoKill:
Traditionally the way I have done migrations to typescript, is doing small pieces at a time. If you find yourself working in a JS/JSX file, the emphasis is on the dev to a) convert it to a typescript file, or b) come back to it later, accept we accrue another piece of 'tech debt' and convert when they get to it. 

Generally I find converting files to typescript help prevent existing bugs, give you a better idea of what the code is doing, and help frame/map a solution in your mind. You're almost always going to get benefits from the conversion. 

However, if it's literally a one-liner change, it might make sense to keep it as JS, and come back to the file later. I don't want to throw any common sense out the door haha. Generally if you have the confidence to update a single line in a file, knowing nothing 'should' break - you've probably also got context around the types, and should at some stage be converting it to typescript.

There's no need to start at the top of the tree, we should be using an adopt as we go strategy so to lessen the probability of breaking parts of the app. General rule of thumb I use: create 'inline types', and if you find yourself exporting these types (as the same type is used across another file) - these should be moved to the base types folder.

More than likely the initial types (the ones aforementioned) will slow us down initially, but as they grow organically, we can leverage existing types and this will significantly increase contribution ability.

PR #2381, opened 2019-12-26
Ejected from create-react-app due to tsconfig being overriden

NoPhaseNoKill:
**Problems**
1) Absolute/relative imports not working occasionally for typescript files.
2) Generics (typescript) linting problem
3) Warnings in linting were being ignored, and app was compiling successfully (even though warning were set to fail)
4) Types could not be used from the types directory

**Fixes**
1 & 4) Ejected from create-react-app due to tsconfig being overridden, and no way to stop this behaviour. Have added a 'baseUrl' and 'include' to tsconfig.
2) Upgraded eslint-config-airbnb to later version - as old version was causing a generics linting error for typescript.
3) Temporary fix of eslint ignoring these, as there's a lot of changes already in this.

**Questions:**

1) I'm not 100% certain whether the newly added production deps all need to be there? This was done automatically via the eject command that came with CRA. Could someone please advise?
2) Seeing as a fair amount has changed, could someone with a little bit more knowledge of the app test this?
3) I'm thinking of starting the linting options from scratch (not using the airbnb-recommended ones). I've found these extremely unhelpful, and each time we upgrade, the linting rules will change. Do we have anywhere that specifies the common rules we want? If not, I can create a new list, as I believe it's going to be much better in the long run to be explicit about the style we want, not implicitly based off of the version of eslint-airbnb we're using.

**Further commits**
1) Will be merging with master when this is ready to go

**Currently working on**
1) Upgrade Heroes/index.jsx to typescript -> should be out in the next few days, so may include in this PR depending on when you get to it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10817,7 +10817,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
           "integrity": "sha512-tQtKA9WIAhBF3+VLAseyMqZeBjW0AHJoxOtYqSUZNJxauErmLbVm2FW1y+J/YA9dUrAC39ITejlZWhVIwawkKw==",
           "dev": true,
           "requires": {
@@ -11362,7 +11362,7 @@
         },
         "chalk": {
           "version": "1.1.3",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
+          "resolved": "http://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
           "integrity": "sha1-qBFcVeSnAv5NFQq9OHKCKn4J/Jg=",
           "requires": {
             "ansi-styles": "^2.2.1",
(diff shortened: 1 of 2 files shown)


howardchung:
I'd like to avoid ejecting if possible.

Absolute/relative imports: I thought we switched to using relative imports everywhere a while back, but I could be mistaken.

Types directory: I assume we can't be the only ones with this problem. Is there an issue for this in the CRA repo? Does it work if we just have a types.d.ts file somewhere that CRA is preconfigured to pick up types from?

bdbch:
I'd also go against ejecting. There should be a simple way to setup Typescript with a CRA application. They even have their own typescript template so it should work fine manually too.

NoPhaseNoKill:
100% agree that we should be avoiding ejecting if possible, but I struggled finding a way to resolve this. I think what needs to happen is we need to update the CRA version? And then we can use the below solutions - but I'm not sure how we do this.

Is it just updating the react-scripts folder?

The issues I could find were:
https://github.com/facebook/create-react-app/issues/5692
https://github.com/facebook/create-react-app/issues/5645
https://github.com/facebook/create-react-app/issues/5118

NoPhaseNoKill:
Will be creating a new pull request, and I've decided to discard the eject changes so I can show you two the problem and see if we can come up with a better solution

NoPhaseNoKill:
Please see new pull request: https://github.com/odota/web/pull/2382

PR #2382, opened 2019-12-27
Typescript fixes

NoPhaseNoKill:
**Problems**

1) Relative imports not working correctly in IDE.
2) Generics (typescript) linting problem
3) Types could not be used from the types directory

**Fixes**
1) Still to be found, need some help for this.
2) Bumped eslint version, had to then set extra definitions so we have same linting as before
3) Fixed when IDE imports correctly, or path is manually typed


src/types/TestImporter.ts
```diff
@@ -0,0 +1,22 @@
+import { MatchCategory } from './Match/MatchCategory';
+import { Match } from './Match/Match';
+
```
NoPhaseNoKill: When using IDE completion, the below is what it automatically generates in IntelliJ. Currently I have to manually type it - which through investigation is the root cause of this issue. Any ideas?

src/types/TestImporter.ts
```diff
@@ -0,0 +1,22 @@
...
+//  Below is what my IDE auto generates for import
+// import { MatchCategory } from 'src/types/Match/MatchCategory'
+// import { Match } from 'src/types/Match/Match'
+
```
NoPhaseNoKill: When using IDE completion, the above is what it automatically generates in IntelliJ. Currently I have to manually type it (as shown in lines 1/2) - which through investigation is the root cause of this issue. Any ideas?

NoPhaseNoKill: https://stackoverflow.com/questions/55198502/using-eslint-with-typescript-unable-to-resolve-path-to-module Is what is meant to fix it, but it doesn't appear to be working for me

NoPhaseNoKill: Errors are as below:

Typescript compile error: 'Cannot find module for 'src/types/Match....'
ESLint: 'Unable to resolve path to module 'src/types/Match....'

NoPhaseNoKill:
Linter has been brought up to date, with all of the new linting rules that were added due to eslint upgrade. Once the import/export issue is resolved, this should be good to go

src/types/TestImporter.ts
NoPhaseNoKill: Related CRA issues:
facebook/create-react-app#5692
facebook/create-react-app#5645
facebook/create-react-app#5118

NoPhaseNoKill:
Have now fixed the TSCompiler error for imports. Remaining issue appears to be ESlint related. Cannot for the life of me fix this. Could someone please assist?

howardchung:
Looks like most of these changes are due to adding parens on the function arguments on arrow functions. Did we want to make this change as part of the PR (since it's kind of unrelated)?

howardchung:
Also, what is the remaining issue you're referencing? Is it that ESLint is failing to pick up files that are absolutely imported?

NoPhaseNoKill:
> Looks like most of these changes are due to adding parens on the function arguments on arrow functions. Did we want to make this change as part of the PR (since it's kind of unrelated)?

Yes. It's related to this point:
Generics (typescript) linting problem (upgrading eslint brought in new rules)

howardchung:
In that case, I'd disable that lint rule in this change so the actual changes are easier to review.

NoPhaseNoKill:
> Also, what is the remaining issue you're referencing? Is it that ESLint is failing to pick up files that are absolutely imported?

Remaining issue that when trying to import like:
`import { Match } from 'src/types/Match/Match'`

It's throwing an error that says:
`'Unable to resolve path to module 'src/types/Match/Match'.(import/no-unresolved)`

howardchung:
I'd argue that we could just disable this ESLint rule, since TypeScript will already fail the compile if we have an unresolvable file.

howardchung:
But it looks like this might be the fix for ESLint:
https://stackoverflow.com/questions/50234858/how-to-change-eslint-settings-to-understand-absolute-import

If we don't need it though, I'd just disable the rule and avoid the extra dependency.

src/components/Announce/index.jsx
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line no-redeclare
```
102: Why do we need to disable `no-redeclare` rule here?

src/components/Collapsible/index.jsx
```diff
@@ -35,10 +35,9 @@ const CollapseButton = ({
   handleClick, collapsed, strings, buttonStyle, handleHoverOn, handleHoverOff,
 }) => (
   <ButtonContainer style={buttonStyle} onClick={handleClick} onMouseEnter={handleHoverOn} onMouseLeave={handleHoverOff}>
-    {collapsed ?
-    [<span>{strings.general_show}</span>, <IconPlusSquare />]
-    :
-    [<span>{strings.general_hide}</span>, <IconMinusSquare />]}
+    {collapsed
+      ? [<span>{strings.general_show}</span>, <IconPlusSquare />]
+      : [<span>{strings.general_hide}</span>, <IconMinusSquare />]}
```
102: **This is absolutely out of scope of this pr**, but I think we should add explicit keys to the elements in array 

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `App`. See https://fb.me/react-warning-keys for more information.
    in p (at src/index.js:9)
    in App (at src/index.js:17)
```

src/components/Announce/index.jsx
NoPhaseNoKill: Honestly I was trying to just push through the linting errors, as it's taken quite a bit longer to produce this minimal change/required functionality of typescript

NoPhaseNoKill:
Commit: Removed arrow parens lint rule, and no-unresolved should have been named 'Added arrow parens lint rule, and no-unresolved'

NoPhaseNoKill:
I apologise for the number of commits. I have now disabled the arrow-parens linting rule, reverted all changes made by it, added the no-unresolved linting rules for TS. Once this builds, this should be good to go

NoPhaseNoKill:
Think I finally got there. Botched the revert, but we should be good to go now

src/components/Collapsible/index.jsx
NoPhaseNoKill: Completely agree. These warnings have been driving me insane

NoPhaseNoKill:
Also if you need to make a clearer PR, let me know and you can close this one off. I've now got a good idea of what needs to be done for a clean commit

src/components/Announce/index.jsx
howardchung: we should probably just remove that `/* global localStorage */` and replace with window.localStorage

Can save for another PR

howardchung:
I can squash and commit

PR #2384, opened 2019-12-27
Report unused ESLint disable directives

102:
A few versions ago, [this useful option](https://eslint.org/docs/user-guide/command-line-interface#report-unused-disable-directives) was added to ESLint


howardchung:
Possibly worth adding a command to package.json?

howardchung:
Actually, it looks like this is causing the master build to have some lint errors. Going to revert for now.

102:
@howardchung in the github branch protection settings, you can set the following:

> Require branches to be up to date before merging
> This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).

This will likely prevent this kind of problems in the future

PR #2389, opened 2019-12-29
Converted columns for HeroPage to Typescript file

NoPhaseNoKill:
Convert columns.js for Heroes page to Typescript file


.eslintrc.js
```diff
@@ -103,7 +103,12 @@ module.exports = {
         "import/prefer-default-export": "off",
         "react/state-in-constructor": "off",
         "import/extensions": "off",
-        "import/no-unresolved": "off"
+        "import/no-unresolved": "off",
+        "semi": "off",
+        "arrow-body-style": "off",
+        "@typescript-eslint/member-delimiter-style": "off",
+        "@typescript-eslint/no-use-before-define": "off",
+        "import/order": "off",
```
102: why do we need to disable all these rules?

src/components/Heroes/RankColumns.ts
```diff
@@ -0,0 +1,271 @@
+import heroes from 'dotaconstants/build/heroes.json';
+import constants from 'src/components/constants'
+import { abbreviateNumber, displayHeroId } from '../../utility';
+import { GlobalStrings } from 'src/types/common/GlobalStrings'
+import { Heroes } from 'src/types/Hero/Heroes'
```
102: what is the reason to mix relative imports with absolute imports?

src/components/Heroes/RankColumns.ts
```diff
@@ -0,0 +1,271 @@
...
+
+type Props = {
+  tabType: HeroesTabType
+  strings: GlobalStrings
+}
+
+enum HeroesTabType {
```
102: can we avoid adding `Type` suffix here?

src/components/Heroes/RankColumns.ts
```diff
@@ -0,0 +1,271 @@
...
+  percentBarsWithValue: (row: Row) => string
+  displayIcon: string
+  colColor: string
+}
+
+type HeroColumn = {
+  displayName: string,
+  tooltip: string,
+  field: string
+  displayFn: () => JSX.Element,
```
102: Would this work? 

```suggestion
  displayFn: () => React.ReactNode,
```

src/components/Heroes/RankColumns.ts
```diff
@@ -0,0 +1,271 @@
...
+  paddingLeft: number,
+  paddingRight: number,
+  tooltip: string
+}
+
+type Row = {
+  [key in string]: number
+}
+
+type ColumnTypes = ProColumn | PublicColumn | HeroColumn
```
102: You usually do not need to add `Type` suffix to type (and types are also usually in singular)

```suggestion
type Column = ProColumn | PublicColumn | HeroColumn
```

src/components/Heroes/RankColumns.ts
```diff
@@ -0,0 +1,271 @@
...
+  displayName: string,
+  tooltip: string,
+  field: string
+  displayFn: () => JSX.Element,
+  sortFn: (row: Row) => string
+}
+
+type PreparedColumn = (ProColumn | PublicColumn | HeroColumn) & ColumnAdditions
+
+type ColumnAdditions = {
```
102:
```suggestion
type ColumnAddition = {
```

src/components/Heroes/RankColumns.ts
```diff
@@ -0,0 +1,271 @@
...
+
+type HeroColumn = {
+  displayName: string,
+  tooltip: string,
+  field: string
+  displayFn: () => JSX.Element,
+  sortFn: (row: Row) => string
+}
+
+type PreparedColumn = (ProColumn | PublicColumn | HeroColumn) & ColumnAdditions
```
102: This looks like a good use case for typescript interfaces, you can declare `Column` interface with all the things that are common & extend it for specific cases

src/components/Heroes/index.jsx
```diff
@@ -12,7 +12,7 @@ import {
   sum,
   abbreviateNumber,
 } from '../../utility';
-import columns from './columns';
+import { RankColumns } from './RankColumns.ts';
```
102: it's kind of confusing to use function that has name in `PascalCase` as plain function, they're usually used as components in react apps. Does it make sense either to rename this fn to `camelCase` or use it as component in this file?

.eslintrc.js
NoPhaseNoKill: In my opinion they offer nothing except friction for particular 'styles' of coding. In fact, some of the rules (arrow-body-style, no-use-before-define) promote less maintainable/unreadable code which I think we should avoid

src/components/Heroes/RankColumns.ts
NoPhaseNoKill: There was no reason, it was simply copied from the old file

src/components/Heroes/RankColumns.ts
NoPhaseNoKill: Yes, I was struggling for a name for this. Suffixing with type is a bad habit I've gotten into for tabs and columns

src/components/Heroes/RankColumns.ts
NoPhaseNoKill: Yes. Is there a reason you prefer one over the other? I don't actually know the difference

src/components/Heroes/RankColumns.ts
NoPhaseNoKill: Agreed on the singular, was just the particular use-case of the function where the type was being used made me consider doing it this way. You're right though, looks cleaner like that

src/components/Heroes/RankColumns.ts
NoPhaseNoKill: Type and interface are exactly the same thing except for syntax. The above is the exact same as had you defined it an interface and extended ProColumn with ColumnAddition

src/components/Heroes/index.jsx
NoPhaseNoKill: Yep really good point. camelCasing much better idea. Appreciate all the feedback btw

src/components/Heroes/RankColumns.ts
102: NOTE: I actually only have flow background, typescript is a new thing for me.

This [stackoverflow answer](https://stackoverflow.com/a/58123882) seems to have a good explanation about the difference.

tl;dr
1. `JSX.Element` is basically `React.ReactElement<any, any>` within react context.
1. `React.ReactNode` includes this, `string`, `null`, `boolean`, and all other things that can be rendered by react component.
1. It doesn't look like this particular `displayFn` requires child to be exactly an element.

None of these points are strong reason to use `React.ReactElement` or `React.ReactNode` over `JSX.Element`, but in my (flow-biased) opinion, explicit type for react-related item will be a better option

src/components/Heroes/index.jsx
102: Btw, KCD has recently published [a post](https://kentcdodds.com/blog/dont-call-a-react-function-component) about the problems that can occur with this

src/components/Heroes/RankColumns.ts
NoPhaseNoKill: Well there you go; TIL. Thanks for that, makes sense

src/components/Heroes/index.jsx
NoPhaseNoKill: Sorry I don't quite understand what you mean. This has no internal state that it manages, so the above article doesn't seem appropriate.

If, however, you're talking about the relationship between expecting it to be a component when PascalCasing it, as opposed to it being a raw function when camelCased, then yes I absolutely agree.

102: > If, however, you're talking about the relationship between expecting it to be a component when PascalCasing it, as opposed to it being a raw function when camelCased, then yes I absolutely agree.

that's the main point, yes (and changing the name to match the convention was and still the only thing I have suggested 😄)

> This has no internal state that it manages, so the above article doesn't seem appropriate.

it does not have internal state yet, but it can be later added by any other developer that could be unaware about this component being used as a plain fn somewhere in the code

NoPhaseNoKill: Agreed. Thanks for the suggestions, have implemented them all

PR #2413, opened 2020-01-25
fix(utility): fix formatGraphValueData to handle 0

chao7150:
Tooltips of histogram do not show x-value when it is 0.
It is because `!data` mistakenly judges 0 as invalid data and returns an empty string.
I fixed this.

image

image


howardchung:
Thanks!

PR #2436, opened 2020-02-09
Translated to Hungarian

amatyas001:
Hey Guys! 
In the last few days I've managed to translate the application into Hungarian. I've tried to do my best to accomplish this. Hope you enjoy :)


amatyas001:
Some of the CS texts were missing from the hu_HU so I've added them. Compared to the current version of en_US seems to be ok.

howardchung:
Awesome, thanks! I think we might have to do something on crowdin if we want to allow people to submit updates online. @albertcui should be able to handle that.

PR #2446, opened 2020-02-16
Refactored App component using Hooks

amatyas001:
Hi there! 
I've done a small refactoring on the App component. Now it uses React Hooks in functional component. It has better performance and code readability. I also moved related content into the Header component so everything in their place now. 
I planed to write unit tests but I saw that there's no Enzyme or such stuff in packages. Any reason for using only expensive e2e tests? :) tdd will turn up manageability a lot! I would recommend adding `enzyme`, `enzyme-adapter-react-16` and `react-testing-library` to devdeps. Or am I missing something? (It's late here now...)
Anyway... Have a nice day and thanks!


howardchung:
Open to adding tests. The reason we don't have unit tests is that the e2e tests cover the most important use cases (does the page crash) without too much effort.

PR #2449, opened 2020-02-16
Fixed initial warnings caused by invalid propTypes and array indices

amatyas001:
Hey there!
I fixed some propTypes definition. Not all of them, it's still work in progress. Also fixed array map functions using indices as keys. Read more: [Why using an index as Key in React is probably a bad idea?](https://medium.com/@vraa/why-using-an-index-as-key-in-react-is-probably-a-bad-idea-7543de68b17c).

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -109,24 +109,30 @@ const LinkGroup = ({ navbarPages }) => (
 );
 
 LinkGroup.propTypes = {
-  navbarPages: PropTypes.shape([{}]),
+  navbarPages: PropTypes.arrayOf(
+    PropTypes.shape({
+      key: PropTypes.string,
+      to: PropTypes.string,
+      label: PropTypes.string,
+    }),
+  ),
 };
 
 const SettingsGroup = ({ children }) => {
   const [anchorEl, setAnchorEl] = useState(null);
   const handleClose = useCallback(() => {
-    setAnchorEl(undefined);
+    setAnchorEl(null);
   }, [setAnchorEl]);
 
   return (
-    <>
+    <div>
       <IconButton color="inherit" onClick={e => setAnchorEl(e.currentTarget)}>
         <Settings />
       </IconButton>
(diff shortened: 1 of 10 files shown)


src/components/Header/Header.jsx
```diff
@@ -303,15 +309,15 @@ const Header = ({
                       <ListItemText primary={strings.app_my_profile} />
                     </ListItem>
                   </DrawerLink>
-                  <DrawerLink as="a" href={`${process.env.REACT_APP_API_HOST}/logout`}>
+                  <DrawerLink as="a" to={`${process.env.REACT_APP_API_HOST}/logout`}>
```
howardchung: I'm pretty sure this needs to be href since the logout is not part of the app

src/components/Header/Header.jsx
```diff
@@ -109,24 +109,30 @@ const LinkGroup = ({ navbarPages }) => (
...
 const SettingsGroup = ({ children }) => {
   const [anchorEl, setAnchorEl] = useState(null);
   const handleClose = useCallback(() => {
-    setAnchorEl(undefined);
+    setAnchorEl(null);
   }, [setAnchorEl]);
 
   return (
-    <>
+    <div>
```
howardchung: what's the purpose of this change?

src/index.js
```diff
@@ -40,7 +40,7 @@ const app = (
   </Provider>
 );
 
-if (rootElement.hasChildNodes()) {
+if (!rootElement.hasChildNodes()) {
```
howardchung: why this change?

src/components/Header/Header.jsx
amatyas001: It's related to the ref forwarding in MUI. It doesn't work with fragments but should be with divs. It's still drops warning so... just a try.

src/components/Header/Header.jsx
amatyas001: It works well with the login button in local env.  Thought it will work here too.

src/index.js
amatyas001: It throws a warning about expected divs. Is SSR enabled here or why the hydrate?

PR #2498, opened 2020-04-02
Update transformMatch.js

leoffx:
Added condition to check if player 'ability_upgrades_arr' isn't null.
Ability draft matches where a player hasn't upgraded any skills aren't loading. 
i.e. https://www.opendota.com/matches/5310636074


howardchung:
lint error:
`  228:110  error  ["ability_upgrades_arr"] is better written in dot notation  dot-notation`

PR #2517, opened 2020-04-06
strip out html content from item descriptions in item tooltips

### LINKED ISSUE #2507: broom handle does font color.
hi,
screen02

bdbch:
This PR fixes #2507 as it strips out HTML content from item tooltip attributes.

Still I think this should also happen on the server as the content produced by the server is not "clean".


howardchung:
Yes, I think we should do this in dotaconstants rather than at runtime.

PR #2533, opened 2020-04-18
Updated heatmap value scaling

TheGuywithTheHat:
For a while now I've noticed that wardmaps often look weird, frequently being giant red blobs with no detailed information. The issue was that 25 was getting added to all nonzero values in the map, but this wasn't taken into account when calculating the max value for heatmap scaling. This could cause the heatmap's max value to be under 25, and thus _all_ data points get rendered as above the max.

I updated the calculations to no longer have that shift of 25. I also changed the scaling to be sqrt, rather than a hard cutoff at 25x the average value. Before/after comparisons (click for larger images):

~10 matches
[](https://i.imgur.com/VECM53i.png) [](https://i.imgur.com/e7wjDBb.png)

~100 matches
[](https://i.imgur.com/7vK5KLq.png) [](https://i.imgur.com/vTTsWmz.png)

~2000 matches (note increased detail in areas such as north of roshpit)
[](https://i.imgur.com/ndFtpY1.png) [](https://i.imgur.com/qLh1t1Q.png)

~6000 matches (different player)
[](https://i.imgur.com/nul7vZL.png) [](https://i.imgur.com/TkyUkyI.png)

### DIFF
--- a/src/components/Heatmap/Heatmap.jsx
+++ b/src/components/Heatmap/Heatmap.jsx
@@ -7,19 +7,15 @@ import DotaMap from '../DotaMap';
 /**
  * Adjust each x/y coordinate by the provided scale factor.
  * If max is provided, use that, otherwise, use local max of data.
- * Shift all values by the provided shift.
  * Returns the adjusted heatmap data.
  */
-function scaleAndExtrema(points, scalef, max, shift) {
-  // the max values should not deviate from the average by more than a factor of 25
-  const maxValue = (points.reduce((a, b) => a + b.value, 0) / points.length) * 25;
-
+function scaleAndExtrema(points, scalef, max) {
   const newPoints = points.map(p => ({
     x: Math.floor(p.x * scalef),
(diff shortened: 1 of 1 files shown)


howardchung:
I believe this was originally done to improve the appearance of individual match maps (where the ward counts are much lower).

Could you check to see how one of those looks?

howardchung:
IIRC we had issues where if a particular cell had 2 wards, it would show up much brighter than all the other spots, so the shift of 25 was meant to reduce the difference between those points. It's possible that we might want to use a dynamic shift based on the local max

TheGuywithTheHat:
As far as I'm aware, right now heatmaps aren't used for match vision maps, right?

I found an example of a player's heatmap with some single wards and one stack of two.

Current:


My PR:


What it could look like with a shift of 3:


What do you think it should look like ideally?

howardchung:
Oh yeah, we switched the match pages to use the new vision display.

The PR map looks fine to me.

PR #2568, opened 2020-05-19
Fix missing icons in Dota Map

### LINKED ISSUE #2566: Missing void spirit icon in minimap
Void spirit doesn't show up in the minimap due to missing icon. https://www.opendota.com/matches/5412330160

I think I uploaded a copy to fix this for https://github.com/odota/web/issues/2323, but minimap probably needs to use it as well.

jgalat:
Hi!

This fixes #2566. It also fixes the Snapfire icon which was also missing from the minimap.

Cheers!


howardchung:
Thanks! I remembered uploading custom icons before, but didn't go back and check all the possible use sites.

PR #2581, opened 2020-05-30
Update captains mode draft page for patch 7.25

zeroclouds:
Bans changed from 3/2/1 to 4/1/1.
Picks remained 2/2/1 but their turn number changed.


src/components/Match/Draft/Draft.jsx
```diff
@@ -186,14 +186,20 @@ const Draft = ({
...
   let orderTwo = [];
+  let picks = [];
   if (startTime > 1525910400) { // post 7.15
     orderOne = [1, 3, 5, 7, 10, 11, 13, 16, 18, 20, 21];
     orderTwo = [2, 4, 6, 8, 9, 12, 14, 15, 17, 19, 22];
-  } else {
-    orderOne = [1, 3, 5, 7, 10, 12, 14, 16, 18, 20, 21];
-    orderTwo = [2, 4, 6, 8, 9, 11, 13, 15, 17, 19, 22];
+    picks = [7, 8, 9, 10, 15, 16, 17, 18, 21, 22];
+  } if (startTime > 1584403200) { // post 7.25
```
howardchung, line 194: shouldn't this block go first and  use else if on the 7.15 one?

zeroclouds: Yes it should. Thank you.

src/components/Match/Draft/Draft.jsx
```diff
@@ -186,14 +186,20 @@ const Draft = ({
   // one-based indexing (since draft[i].order starts at 1)
   let orderOne = [];
   let orderTwo = [];
-  if (startTime > 1525910400) { // post 7.15
+  let picks = [];
+  if (startTime > 1584403200) { // post 7.25
+    orderOne = [1, 3, 5, 7, 9, 12, 12, 14, 16, 18, 20, 21];
```
howardchung, line 191: Is 12 supposed to appear twice?

zeroclouds: It should not.

PR #2592, opened 2020-06-02
Add new banner

albertcui:


src/components/App/App.jsx
```diff
@@ -150,7 +150,12 @@ const App = (props) => {
         <Helmet
           defaultTitle={strings.title_default}
           titleTemplate={strings.title_template}
-        />
+        >
```
howardchung, line 153: isn't this the top?

howardchung:
Should also probably update https://github.com/odota/web/blob/master/.env

Netlify preview should also let you see if it worked: https://deploy-preview-2592--opendota-staging.netlify.app/

src/components/App/App.jsx
albertcui: Yes, we have to add a script to the header and to where we want it to render.

albertcui:
> Should also probably update https://github.com/odota/web/blob/master/.env
> 
> Netlify preview should also let you see if it worked: https://deploy-preview-2592--opendota-staging.netlify.app/

We need to get verified before anything actually will work, so we basically need to get this deployed and wait a few days...

howardchung:
OK, do you remember how to deploy?

PR #2610, opened 2020-06-18
fix: hint interpolation via regex

### LINKED ISSUE #2256: Tome of Experience string isn't interpolated
**Current Behavior**
When hovering over 'tome of experience' in a match, the following is shown:
image

The description is gotten from https://github.com/odota/dotaconstants/blob/master/build/items.json

and then rendered here:
https://github.com/odota/web/blob/master/src/components/ItemTooltip/index.jsx

at line 232. This looks like it should be an easy fix,  but I don't see where the "%customval_team_tomes_used% is meant to be interpolated

**Expected behavior/code**
The description should interpolate "%customval_team_tomes_used%" for the actual value (other items possibly affected by this too)

ZoraizQ:
Tomes Used By Team: %%, Ability Range: %%, all such phrases inside
the Item Hint will be removed from the Item Tooltip using a replace
with regex.

Closes #2256

### DIFF
--- a/src/components/ItemTooltip/index.jsx
+++ b/src/components/ItemTooltip/index.jsx
@@ -235,7 +235,7 @@ const ItemTooltip = ({ item, inflictor }) => (
       }
       return null;
     })}
-    {item.hint && <Hint>{item.hint}</Hint>}
+    {item.hint && <Hint>{item.hint.map(hintLine => hintLine.replace(/[ a-zA-Z]+: %\w+%/g,""))}</Hint>}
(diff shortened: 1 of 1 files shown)


howardchung:
This kind of fix should go in the odota/dotaconstants repo script, so we don't execute it at runtime

ZoraizQ:
Referenced the related issue #78 on odota/dotaconstants. For the most special case, it removes the "Team Tomes Used: %customval_team_tomes_used%" line as it is out of context and is required in-game and not on the Item ToolTip.

PR #2612, opened 2020-06-20
Enable Fade-in when choosing hero in Combo page 

jzhao62:
Enable Fade-in when choosing heros into teamA teamB in combo page. 
- try to enable fade out, but it requres refactor the deselectHero function
-fix some linting erros

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5099,7 +5099,7 @@
         },
         "minimist": {
           "version": "1.2.0",
-          "resolved": "http://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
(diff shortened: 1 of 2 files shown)


howardchung:
lgtm

Argn0:
I think the transition time should be shorter. 1 second feels too slow.

jzhao62:
yeah, 1 sec is a little bit too slow, I will modify that with my next PR

howardchung:
@jzhao62 do you just want to update this one?

jzhao62:
I made fade-out partially works, could be on a different PR, or if you want to wait

Argn0:
We can wait, there is no rush.

howardchung:
I think if you want to make that change it should just be on this PR rather than opening another

PR #2624, opened 2020-07-02
a fix for negative stun duration

iamBevan:


howardchung:
Should probably do this in the parsing/core code so the API data is 0 as well

PR #2657, opened 2020-09-05
Added magnifying glass icon to the StyledHorizontalMenuFilter 

MaksymKarunos:
- Added a new icon of Magnifying Glass to the src/components/Icons (the icon is scraped from the original search icon located in the header, for some reasons it was not in the folder with all .svg icons).
- Updated StyledHorizontalMenuFilter with the icon so it can be easily reused.

### DIFF
--- a/src/components/Icons/MagnyfingGlass.jsx
+++ b/src/components/Icons/MagnyfingGlass.jsx
@@ -3,16 +3,15 @@ import styled from 'styled-components';
 
 const icon = props => (
  // Taken from main page
- <svg {...props} viewBox="0 0 24 24">
-   <path
-     fill="#d3d3d3 "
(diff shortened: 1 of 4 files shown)


Argn0 (changes requested):
I'm pretty sure there is a magnifying glass graphic somewhere already. It's been used for the submit button on the Explorer page for example.

howardchung:
Yeah looks like you can use the ActionSearch built-in icon here:
https://github.com/odota/web/blob/e7db47f37d1d07af7e43bb3852b3a29f5d0d52c5/src/components/Explorer/index.jsx#L267

MaksymKarunos:
> Yeah looks like you can use the ActionSearch built-in icon here:
> https://github.com/odota/web/blob/e7db47f37d1d07af7e43bb3852b3a29f5d0d52c5/src/components/Explorer/index.jsx#L267

Thank you so much! I am sorry I am still getting used to the project!
Making changes now!

Issue #2663, opened 2020-09-07
Turbo mode button


howardchung:
Add a button to player profiles that makes all player stats API requests send the turbo parameters (significant=0 and game_mode=23). Click to toggle on/off.

Store it in localStorage so it's persistent.

I only play turbo now, so this would be nice.

ohnickmoy:
Hey @howardchung. I'm interested in contributing, but I'm also new to open source and I'm not sure if this akin to a good-first-issue.

howardchung:
You can give it a shot, I don't think it's overly complicated. I'm available to answer questions if you've got them.

bdbch:
I started to work on this on the branch `turbomode-stats`, see here:
https://github.com/odota/web/tree/turbomode-stats

bdbch:
@howardchung do you only want matches to be filtered by turbomode? Or do you also want to update your profile stats? If so - does the query for stats allow the game_mode differentiation?

howardchung:
I think we should be able to pass through the turbo params to all the /players endpoints

bdbch:
Alright, I'll see if I can easily get this into all requests.

Just a question - for the overview we already have the filter for gamemode. I guess this should overwrite the global toggle?

bdbch:
@howardchung I got it but it seems the recentMatches query ignores the gamemode param.
Could you maybe check my PR? I can't find out why the stats/summary is not updated in combination with the game_mode.

PR #2674, opened 2020-10-07
WIP: add turbomode toggle to player profile

bdbch:
This PR implements a reusable gamemode toggle that is synced with localStorage and allows the user to switch between turbomode and default mode globally.

To access the global gamemode state easier I refactored the RequestLayer component for Matches into a functional component and accessing the context via a hook (`useGamemode`) which just requests data from the provider.

I wasn't sure about tracking the gamemode, right now I'm using strings (`default` and `turbo`), should we rather use a enum for that with the gamemode number?

See #2663

### DIFF
--- a/src/components/Player/Pages/Matches/Matches.jsx
+++ b/src/components/Player/Pages/Matches/Matches.jsx
@@ -1,19 +1,20 @@
-import React from 'react';
 import PropTypes from 'prop-types';
+import React from 'react';
 import { connect } from 'react-redux';
+
 import { getPlayerMatches } from '../../../../actions';
-import Table from '../../../Table';
+import { useGamemode } from '../../../../context/GamemodeContext';
 import Container from '../../../Container';
+import Table from '../../../Table';
 import playerMatchesColumns from './playerMatchesColumns';
 
-const Matches = ({
-  data,
-  error,
-  loading,
-  strings,
-}) => (
+const Matches = ({ data, error, loading, strings }) => (
   <Container title={strings.heading_matches} error={error} loading={loading}>
-    <Table paginated columns={playerMatchesColumns(strings, true)} data={data} />
+    <Table
+      paginated
+      columns={playerMatchesColumns(strings, true)}
+      data={data}
+    />
   </Container>
 );
 Matches.propTypes = {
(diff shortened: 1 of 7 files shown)


bdbch:
@howardchung I'll resolve the issues with this PR this evening, but maybe you could check it out? I can't get the player stats to update with the turbomode toggle.

src/components/Player/Pages/Matches/Matches.jsx
```diff
@@ -23,47 +24,47 @@ Matches.propTypes = {
...
-    playerId: PropTypes.string,
-    strings: PropTypes.shape({}),
-  }
-
-  componentDidMount() {
-    getData(this.props);
+const getData = (props, gamemode) => {
+  let params = props.location.search;
+  if (gamemode === 'turbo') {
+    params = `${props.location.search}&game_mode=23`;
```
howardchung, line 30: you'll probably want to apply the game_mode filter in the actions file, so it can be applied to all of the possible /players calls. You also need significant=0

src/components/Player/Pages/Matches/Matches.jsx
```diff
@@ -23,47 +24,47 @@ Matches.propTypes = {
...
   }
+  props.getPlayerMatches(props.playerId, params);
+};
 
-  componentDidUpdate(prevProps) {
-    if (this.props.playerId !== prevProps.playerId || this.props.location.key !== prevProps.location.key) {
-      getData(this.props);
+const RequestLayer = (props) => {
+  const gamemodeState = useGamemode();
```
howardchung, line 36: don't think this is necessary, if you move the logic to the actions file and just check localstorage there. You may not be able to use Context there as that's not in the React tree.

src/components/Player/Pages/Overview/Overview.jsx
```diff
@@ -238,57 +270,46 @@ Overview.propTypes = {
...
+    props.location.search &&
+    props.location.search.indexOf('game_mode') !== -1;
+  if (gamemode === 'turbo' && !hasGameMode) {
+    params = `${props.location.search}&significant=0&game_mode=23`;
+  }
 
-const getData = (props) => {
   if (props.location.search) {
-    props.getPlayerMatches(props.playerId, props.location.search);
+    props.getPlayerMatches(props.playerId, params);
```
howardchung, line 284: similarly, i'd delegate the switch to the actions file instead of changing all this

src/lang/en-US.json
```diff
@@ -92,6 +92,7 @@
   "app_rivalry_team": "Bet on {0} Matches with Rivalry",
   "app_refresh": "Refresh Match History: Queue a scan to find missing matches due to privacy settings",
   "app_refresh_label": "Refresh",
+  "app_use_turbo_label": "Use Turbomode",
```
howardchung, line 95: I'd probably call this "Show Turbo Stats"

howardchung:
Note: recentMatches I don't believe supports the filter, but you can apply it to all the other calls

PR #2678, opened 2020-10-14
Added button to toggle starting items on overview

27bslash:


howardchung:
Does this require the corresponding backend change to preview?

In general though I feel that we should keep the overview as simple as possible. Maybe we can put this on another tab?

27bslash:
Here's an image of the changes:

image

If the replay is not parsed the starting items button does not show up.

image

Argn0:
I don't like this button from an aesthetic point of view.

Argn0:
Also the starting items are already in the Items tab

27bslash:
The starting items in the items tab are inaccurate, it does not account for shared items, sold items or item charges:
image
each one of these should have 2 sets of tangos for example.

Also the button is the same as the show consumables button from the items tab, I tried to play it safe and re-use an existing element. Do you have any suggestions on how to change it?

Argn0:
I suggest fixing the items tab instead so it displays the starting items correctly.

PR #2685, opened 2020-10-28
Added overlay for charges to items in items tab

27bslash:
image

### DIFF
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -657,7 +657,7 @@ export default (strings) => {
                 })
                 .map((purchase) => {
                   if (items[purchase.key] && (showConsumables || items[purchase.key].qual !== 'consumable')) {
-                    return inflictorWithValue(purchase.key, purchase, 'purchase');
+                    return inflictorWithValue(purchase.key, formatSeconds(purchase.time), 'purchase', null, null, purchase.charges);
                   }
                   return null;
                 })
--- a/src/components/Match/matchColumns.jsx
+++ b/src/components/Match/matchColumns.jsx
@@ -657,7 +657,7 @@ export default (strings) => {
                 })
                 .map((purchase) => {
                   if (items[purchase.key] && (showConsumables || items[purchase.key].qual !== 'consumable')) {
-                    return inflictorWithValue(purchase.key, formatSeconds(purchase.time), 'purchase', null, null, purchase.charges);
+                    return inflictorWithValue(purchase.key, formatSeconds(purchase.time), null, null, null, purchase.charges);
                   }
                   return null;
                 })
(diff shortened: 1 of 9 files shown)


src/components/Match/matchColumns.jsx
```diff
@@ -657,7 +657,7 @@ export default (strings) => {
                 })
                 .map((purchase) => {
                   if (items[purchase.key] && (showConsumables || items[purchase.key].qual !== 'consumable')) {
-                    return inflictorWithValue(purchase.key, formatSeconds(purchase.time));
+                    return inflictorWithValue(purchase.key, purchase, 'purchase');
```
howardchung, line 660: instead of changing the type of the second argument can we just add a new additional one for the charges overlay?

27bslash: I think I've made the changes you've requested, however I'm pretty sure I've screwed up the commit history a bit. Sorry about this I'm new to prs.

src/components/Match/matchColumns.jsx
```diff
@@ -657,7 +657,7 @@ export default (strings) => {
                 })
                 .map((purchase) => {
                   if (items[purchase.key] && (showConsumables || items[purchase.key].qual !== 'consumable')) {
-                    return inflictorWithValue(purchase.key, formatSeconds(purchase.time));
+                    return inflictorWithValue(purchase.key, formatSeconds(purchase.time), 'purchase', null, null, purchase.charges);
```
howardchung, line 660: is the new type 'purchase' needed here? We can just display charges if it's provided, right?

src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -230,6 +238,16 @@ class InflictorWithValue extends React.Component {
             <div className="buffOverlay">
               {value > 0 && value}
             </div>
+          }
+            {type === 'purchase' &&
```
howardchung, line 242: Don't think we need to check for type here if we aren't providing charges anywhere else

howardchung (approved):

Argn0:
I suspect this PR broke the times stamps for backpack items in the match overview.

howardchung:
Looks like they're still there but got moved to the top somehow?

src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -231,6 +239,14 @@ class InflictorWithValue extends React.Component {
               {value > 0 && value}
             </div>
           }
+            <div className='textOverlay'>
```
howardchung, line 242: think we want to not render this div at all, unless we have charges content to put here. That should fix the issue with backpack items (unless there's also a weird interaction between backpack items with charges)

howardchung: we should be able to do just chargeOverlay right?  textOverlay didn't exist before this change?

PR #2687, opened 2020-11-05
implement turbomode filter

### LINKED ISSUE #2663: Turbo mode button

howardchung:
fixes #2663 

@bdbch you might want to check this, it's a simplified version of your PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,11 +20,6 @@
     "storybook": "start-storybook -p 6006 -s public",
     "storybook:build": "build-storybook -s public"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm run lint"
-    }
-  },
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -20,11 +20,6 @@
     "storybook": "start-storybook -p 6006 -s public",
     "storybook:build": "build-storybook -s public"
   },
-  "husky": {
```
howardchung, line 23: This is taking way too long, we can bring it back if we make it lint-staged like watchparty does

bdbch: Sure thing!

bdbch (approved):

bdbch:
Looks good to me

Issue #2694, opened 2020-11-12
turbo toggle missing label/looks weird on mobile


howardchung:
Visit a player profile on mobile, toggle looks off. Likely need to fix some css

Shaka-n:
Hey @howardchung, is anyone working on this? Because if not, I'm interested in picking this up.

howardchung:
It's all yours if you want it!

Edu93Jer:
@howardchung Hi there, dude Im new to the project, could you give me like more context about the issue, so I can take a look at it in the code

PR #2704, opened 2020-12-06
 fix turbo toggle looks weird on mobile (#2694) 

tmmluis:
It seems this line was doing nothing other than making the turbo toggle look weird.

### DIFF
--- a/src/components/Player/Header/PlayerButtons.jsx
+++ b/src/components/Player/Header/PlayerButtons.jsx
@@ -26,7 +26,6 @@ const Styled = styled.div`
     }
 
     & * {
-      font-size: 0 !important;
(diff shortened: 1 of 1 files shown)


howardchung:
Thanks! Looks like the label is still missing though which is kind of confusing. Would you like to look into that or make it a separate issue?

tmmluis:
> 
> 
> Thanks! Looks like the label is still missing though which is kind of confusing. Would you like to look into that or make it a separate issue?

Oh Hi, you are right, I just wasn't sure the missing labels were an issue too. They now show both on the turbo and refresh button. Hope that's OK!

howardchung (approved):

PR #2708, opened 2020-12-11
fix backpack time stamps

Argn0:
this appears to fix the issue

### DIFF
--- a/src/components/Visualizations/inflictorWithValue.jsx
+++ b/src/components/Visualizations/inflictorWithValue.jsx
@@ -239,14 +239,14 @@ class InflictorWithValue extends React.Component {
               {value > 0 && value}
             </div>
           }
+            {charges &&
             <div className='textOverlay'>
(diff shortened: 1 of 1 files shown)


src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -239,14 +239,14 @@ class InflictorWithValue extends React.Component {
...
             </div>
           }
+            {charges &&
             <div className='textOverlay'>
-              {charges &&
               <div className="chargeOverlay">
                 {charges}
               </div>
-              }
               <div className="overlay">{value}</div>
```
howardchung, line 247: aren't we rendering the value already in line 239? Not sure if this is needed here

Argn0: I think you're right.

howardchung (approved):

PR #2718, opened 2020-12-26
circular percentage chart for levels

Argn0:
https://deploy-preview-2718--opendota-staging.netlify.app/matches/5756960669

### DIFF
--- a/src/components/Match/StyledMatch.jsx
+++ b/src/components/Match/StyledMatch.jsx
@@ -68,6 +68,31 @@ export const StyledAghanimsBuffs = styled.div`
   }
 `
 
+export const StyledLevel = styled.div`
+  position: relative;
+
+  .level_number {
+    position: absolute;
+    top: 4px;
+    right: 5px;
+    width: 14px;
+    text-align: center;
+  }
+
+  .circular_chart {
+    display: block;
+    max-width: 80%;
+    max-height: 25px;
+    fill: none;
+  }
+
+  .circle {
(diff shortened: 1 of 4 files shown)


howardchung:
Nice! Think it could use a little more padding between number and circle.

Also this might be a bit confusing as in game it shows the amount of exp needed to get to the next level

Argn0:
I've slightly increased the padding.
I'm afraid I can't think of a solution to solve the second issue, but I assume people figure it out quickly.

howardchung:
Also it doesn't look quite centered (too far to the right?)

src/components/Match/StyledMatch.jsx
```diff
@@ -68,6 +68,31 @@ export const StyledAghanimsBuffs = styled.div`
   }
 `
 
+export const StyledLevel = styled.div`
+  position: relative;
+  left: 2px;
+
+  .level_number {
+    position: absolute;
+    top: 6px;
```
howardchung, line 77: would be cool if we can center it automatically with properties rather than pixel values

Argn0: I'll try but I'm not sure if it's possible.

Argn0:
If there is still something off just let me know.

Argn0:
OK, I think this is as good as it gets. There was an issue with the table headers that prevented me from centering properly.

howardchung (approved):

PR #2721, opened 2020-12-30
Add net worth to match overview table

### LINKED ISSUE #2649: Include Player's Net Worth in Match API
A Player's Net Worth is one of the most important stats of a match summary. Please include this in the Match API.

btRooke:
Attempts to fix #2649.

Preview: https://jovial-poincare-7b90c0.netlify.app/matches/5753863230

Adds a net worth column to the match overview table, and resizes the column paddings around to make it fit. I'm very happy to change the column position if anyone's unhappy with it.

I'm unfortunately only able to provide the `en-US` text for this so that's going to need translating, not sure what the protocol is for that.

There's also some funkiness going on with the buffs column, there doesn't seem to be enough room to fit more than 2 there without overflow. An issue should probably be created to sort that.

This is effectively my first proper code contribution to open source so to criticism is incredibly welcome.


howardchung:
Seems fine to me!

I believe this field will be wrong if the net worth goes over 65535 but that shouldn't be too common.

howardchung:
Also your PR will auto-trigger a netlify preview

howardchung:
and it's fine to do English only, we add the translations later

howardchung (approved):

btRooke:
> Also your PR will auto-trigger a netlify preview

Oh lovely - not sure how everyone else was doing those and thought they were very nice.

> I believe this field will be wrong if the net worth goes over 65535 but that shouldn't be too common.

Right - presumably that's an API quirk and not something that should have been fixed here?

PR #2724, opened 2021-01-10
Add a temporary fix for hoodwink icon

AsherWright:
Follows this temporary solution: https://github.com/odota/web/pull/2330

image

Definitely a better way than this, but easy enough to follow along with the method for previous couple heroes.

**This isn't tested**


src/components/Visualizations/HeroImage.jsx
```diff
@@ -7,9 +7,9 @@ import { customHeroImage, IMAGESIZE_ENUM } from '../../utility';
 const HeroImage = ({
   id, isIcon, imageSizeSuffix = IMAGESIZE_ENUM.LARGE.suffix, heroImageEventProps, ...rest
 }) => {
-  // Temporary fix for the CDN icons for void spirit and snapfire:
-  if (isIcon && (Number(id) === 126 || Number(id) === 128)) {
-    return <img src={`/assets/images/dota2/heroes/${id}${isIcon ? '_icon' : ''}.png`} alt="" {...rest} {...heroImageEventProps} />;
```
AsherWright, line 12: We don't need `isIcon` since it's in the conditional check

AsherWright:
@howardchung

howardchung (approved):

PR #2794, opened 2021-05-24
fix blurry item images in Chromium

Argn0:


src/components/Visualizations/inflictorWithValue.jsx
```diff
@@ -33,6 +33,7 @@ display: inline-block;
   margin: 1px;
   height: 27px;
   z-index: 1;
+  image-rendering: -webkit-optimize-contrast;
```
howardchung, line 36: why is this specific to this one image class?

PR #2797, opened 2021-06-02
Improve 'Abilities Drafted' table contents.

jlyneu:
-Previously the abilities that appear in the 'Abilities Drafted' table
were added if they had an ability id of less than 5900. This was okay
for most abilities, but it would include some talent upgrades and
exclude abilities from newer heroes such as Mars, Hoodwink, Dawnbreaker,
etc. since their ability ids are greater than 5900.
-Now the `abilities` array that is populated in transformMatch will
include ability upgrades regardless of the id value as long as the
ability name doesn't include 'special_bonus'.

Before:
ad_before

After:
ad_after

### DIFF
--- a/src/actions/transformMatch.js
+++ b/src/actions/transformMatch.js
@@ -144,7 +144,7 @@ function generateVisionLog(match) {
 }
 
 function transformMatch(m) {
-  const { strings } = store.getState().app;
+  const { abilityIds, strings } = store.getState().app;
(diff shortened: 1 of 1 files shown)


howardchung:
Thanks this seems like a better version of https://github.com/odota/web/pull/2444

howardchung (approved):

PR #2833, opened 2021-10-01
Updated README.md

shi-wal:


builder-247 (changes requested):
Aside from being low effort, the README file refers to a library called `styled-components`

PR #2869, opened 2021-11-07
Combos page: display results in the exact same order as selection

Argn0:
Before:
dd2

After:
ddw

### DIFF
--- a/src/components/Combos/Combos.jsx
+++ b/src/components/Combos/Combos.jsx
@@ -178,8 +178,8 @@ class Combos extends React.Component {
 
   state = {
     queryType: this.parsedUrlQuery.queryType || 'public',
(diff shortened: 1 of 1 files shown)


howardchung:
Do you think this is better for usability? I kind of like preserving the order from the match

Argn0:
> Do you think this is better for usability? I kind of like preserving the order from the match

Yes, this is something that has bothered me personally. I believe it is much clearer this way.

Argn0:
> Do you think this is better for usability? I kind of like preserving the order from the match

Also isn't the order from the match random anyway and has no meaning?

Argn0:
One advantage is that you can much easier see the other heroes in the match that were not selected.

Issue #2880, opened 2021-12-24
Crash on event matches


howardchung:
Looks like a client exception, we should fix this:
https://www.opendota.com/matches/6336125925

RiteshK555:
can I take ths?

Vivek-04022001:
Can you please , help me how can i solve this issue

PR #2882, opened 2021-12-24
Fix: crash on event matches 

RiteshK555:
Tried to fix Crash on event matches #2880
I'm new to opensource contribution so please point out my mistakes


howardchung:
This does seem to fix the problem but it's not clear why the pad function is wrong. Can you look at where it's being used and what's being padded that's causing this issue?

howardchung:
Also please remove package-lock from the commit

archcorsair (approved):

Issue #2865, opened 2021-11-02
Label overflows hero icon in picks/bans section at match page


102:
**Current Behavior**
Label for hero pick icon does not fit icon at match page

**Expected behavior/code**
Label should not overflow the icon's container

**Environment**
- Browser Version: Chrome 95.0.4638.54 
- OS: Windows 11
- Language: German (Deutsch)

**Possible Solution**
Options:
1. Container might be expanded to fit the long text
2. Container might be updated to make sure that text of any width will fit
3. Text might be truncated

**Additional context/Screenshots**
![image](https://user-images.githubusercontent.com/5839225/139845881-3b712fc5-c406-4972-9de4-3bced10421cd.png)

https://www.opendota.com/matches/6254103233

jzhao62:
This PR is fixed and should be closed

PR #2889, opened 2022-01-18
Pickbans UI improvements

TDP17:
Fixes issue #2865 by making image size equal to label size (changes only affect German language currently). 
Ban numbering also starts from 1 now for All Draft matches (could not find matches with mode All Pick in recent games)

PC 
image
image

Mobile
image
image

### DIFF
--- a/src/components/Match/Overview/PicksBans.jsx
+++ b/src/components/Match/Overview/PicksBans.jsx
@@ -40,6 +40,7 @@ margin-top: -20px;
 img {
   position: relative;
   height: 29px;
+  width:100%;
   box-shadow: 0 0 5px ${constants.defaultPrimaryColor};
   margin-right: 0;
   z-index: 0;
@@ -49,7 +50,6 @@ img {
   }
 
   @media only screen and (max-width: 716px) {
-    width: 42px;
     object-fit: cover;
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/Match/Overview/PicksBans.jsx
```diff
@@ -70,21 +66,26 @@ img {
...
-          </aside>
-        </section>
-      ))}
-    </div>
-  </Styled>
-);
+
+const PicksBans = ({ gameMode, data, strings, style }) => {
+  let counter = 0;
+  console.log(gameMode);
```
howardchung, line 72: remove

src/components/Match/Overview/PicksBans.jsx
```diff
@@ -70,21 +66,26 @@ img {
...
-            {pb.is_pick ? strings.match_pick : strings.match_ban} <b>{pb.order + 1}</b>
-          </aside>
-        </section>
-      ))}
-    </div>
-  </Styled>
-);
+
+const PicksBans = ({ gameMode, data, strings, style }) => {
+  let counter = 0;
```
howardchung, line 71: what's this for?

src/components/Match/Overview/PicksBans.jsx
```diff
@@ -31,15 +31,12 @@ margin-top: -20px;
       line-height: 1.6;
     }
   }
-
```
howardchung, line 34: I assume this is a mobile optimization--does this change work well with mobile?

src/components/Match/Overview/PicksBans.jsx
TDP17: Whoops, that is supposed to be the ban number, renamed now. I just added a variable because I could not think of a more elegant way to do it

src/components/Match/Overview/PicksBans.jsx
```diff
@@ -70,21 +66,25 @@ img {
...
+  let banNumber = 0;
+  return (
+    <Styled style={style}>
+      <div className="PicksBans">
+        {data.map(pb => (
+          <section key={pb.order}>
+            <HeroImage id={pb.hero_id} imageSizeSuffix={IMAGESIZE_ENUM.SMALL.suffix} data-isPick={pb.is_pick} />
+            {!pb.is_pick && <div className="ban" />}
+            <aside>
+              {pb.is_pick ? strings.match_pick : strings.match_ban} <b>{gameMode === 22 ? pb.is_pick ? pb.order + 1 : banNumber+=1 : pb.order + 1}</b>
```
howardchung, line 80: can we avoid mutations in this rendering code? It can be confusing.

TDP17: [82055b6](https://github.com/odota/web/pull/2889/commits/82055b6d3983922aa3d09f0149b7d944ecfeb9d3) seems to work.

PS Hit some issues merging causing extra commits,  please lmk if theres anything else

PR #2898, opened 2022-02-15
Simplify lane role icons

Argn0:


howardchung:
hmm, I don't know if I like these better. They're more readable but don't look as good IMO. Open to other opinions

radiusgreenhill:
The new icons are too Radiant-centric. 
I prefer the current icons.

PR #2940, opened 2022-05-05
Changed item tooltip display to separate multiple hints

tinvernizzi:
Old :
image

New :
new

### DIFF
--- a/src/components/ItemTooltip/index.jsx
+++ b/src/components/ItemTooltip/index.jsx
@@ -235,7 +235,7 @@ const ItemTooltip = ({ item, inflictor }) => (
       }
       return null;
     })}
-    {item.hint && <Hint>{item.hint}</Hint>}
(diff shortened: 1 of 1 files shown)


src/components/ItemTooltip/index.jsx
```diff
@@ -235,7 +235,7 @@ const ItemTooltip = ({ item, inflictor }) => (
       }
       return null;
     })}
-    {item.hint && <Hint>{item.hint}</Hint>}
+    {item.hint && item.hint.map((hint) => <Hint>{hint}</Hint>)}
```
howardchung, line 238: Are there any cases where the hint is not an array yet? To prevent a crash should we do like hint?.map

tinvernizzi: Good catch, I'll add that

howardchung (approved):

PR #2953, opened 2022-06-22
Added npmrc and updated package.json for build

kshammer:


akronb:
@kshammer Hi! I'm the one who updated app to `create-react-app 4.0.3`. It seems that something has changed in `npm 8.11`, and now the `@typescript-eslint/parser` is not loaded from `react-scripts`.

Can you downgrage `@typescript-eslint/parser` to `4.33.0` (this version is installed with `react-scripts`) for better compatibility, please?

PR #3101, opened 2023-06-13
fix map for 7.33 version

vik-borisov:


package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "start": "react-scripts start",
+    "start": "react-scripts --openssl-legacy-provider start",
```
howardchung, line 8: Is this change needed?

src/components/Match/BuildingMap/buildingData.js
```diff
@@ -18,31 +18,31 @@ export default [
   },
   {
     id: 't1br',
-    style: { top: '83%', left: '78%' },
+    style: { top: '82%', left: '75%' },
```
howardchung, line 21: Hmm, changing these values directly may cause old matches to render improperly. Can we put the new values under a new var and use it conditionally?

howardchung (approved):

PR #3103, opened 2023-06-15
Move from CRA to Vite

bdbch:
## THIS PR IS STILL WIP

This PR migrates the project from Create React App to Vite. Why?

https://semaphoreci.com/blog/vite

In general CRA is slower than Vite. Vite is pretty fast as it is using ESMby default skipping the whole bundling process for es modules.

From my tests everything looked okay. I guess secrets would need to be updates on the hosting part (as instead of using `REACT_APP_*` for env vars Vite is falling back to `VITE_`)

From my feeling Vite feels way faster and snappier on building & starting up the project, specially after the app was cached.

**Important**
One thing that didn't work anymore was the `heatmap.js` package as it doesn't seem to support Vite. Looking at the activity of the package (https://github.com/pa7/heatmap.js) it seems to be quite dead.

To fix this issue for now I built a very basic heatmap with canvas on my own. It's not as beautiful as heatmap.js heatmaps because of the missing color grading between points - but this could be built afterwards.

Otherwise the PR currently does not load the page since something is not working with Recharts. Looking at this tomorrow.


src/components/Heatmap/Heatmap.jsx
```diff
@@ -30,38 +28,61 @@ const drawHeatmap = ({
...
+    const greenValue = Math.floor(255 * (1 - (p.value / adjustedData.max)));
+    const alphaValue = (p.value / adjustedData.max) - ((p.value / adjustedData.max) * .35);
+
+    ctx.fillStyle = `rgba(${redValue}, ${greenValue}, 0, ${alphaValue})`;
+    ctx.fill();
+  })
+
+  // blur whole heatmap
+  ctx.drawImage(heatmap.current, 0, 0);
+  // heatmap.setData(adjustedData);
```
bdbch, line 53: can be removed

howardchung:
I've done a little research into converting CRA projects and it seems like Vite is an easier drop-in replacement for CRA, so if the Next.js effort is stalled it seems we could get some dev wins out of figuring out the blockers here

PR #3107, opened 2023-06-21
show result of the laning stage

vik-borisov:
image


src/actions/transformMatch.js
```diff
@@ -5,7 +5,7 @@ import {
   isRadiant,
   isSupport,
   getLevelFromXp,
-  unpackPositionData,
+  unpackPositionData
```
howardchung, line 8: we do trailing commas

src/components/Match/Laning/index.jsx
```diff
@@ -39,7 +47,7 @@ class Laning extends React.Component {
         <StyledFlexElementFullWidth>
           <Heading title={strings.heading_laning} />
           <Table
-            data={match.players}
+            data={match.players.sort(this.defaultSort)}
```
howardchung, line 50: will this mutate the order of match.players for other tabs?

src/lang/bg-BG.json
```diff
@@ -733,6 +733,7 @@
   "th_biggest_hit": "Biggest Hit",
   "th_lane": "Lane",
   "th_map": "Map",
+  "th_win_lane": "Win",
```
howardchung, line 736: Only the en file needs to be updated--we have processes to do translations but they haven't been run in a while

howardchung: If it's not too much trouble--otherwise it may cause conflicts if we ever run the language update script (but TBH we haven't done that in a long time)

howardchung (approved):

PR #3125, opened 2023-11-21
convert App/ directory to use TypeScript

mehrdadrafiee:
Please review.

As part of refactoring, this PR converts `src/components/App/` directory to use TypeScript. There were some spots that I wasn't sure how to handle the types so I had to ignore them (@ts-nocheck) for now but feel free to pull down the changes and update the PR.

Since the styled-component is a bit old, I wasn't able to convert the `src/components/App/GlobalStyles.jsx` to TypeScript. Can be done once we upgrade to a newer version.


src/components/App/App.tsx
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
howardchung, line 1: can we just apply ts-ignore to specific lines instead

src/components/App/App.tsx
```diff
@@ -33,57 +33,78 @@ import GlobalStyle from './GlobalStyle';
...
+interface Back2TopStylesProps {
+  open?: boolean;
+  location: {
+    pathname: string;
+  };
+}
+
+interface AppProps {
+  location: { [key: string]: string }
+  strings: { [key: string]: string }
```
howardchung, line 52: Can we use one shared type for strings so if we figure out how to type it better later we only have one spot to update?

src/components/App/App.tsx
```diff
@@ -137,7 +161,7 @@ const App = (props) => {
     return () => {
       window.removeEventListener('scroll', handleScroll);
     };
-  });
+  }, []);
```
howardchung, line 164: doesn't this change the behavior to only happen once?

src/components/App/App.tsx
mehrdadrafiee: can use `GlobalStrings` in `types/common` directory, right?

src/components/App/App.tsx
mehrdadrafiee: isn't it the expected behavior? clicking on the "back to top" button takes the user to the top of the screen and then the button disappears. user won't be able to click it multiple times.

src/components/App/App.tsx
mehrdadrafiee: done!

src/components/App/App.tsx
howardchung: If they scroll back down though we want to reenable it right?

src/components/App/App.tsx
howardchung: Yeah if that's an alias

mehrdadrafiee: hmmm... I think will be facing the second option mentioned here: https://github.com/odota/web/pull/3124#discussion_r1398334357

howardchung: OK, in the meantime can you point all the string references to one place instead of duplicating { [key: string]: string }?

howardchung: Actually it's probably best to use Record<string, string> which is a builtin type

mehrdadrafiee: I just updated the PR with the builtin type as you suggested. I can go ahead and update other instances of `{ [key: string]: string }` in a separate PR if you approve.

thanks and sorry for late reply :)

howardchung (approved):

PR #3138, opened 2024-01-08
fix: 404 images on hero page

### LINKED ISSUE #3137: 404 images on hero page, https://www.opendota.com/heroes/86
**Current Behavior**

https://www.opendota.com/heroes/86

404 image

image

**Expected behavior/code**

show image

Geczy:
Closes #3137

### DIFF
--- a/src/components/Hero/Abilities.jsx
+++ b/src/components/Hero/Abilities.jsx
@@ -34,8 +34,8 @@ const renderAbilities = abilities => abilities.map(ability => (
 ));
 
 const Abilities = ({ hero }) => {
-  const filterAbilities = toFilterAbs => toFilterAbs.filter(ability => (ability !== 'generic_hidden'));
-
+  const filterAbilities = (toFilterAbs) =>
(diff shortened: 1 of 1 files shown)


src/components/Hero/Abilities.jsx
```diff
@@ -34,8 +34,8 @@ const renderAbilities = abilities => abilities.map(ability => (
 ));
 
 const Abilities = ({ hero }) => {
-  const filterAbilities = toFilterAbs => toFilterAbs.filter(ability => (ability !== 'generic_hidden'));
-
+  const filterAbilities = (toFilterAbs) =>
+    toFilterAbs.filter((ability) => ability !== 'generic_hidden' && abilities[ability].behavior !== 'Hidden');
```
howardchung, line 38: We can remove generic_hidden too?

Geczy: We can, but it's more efficient to leave it in since it's a quick check that doesn't require an object lookup in a large json. Happy to remove though, an object lookup shouldn't be too bad since it's a direct key substitute.

PR #3144, opened 2024-01-20
update backpack images width

Argn0:
forgot backpack images


Argn0:
Also it seems like ublock origin on Firefox severely slows down image loading times. Maybe it's because we're using the <object> tag?

Argn0:
But do we even want to "support" adblock?

builder-247:
This breaks the backpack item hover image dimensions

PR #3150, opened 2024-02-07
Clear w/l display on player page switch

### LINKED ISSUE #3148: Clear w/l display on player page switch
Currently the numbers for the previous player still appear when navigating to a new player page. We should show a spinner or at least clear the old values to avoid confusing users.

jeyren95:
Cause of issue:
- the "getPlayerWinLoss" API is being called while navigating between player profiles
- however, the "playerWinLoss.loading" state is not being checked while rendering the components

Proposed changes:
- added a check for "playerWinLoss.loading"
- placed the above check inside "PlayerHeader" component because the component relies on data sets from 2 individual APIs
- this is to prevent different portions of the component from displaying the desired content at different speeds

fixes #3148

### DIFF
--- a/src/components/Visualizations/Table/HeroImage.jsx
+++ b/src/components/Visualizations/Table/HeroImage.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { withRouter } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import ReactTooltip from 'react-tooltip';
@@ -464,6 +465,7 @@ class TableHeroImage extends React.Component {
       predictedVictory,
       leaverStatus,
       strings,
+      match,
       hero = {},
     } = this.props;
     const { tooltipVisible } = this.state;
@@ -476,6 +478,9 @@ class TableHeroImage extends React.Component {
         this.setTooltipVisibility(false);
       },
     };
+
+    const { params } = match;
+    const { playerId } = params;
 
     return (
       <Styled style={expand}>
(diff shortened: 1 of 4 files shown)


howardchung:
I think we can achieve this without changing to link elements. Note that when navigating player profile to player profile, the matches, heroes, and peers show a loader without a page refresh. We want to achieve the same thing for the W/L number.

jeyren95:
i see, ok after digging deeper, i managed to find the cause of the issue. Updated the PR! Please help to take a look, thanks Howard!

src/components/Visualizations/Table/HeroImage.jsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { withRouter } from 'react-router-dom';
```
howardchung, line 2: I don't think this file needs to be modified?

howardchung:
Thanks! This is definitely better--ideally we would only show a separate loading state for the win loss section since that tends to load a little slower than the profile info, but that's not required

jeyren95:
hmmm yeah, actually if you want to show a separate loading state for the win loss section, i could just remove the extra condition in the "PlayerHeader" component. 

But my initial thought was that it would be more UI-friendly to load the entire component together rather than seeing portions of it loading separately.

What dyou think?

howardchung:
I think we should show a loader only for the w/l data. If the name/picture etc. are available first there's no reason to block on it.

jeyren95:
ok done, have removed the extra condition

howardchung:
Looks good! I guess we can leave for a future item improving the loader apperance (maybe a shimmer instead of a spinner) and left align it.

PR #3151, opened 2024-02-11
feat(issue-3133): hide player matches if player profile is private

jeyren95:
1. Added retrieving of list of pro player IDs
2. Player profile should be private if player account id is not within the list of pro player IDs, and the player's "fh_unavailable" property is true
3. Disabled tabs if player profile is private
4. Added CTA to direct players to login if their profile is private, but they wish to see their stats

### DIFF
--- a/src/components/Player/Header/PlayerHeader.jsx
+++ b/src/components/Player/Header/PlayerHeader.jsx
@@ -3,13 +3,14 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import Avatar from 'material-ui/Avatar';
 import Badge from 'material-ui/Badge';
+import LockIcon from '@material-ui/icons/Lock';
 import styled from 'styled-components';
 import { Facebook } from 'react-content-loader';
 import { rankTierToString } from '../../../utility';
 import Error from '../../Error';
 import PlayerStats from './PlayerStats';
 import PlayerBadges from './PlayerBadges';
-import PlayerButtons from './PlayerButtons';
+import PlayerButtons from './PlayerButtons'
 import constants from '../../constants';
 
 const Styled = styled.div`
@@ -42,6 +43,7 @@ grid-template-columns: 1fr minmax(min-content, ${constants.appWidth}px) 1fr;
   flex-direction: row;
   justify-content: flex-start;
   flex-wrap: wrap;
+  margin-bottom: 2%;
 
   @media only screen and (max-width: 768px) {
     flex-direction: column;
@@ -188,6 +190,11 @@ grid-template-columns: 1fr minmax(min-content, ${constants.appWidth}px) 1fr;
   }
 }
 
+.lockIcon {
+  @media only screen and (max-width: 768px) {
+    text-align: center;
+  }
+}
 `;
 
 const LARGE_IMAGE_SIZE = 124;
@@ -285,6 +292,7 @@ const PlayerHeader = ({
   rankTier,
   leaderboardRank,
   strings,
+  isPlayerProfilePrivate,
 }) => {
   if (error) {
     return <Error />;
@@ -340,8 +348,16 @@ const PlayerHeader = ({
               <span className="playerName">{officialPlayerName || playerName}</span>
               <PlayerBadges playerId={playerId} />
             </div>
-            <PlayerStats playerId={playerId} loggedInId={loggedInUser && String(loggedInUser.account_id)} compact={!small} />
-            <PlayerButtons playerId={playerId} playerSoloCompetitiveRank={playerSoloCompetitiveRank} compact={!small} />
+            {isPlayerProfilePrivate ? (
+              <div className="lockIcon">
+                <LockIcon />   
+              </div>
+            ) : (
+              <>
+                <PlayerStats playerId={playerId} loggedInId={loggedInUser && String(loggedInUser.account_id)} compact={!small} />
+                <PlayerButtons playerId={playerId} playerSoloCompetitiveRank={playerSoloCompetitiveRank} compact={!small} />
+              </>
+            )}
           </div>
           <div style={{ display: 'flex' }}>
             {getDotaPlusBadge(plus, strings)}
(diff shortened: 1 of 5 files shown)


src/components/Player/Header/PlayerHeader.jsx
```diff
@@ -340,8 +348,16 @@ const PlayerHeader = ({
               <span className="playerName">{officialPlayerName || playerName}</span>
               <PlayerBadges playerId={playerId} />
             </div>
-            <PlayerStats playerId={playerId} loggedInId={loggedInUser && String(loggedInUser.account_id)} compact={!small} />
-            <PlayerButtons playerId={playerId} playerSoloCompetitiveRank={playerSoloCompetitiveRank} compact={!small} />
+            {isPlayerProfilePrivate ? (
+              <div className="lockIcon">
+                <LockIcon />
```
howardchung, line 353: I think we can just not show anything here

src/components/Player/Player.jsx
```diff
@@ -7,11 +7,13 @@ import Long from 'long';
 import {
   getPlayer,
   getPlayerWinLoss,
+  getProPlayers,
```
howardchung, line 10: I don't think we need to fetch the entire list of pro players. The player response should have a "name" field which is the pro player name if available. If this is set, the player is a pro.

howardchung: I think this is what "officialPlayerName" is here

src/lang/en-US.json
```diff
@@ -1203,5 +1203,8 @@
   "killed": "killed",
   "team_courier": "{team}'s courier",
   "slain_roshan": "Have slain Roshan",
-  "drew_first_blood": "Drew First Blood"
+  "drew_first_blood": "Drew First Blood",
+  "player_profile_private_title": "THIS PROFILE IS PRIVATE" ,
```
howardchung, line 1207: We should just use regular casing here and css to transform it to uppercase

src/components/Player/playerPages.jsx
```diff
@@ -90,7 +90,8 @@ const playerPages = strings => [{
   content: (playerId, routeParams, location) => (<ActivityPage playerId={playerId} routeParams={routeParams} location={location} />),
 }];
 
-export default (playerId, strings) => playerPages(strings).map(page => ({
+export default (playerId, strings, isPlayerProfilePrivate) => playerPages(strings).map(page => ({
   ...page,
   route: `/players/${playerId}/${page.key}`,
+  disabled: isPlayerProfilePrivate,
```
howardchung, line 96: where is this getting used?

jeyren95: image

The existing "Tab" component accepts a "disabled" prop. The idea here is to disabled all the tabs if the profile is private.

So in "playerPages.jsx", where the pages are being mapped to the tabs, I added in an extra "disabled" property that will eventually be passed to the "Tab" component

src/components/Player/Header/PlayerHeader.jsx
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import Avatar from 'material-ui/Avatar';
 import Badge from 'material-ui/Badge';
+import LockIcon from '@material-ui/icons/Lock';
```
howardchung, line 6: lockicon stuff no longer needed here?

src/components/Player/Player.jsx
howardchung: import not needed?

src/components/Player/Player.jsx
```diff
@@ -32,6 +34,8 @@ class RequestLayer extends React.Component {
     playerName: PropTypes.string,
     playerLoading: PropTypes.bool,
     strings: PropTypes.shape({}),
+    proPlayerIds: PropTypes.arrayOf(PropTypes.number),
```
howardchung, line 37: also remove this

src/components/Player/Player.jsx
```diff
@@ -83,7 +93,8 @@ const mapStateToProps = state => ({
   playerLoading: (state.app.player.loading),
   officialPlayerName: (state.app.player.data.profile || {}).name,
   strings: state.app.strings,
-});
+  isPlayerMatchHistoryDisabled: (state.app.player.data.profile || {}).fh_unavailable,
```
howardchung, line 96: We could just do the officialPlayerName check here right?

src/components/Player/Header/PlayerHeader.jsx
```diff
@@ -188,6 +189,11 @@ grid-template-columns: 1fr minmax(min-content, ${constants.appWidth}px) 1fr;
   }
 }
 
+.lockIcon {
```
howardchung, line 192: This also isn't needed?

src/components/Player/Header/PlayerHeader.jsx
```diff
@@ -42,6 +42,7 @@ grid-template-columns: 1fr minmax(min-content, ${constants.appWidth}px) 1fr;
   flex-direction: row;
   justify-content: flex-start;
   flex-wrap: wrap;
+  margin-bottom: 2%;
```
howardchung, line 45: and probably this

src/components/Player/Player.jsx
howardchung: import still not needed?

src/components/Player/Header/PlayerHeader.jsx
jeyren95: yeah so sorry my bad, have removed the unnecessary code and have double checked again if i left out anything else

howardchung:
I think we missed a case here--for players who are signed in and trying to view their own profile.

howardchung:
Also, we have to allow access to the refresh button to update the availability status--so we should either only hide the w/l row, or not hide anything at all

howardchung:
I will revert this for now and we can re-merge when fixed

jeyren95:
ok noted on this, my bad for missing out on the scenario, will be fixing it asap. Anyway, I think it is alright to not hide anything at all ---> saw some dotabuff private profile examples, and they also do not hide the W/L stat

PR #3153, opened 2024-02-15
Add 7.34 draft order

ZacheryJohnson:
[Patch 7.34](https://www.dota2.com/patches/7.34) changed the draft order and a corresponding mapping was not added to OpenDota. This PR adds support for this new draft order.

---

Tested by using the Battle Cup match [7580513318](https://www.opendota.com/matches/7580513318/draft), which was played on 2024-02-10. 

The draft should look as follows (the Dota client has some issues when jumping ahead in draft in replays it seems; the last pick for Dire was a Dragon Knight)
draft

Prior to this change, the draft of a parsed game on OpenDota incorrectly showed the following:
before

After this change, the draft appears correct in a parsed game:
after

The [Team Liquid v Azure Ray game](https://www.opendota.com/matches/7588089287/draft), played today, also now displays correctly:
tl-xg

Happy to test any additional matches as desired.

---

Note: the timestamp correlates to Tuesday, August 8, 2023 22:46:00 GMT, which I correlated to build ID **11896549** in [Dota's SteamDB entry](https://steamdb.info/app/570/patchnotes/). I'm not sure if there's a more formal way to get an exact timestamp on when this new draft order should take effect, but I'm happy to do more digging if we'd prefer a more precise time.

734-steamdb

### DIFF
--- a/src/components/Match/Draft/Draft.jsx
+++ b/src/components/Match/Draft/Draft.jsx
@@ -188,7 +188,11 @@ const Draft = ({
   let orderOne = [];
   let orderTwo = [];
   let picks = [];
-  if (startTime > 1629255201) { // post 7.30
+  if (startTime > 1691534760) { // post 7.34
+    orderOne = [1, 4, 7, 8, 10, 11, 14, 15, 18, 19, 22, 23];
+    orderTwo = [2, 3, 5, 6, 9, 12, 13, 16, 17, 20, 21, 24];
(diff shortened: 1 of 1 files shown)


howardchung:
I think we have a patch field in the match data (which is controlled by dotaconstants). That would save us having to define timestmaps here, but I think this is fine (we also have something similar for showing minimaps)

PR #3156, opened 2024-02-20
feat(issue-3133): include logged in user scenario

jeyren95:
Player profile is public if:
1. Player is a pro player
2. Player's match history is not disabled (i.e. "fh_unavailable" is false)
3. Player's match history is disabled, but the player is logged in and is viewing own profile

Player profile is private for all other scenarios. It is much easier to just check if the player profile is public, since there seems to be fewer + less complicated scenarios

Do let me know if I have missed out on any scenarios!

image
^ this is how it will look like if we don't hide the W/L stat


howardchung:
This seems fine. Although it's probably cleaner to return the visibility from the backend so we're not implementing this logic twice.

PR #3164, opened 2024-03-17
Fixed size of icons for backpack item costs

robsassack:
This fixes the issue with the sizing of the cost icons described in #3163 by @builder-247.

image

### DIFF
--- a/src/components/ItemTooltip/index.jsx
+++ b/src/components/ItemTooltip/index.jsx
@@ -119,8 +119,8 @@ const HeaderText = styled.div`
 `;
 
 const ResourceIcon = styled.img`
-  width: 16px;
-  height: 16px;
+  width: 16px !important;
+  height: 16px !important;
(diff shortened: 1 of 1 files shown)


src/components/ItemTooltip/index.jsx
```diff
@@ -119,8 +119,8 @@ const HeaderText = styled.div`
 `;
 
 const ResourceIcon = styled.img`
-  width: 16px;
-  height: 16px;
+  width: 16px !important;
```
howardchung, line 122: can we achieve this without !important?

robsassack: It should probably be possible, I can look into it.

robsassack:
@howardchung Looks like it works by changing it to max-height+max-width instead. Does this work?
image

howardchung (approved):

PR #3165, opened 2024-03-19
css fixes

builder-247:
* Reduced font size of item tooltip ability title - even the longer ones should now fit on one line, e.g. Book of the Dead
* Fixed a regression from https://github.com/odota/web/pull/3157 where hero ability icons would be too wide


howardchung:
Not directly related to this but I think we can avoid using the object tag to avoid firefox perf issues

PR #3171, opened 2024-04-16
Display win rate in heroes page

ThePeregrinus:
localhost_5173_heroes_102(iPhone SE)


howardchung:
This issue is actually referring to the heroes summary page where we show all the heroes. We want to show a trendline/sparkline of the win rate over the last 7 days

PR #3176, opened 2024-05-09
Improve item abilities and attributes

builder-247:
This PR fixes two issues with item tooltips:
* Ability name detection was detection based on a regex relying on capitalization, which was inconsistent for a few items e.g. Bloodstone
* Added support for uppercase stats displayed at the bottom:
image

Requires a release from https://github.com/odota/dotaconstants/pull/156


howardchung:
Published new dotaconstants

PR #3350, opened 2026-07-03
feature: style ui

marktoh:
## Changes
- Style landing page
- Add `Roboto Serif` as a serif font
- Increase vertical spacing of list items
- Align text on mobile


src/components/Spacer/index.ts
```diff
@@ -0,0 +1 @@
+export { Spacer } from "./Spacer";
```
howardchung: I don't think we need to re-export this component, just import Spacer/Spacer from the callsites.

marktoh: Alright. Updated.

PR #3354, opened 2026-07-05
style(ui): style histogram, adjust alignment on Match page, add futuristic font

marktoh:
## Changes
- Abstract `Grid` component
- Style `Histogram`
- Adjust alignment on `Match` Page
- Add futuristic font, `Quantico`

### DIFF
--- a/index.html
+++ b/index.html
@@ -74,6 +74,10 @@
       href="https://fonts.googleapis.com/css2?family=Roboto+Serif:wght@400;600;700;800&display=swap"
     />
     <link
+      href="https://fonts.googleapis.com/css2?family=Quantico:ital,wght@0,400;0,700;1,400;1,700&display=swap"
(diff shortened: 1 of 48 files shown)


index.html
```diff
@@ -73,6 +73,10 @@
       rel="stylesheet"
       href="https://fonts.googleapis.com/css2?family=Roboto+Serif:wght@400;600;700;800&display=swap"
```
howardchung, line 74: Have we considered using sans-serif fonts only? I've noticed that we have a mix currently

howardchung (approved):

index.html
marktoh: Yes. The use of sans-serif fonts with serif fonts are to help with contrast and visual hierarchy.

See here: https://elements.envato.com/learn/combining-serif-sans-serif-fonts

Issue #3072, opened 2023-02-13
Revisit CSS solution


howardchung:
Five years after #1189, I'm not such a big fan of styled-components:

* It doesn't work well with syntax highlighting
* It's a weird amalgamation of JS and CSS that loses out on tooling for both

It's probably too much work and not enough benefit to actually migrate everything, but if I were redoing this I'd probably just use CSS modules. Not sure what the state of the art is, so please chime in with thoughts!

talal-najam:
Agreed! I'd vote for CSS modules with Scss. Shared styles, theme / vars files, mixins and other scss goodies along with scoped module styles would be pretty clean in my opinion. 

Alternate options could be Tailwind but I imagine that's not really everyone's cup of tea. Keen to hear what others think.

geracosta:
bumping this since it came up on #3368, where you asked me to use a css module for a new component.

three things changed since 2023 that i think matter here:

styled-components went into maintenance mode in march 2025. not deprecated, but the maintainer said he wouldn't recommend adopting it for new projects. we're on 4.4.1, which is from 2019.

native css nesting hit baseline widely available in june 2026. the suggestion above was css modules + scss, and the nesting half of that is just css now. i don't think we'd need sass at all unless someone wants mixins.

mui added official tailwind v4 support in 7.1 through cascade layers, so the preflight mess that made that combination painful in 2023 is gone. mentioning it because the proposal in #3104 was tailwind and that door is actually open now, even if i don't think a rewrite is the way.

but here's the thing i ran into that blocks all of this regardless of which one we pick. constants.ts is javascript, so no css file can read it. i converted one small component in #3368 and colorMutedLight turned into a hardcoded rgba() sitting in the module. multiply that by the 101 files importing constants and we'd have colors scattered everywhere, which is worse than what we have now.

so whatever we choose, i'd do the tokens first as their own PR: emit constants.ts as custom properties on :root and have constants.ts export var(--od-*) instead of literals. after that mui's sx, the styled-components that are still around, and any new css module all read the same source, and components can move one at a time without freezing anything.

on which one, i'd go plain css modules. native to vite so no new dependency, actual css files which is what you said you wanted in #3104, and it's the only option that doesn't ask someone dropping in for a one-line fix to learn a new system. vanilla-extract would give us typed tokens but .css.ts isn't really css files. panda and stylex feel like a lot of machinery for this codebase, and we're still on react 17 anyway so anything assuming 18+ is out for now.

[...]

PR #3368, opened 2026-08-21
Add deaths summary table to Combat tab

geracosta:
Adds a per-team Deaths table to the Combat tab, built from the `deaths_log` field proposed in odota/parser#88: deaths, gold lost and time dead per player with totals, plus a per-death breakdown of who killed them and when (gold lost and time dead in the tooltip). This is the death summary that was asked for in odota/core#1465.

The table only renders when the match has `deaths_log`, so nothing changes for matches parsed before that field exists.

Screenshot from a local run (match 8754463676, parsed with the parser branch and injected into the dev server):

deaths summary table

The tooltip in the shot is the illusion kill fixed in odota/parser#90, showing the 480 gold and 22s it cost.

Draft until parser#88 lands.


src/components/Match/StyledMatch.tsx
```diff
@@ -231,6 +231,32 @@ export const StyledPlayersDeath = styled.div`
     height: 29px;
   }
 `;
+export const StyledDeathsSummary = styled.div`
```
howardchung, line 234: Ideally we move away from styled-components for new code. I think we're trying to replace it with css modules

geracosta: happy to, but i'd rather not set the pattern from a feature PR. there's no .module.css in the repo yet and the README still lists styled-components, so whatever i write here becomes the convention by default.

what stopped me is constants.ts. these styles use colorMutedLight and a css module can't read it, so converting means hardcoding rgba(255, 255, 255, 0.6) in the file.

i wrote up the longer version in #3072 along with what's changed since you opened it. would rather settle it there and keep this PR consistent with the rest of the Match components. if you'd still prefer the css module here i'll just do it.

howardchung (approved):

PR #3382, opened 2026-09-10
feat: show team icons in matches/pro

kakoed337:
{7342A2B2-36B0-4847-AB74-B950336BBEDE}


howardchung:
This works, but I think it'd be cleaner to have the server side populate the logo URL in the returned results

kakoed337:
ah ok, i'm gonna make pr in core and change this, thanks

PR #3383, opened 2026-09-11
Add per-minute records to the player Records page

Xvie:
Adds **Kills Per Min**, **Hero Damage Per Min** and **Tower Damage Per Min** to the player Records page, so a long game no longer takes every damage record by default. On my own history the top Hero Damage game is 150,773 over 72 minutes, while the top Hero Damage Per Min is a 35 minute game — the two lists rank differently, which is the point.

Computed client side rather than in the API, per the discussion on odota/core#2983: the projection stays limited to columns that actually exist in Cassandra.

### How it works

`getPlayerRecords` branches on the field. Stored columns keep the existing `?sort=&limit=20` request. The three per-minute fields instead fetch the full history and rank it in `transformPerMinuteRecords`, in the same shape as the existing `transformTrends` / `transformHistograms` transforms.

Since `project` replaces the default column set rather than adding to it, the request asks only for what the records table actually renders — hero, result, duration, and the source column. That is 10 fields per row instead of the 18 the default returns, which is what keeps the payload down.

### Payload

Measured against the production API, gzipped, for one fetch that a button click triggers:

| history | rows | gzipped |
| --- | --- | --- |
| 975 matches | 975 | 22 KB |
| 3,084 matches | 3,084 | 64 KB |
| ~10,000 (extrapolated) | — | ~210 KB |

It scales linearly at roughly 21-23 bytes per match. Worth noting this costs the API no extra work: `getPlayerMatches` already reads and sorts a player's entire match set on every call, and `limit` only slices the result at the end, so the existing `limit=20` requests were doing the same reads.

Sorting is negligible — a few ms at 10k rows.

### Testing

`npm run build` (build and typecheck) passes. Driven against the production API with a real 975 match account; all three columns match a value-for-value check computed independently from the raw endpoint:

```
HDM: 2085.55 (35:10), 2073.43 (72:43), 1899.79 (49:56)
KPM: 1.05 (35:10), 0.91 (28:33), 0.91 (32:01)
TDM: 573.54 (31:31), 518.94 (36:20), 473.91 (30:12)
```

No new console errors. `npm test` was not run — `testcafe/testsUtility.ts` targets `localhost:5000`, which is taken by AirPlay Receiver on my machine.

The `th_` strings already existed from the benchmarks work, so only the three `heading_` button labels are new.


howardchung:
Commented on the other PR that I think we should do this calculation client side

Xvie:
Updated to do the calculation client side as suggested — no API change needed, so odota/core#2983 is closed. Details and payload measurements are in the description.

Xvie:
@howardchung can you recheck again please?

howardchung (approved):

PR #3384, opened 2026-09-17
Define design tokens as CSS custom properties

geracosta:
First step of #3072, following what you said about converting the vars to a CSS-compatible format. This doesn't convert any component to css modules, it just makes the values readable from a css file, so the conversions after this don't each end up hardcoding their own colors.

New `src/tokens.css` puts every token on `:root` as `--od-<kebab-name>`, and `constants.ts` returns `var(--od-*)` instead of the literal. Both sides read one value, so mui's `sx`, the styled-components that are still around and any future css module stay in sync, and components can move one at a time.

Four stay literal, each with a comment saying why:

- `wrapTablet` and `wrapMobile` are used inside `@media` conditions across 21 call sites, and `var()` doesn't resolve there.
- `appWidth` is in `@media` too and in arithmetic (`constants.appWidth * 0.483` in Wardmap).
- `dotaMaxLevel` is a plain number used in arithmetic, not a style value.

One thing to flag: this depends on `var()` resolving in SVG presentation attributes, because recharts takes some colors as props rather than styles, the `stroke` on the lines in MatchGraph and the `<Brush>` stroke/fill in MMRGraph. I checked those resolve to the same values they had before. If you'd rather not lean on that, those few can stay literal too.

The MUI theme only uses constants inside `styleOverrides`, which becomes css, so it's unaffected. Nothing in the codebase does string math on a color (no `alpha()` / `darken()`), which would have broken.


howardchung (approved):

howardchung:
This is causing an issue with the abcolor module's gradient function: https://github.com/odota/web/blob/master/src/components/Visualizations/Table/Percent.tsx

I've fixed it by hardcoding the original hex values in that file, but we might want to consider finding an alternative that works with css variables (this is the only place we use the abcolor module: https://github.com/blukai/abcolor)

PR #3387, opened 2026-09-20
Revert "Add per-minute records to the player Records page"

howardchung:
Reverts odota/web#3383

I think this is causing the crash on player pages


howardchung:
Correction: Not this change but the color one. Reverting the revert

